PR #19, opened 2019-02-25
Add AST printer

jensneuse:
This PR will add an AST printer.
The first release will only cover printing  ExecutableDefinitions.
Adding TypeSystemDefinition printing later on will be no big deal.


jensneuse:
Currently fragments are missing.

jensneuse:
Fragments done, next up Directives.

jensneuse:
There should be a FragmentDefinitionIterable with a filter to only emit every fragment definition once to prevent re-printing fragments that were walked across multiple times. Walking fragments multiple times is necessary to find fragments without usage and distinguish between fragments that are being used in an operation. This should be fixed before merging this branch.

jensneuse:
> There should be a FragmentDefinitionIterable with a filter to only emit every fragment definition once to prevent re-printing fragments that were walked across multiple times. Walking fragments multiple times is necessary to find fragments without usage and distinguish between fragments that are being used in an operation. This should be fixed before merging this branch.

I've refactored the walker to prevent the behaviour at all. It was necessary to introduce a new api to the walker which helps easily finding the root operation usages of any node. As a side effect validation times for the introspection query dropped from ~ 400k ns to ~88k ns. All in all I think this is a better solution than filtering.

PR #22, opened 2019-03-02
add proxy poc

jensneuse:
This PR is used to track the progress building a poc graphql proxy.
Initially there's a middleware in the example folder which adds one field/removes another field on a specific type.


jensneuse:
I'll just merge this PR to bring the POC into the master. In addition type system definition printing was added in preparation of fetching remote schemas and printing them out.
Further progress on the proxy will be tracked in more specific branches.

PR #30, opened 2019-03-09
Add context based middleware

jensneuse:
This PR tracks the progress of adding a http context based middleware.
This middleware should react based on schema annotations via directives.

The feature is currently not implemented.
There needs to be a discussion about the package structure and some implementation details.

Thanks to @mvid helping with this feature.


jensneuse:
Hey @mvid, first and foremost thanks for driving this feature.
Your scaffold of the feature is really helpful!

I've tried to cleanup the package structure a bit.
If you are not happy with the updated structure please let me know. I'm open for improvements here.

That being said I think there are two things we should improve on:

1. The middleware shouldn't know about the transport. So I think we should do the following:
1.1 add a context object to the middleware
1.2 extract the context from the http request
1.3 provide the context from the http request handler to the middleware

This way we could ensure to use the middleware independently from the transport. E.g. it should be possible to easily write a grpc handler.

2. The proxy should not use the default http handler. I think it would be best if the user could configure the transport (http client) and configure timeouts etc.. We could provide a default client for ease of use but this should not be the default client, because it doesn't handle timeouts.

Please let me know what you think.

mvid:
@jensneuse I think the new structure is great. I agree about the middleware being transport independent. My reason for making the HTTP handler proxy was 1) Because it was a more realistic testing environment for reasoning about the context based middleware and 2) Because it's likely a common use case in the future.

This looks great, please let me know if there is anything else I can do!

jensneuse:
I've added context to the middleware.
I'll implement the context handler now.
As discussed I'll leave the surrounding parts up to you @mvid .

jensneuse:
@mvid The middleware does now work as expected. The test you provided that was failing initially does now succeed. I think the api could be assumed to be stable for this feature. The middleware code itself needs some refactoring and cleanup. I'll do this on my own and think this should not break anything external.

As we discussed already you might now contribute the following parts:
- providing tests with edge cases to make the feature stable
- make the http proxy client configurable including a default client (which will have timeouts etc.)
- setup a realistic end2end test including http request handling + setting the context

I'll add some internal refactoring as stated. Other than that I'll await your feedback on the implementation + open tasks.

jensneuse:
I've added basic documentation.

PR #41, opened 2019-03-29
Proxy updates

mvid:


cmd/staticProxy.go
```diff
@@ -92,15 +90,19 @@ func runProxyBlocking() {
...
+	prox := fastproxy.NewFastStaticProxy(fastproxy.FastStaticProxyConfig{
 		MiddleWares: []middleware.GraphqlMiddleware{
 			&middleware.ValidationMiddleware{},
 			&middleware.ContextMiddleware{},
 		},
 		RequestConfigProvider: proxy.NewStaticSchemaProvider(proxy.RequestConfig{
 			Schema:              &schema,
-			BackendAddr:         []byte(staticProxyBackendURL),
-			BackendHost:         staticProxyBackendAddr,
+			BackendURL:          *backendURL,
```
mvid, line 105: Curious as to your opinion on this, but I couldn't find a reason to keep the addr and url separate. Using the stdlib url.URL type, and using the access methods on it as needed.

pkg/proxy/proxy.go
```diff
@@ -2,13 +2,12 @@ package proxy
...
 import (
 	"bytes"
-	"context"
 	"io"
 )
 
 type Proxy interface {
-	AcceptRequest(uri string, body io.ReadCloser, ctx context.Context) (*bytes.Buffer, error)
-	DispatchRequest(input []byte) ([]byte, error)
+	AcceptRequest(contextValues map[string][]byte, requestURI []byte, body io.Reader, buff *bytes.Buffer) error
```
mvid, line 9: Updated the interface with the new signature you implemented

pkg/proxy/proxy.go
```diff
@@ -2,13 +2,12 @@ package proxy
...
+	DispatchRequest(buff *bytes.Buffer) (io.ReadCloser, error)
```
mvid, line 10: Changed this from returning http.Response, as that would tie this interface to the net/http lib

jensneuse:
Thanks for your contribution.

I've pushed a few minor changes to this branch:
https://github.com/jensneuse/graphql-go-tools/tree/mvid-proxy-update

Please review and comment here if you're okay with it.
If you feel the need for additional commits please PR on branch mvid-proxy-update.

If everything is OK I'll close this PR and merge mvid-proxy-update.

mvid:
Yes, seems good

PR #47, opened 2019-03-31
44 scaling parse type system definition

jensneuse:
This PR improves performance by reducing memory usage.
Instead of using a index slice per node to track child nodes (in hot places) I switched to a single linked list so the parent node has only one reference to the first child (the last child in this first simple implementation). All additional child nodes will know the previous child node (keep a reference) so that iterating all children is possible while having only a really tiny overhead.


jensneuse:
There's more work to be done but it's worth a merge at this stage (50% memory usage per operation).

Issue #66, opened 2019-04-17
ServeHTTP should handle DispatchRequest?


flexzuu:
The error is currently not returned neither logged so debugging when the backend URL is not reachable is hard.

https://github.com/jensneuse/graphql-go-tools/blob/58e708802492050485a4133458f0125ff452a06a/pkg/proxy/http/proxy.go#L123-L128

jensneuse:
Seems like this error is indeed swallowed. This should be fixed!

PR #72, opened 2019-04-24
Handle DispatchRequest error in a non-silent way

mvid:
In response to #66


jensneuse:
@mvid Could you add a test to make this bug fix safe to regressions?

mvid:
@jensneuse take a look now

PR #80, opened 2019-05-20
Failing empty list bug

mvid:
Also ran `make format`, so there might be other things in the PR


jensneuse:
This but was introduced by some logic that I implemented to de-duplicate list values. This de-duplication logic wasn't aware of empty lists which is now fixed.

PR #108, opened 2019-11-17
normalization of type extensions

Java-Jonas:
This PR will implement the normalization of the remaining type extensions.
It also contains a [commit](https://github.com/jensneuse/graphql-go-tools/commit/78fb9edba4ddf28f45661788f351e9f897fd99e0) which fixed some naming inconsistencies.

### DIFF
--- a/pkg/astnormalization/union_type_extending_test.go
+++ b/pkg/astnormalization/union_type_extending_test.go
@@ -3,13 +3,31 @@ package astnormalization
 import "testing"
 
 func TestExtendUnionType(t *testing.T) {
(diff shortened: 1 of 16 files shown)


pkg/astnormalization/union_type_extending_test.go
```diff
@@ -0,0 +1,33 @@
+package astnormalization
+
+import "testing"
+
+func TestExtendUnionType(t *testing.T) {
+	t.Run("extend union type by directive", func(t *testing.T) {
+		run(extendUnionTypeDefinition, testDefinition, `
+					union Mammal
```
jensneuse, line 8: please add one testcase which starts with a union containing union members which then gets extended

PR #166, opened 2020-01-17
feature/use http hanle funcs

jensneuse:
This PR is a small modification of the work of @spetrunin
Thanks for the contribution.
Merging should close #165 

@spetrunin please Review and give me feedback if you're ok with the modifications.

### DIFF
--- a/pkg/playground/playground.go
+++ b/pkg/playground/playground.go
@@ -44,19 +44,19 @@ type fileConfig struct {
 	fileContentType string
 }
 
-type HttpServeMux interface {
-	HandleFunc(pattern string, handler func(http.ResponseWriter, *http.Request))
-}
+type HandlersMap map[string]http.HandlerFunc
 
-func ConfigureHandlers(mux HttpServeMux, config Config) error {
+func ConfigureHandlers(config Config) (handlers HandlersMap, err error) {
+	handlers = make(HandlersMap)
+
 	box := packr.NewBox("./files")
 	playgroundHTML, err := box.FindString("playground.html")
 	if err != nil {
-		return err
+		return
 	}
 	templates, err := template.New(playgroundTemplate).Parse(playgroundHTML)
 	if err != nil {
-		return err
+		return
 	}
 
 	playgroundURL := path.Join(config.URLPrefix, config.PlaygroundURL)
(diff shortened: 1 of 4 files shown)


pkg/playground/playground.go
```diff
@@ -22,11 +22,16 @@ const (
 	contentTypeTextJavascript = "text/javascript"
 )
 
+// Config is the configuration Object to instruct ConfigureHandlers on how to setup all the http Handlers for the playground
 type Config struct {
+	// URLPrefix is a prefix you intend to put in front of all handlers
 	URLPrefix                   string
+	// PlaygroundURL is the URL where the playground website should be hosted
 	PlaygroundURL               string
```
devsergiy: I suppose `PlaygroundURL`, `GraphqlEndpointURL` is not a URLs, but a path
maybe name them as `PlaygroundPath`?

pkg/playground/playground.go
```diff
@@ -44,19 +49,36 @@ type fileConfig struct {
...
+
+func (h *Handlers) add(url string, handler http.HandlerFunc) {
+	*h = append(*h, HandlerConfig{
+		URL:     url,
+		Handler: handler,
+	})
+}
+
+// ConfigureHandlers takes your Config and sets up all handlers to the supplied Handlers slice
+func ConfigureHandlers(config Config, handlers *Handlers) (err error) {
```
devsergiy: Why do you think we have to supply handlers outside of this function?
Maybe replace this method with
```go
func NewPlaygroundHandlers(config Config) (handlers *Handlers, err error) {
}
```

pkg/playground/playground.go
```diff
@@ -102,22 +124,22 @@ func ConfigureHandlers(mux HttpServeMux, config Config) error {
...
 			return err
 		}
 	}
 
 	return nil
 }
 
-func configureFileHandler(mux HttpServeMux, box packr.Box, fileName, fileURL, contentType string) error {
-
+func configureFileHandler(handlers *Handlers, box packr.Box, fileName, fileURL, contentType string) error {
```
devsergiy: this could be a method of `Handlers`

pkg/playground/playground.go
devsergiy: another possible way
```go
type Playground struct {
  cfg Config
  box packr.Box
  files
}

func NewPlayground(config Config) (*Playground, err error) {
   return &Playground{
     cfg: config,
     box: packr.NewBox("./files"),
     files: ...
   }
}
func (h *Plaground) GetHandlers() (Handlers, error) {
}
func (h *Plaground) addFileHandler(fileName, fileURL, contentType string) error {
}

```

PR #176, opened 2020-01-23
Feature/implement union interface selection for httpjsondatasource

### LINKED ISSUE #171: Unions doesn't work for json data source
Unions are not working in case rest server doesn't supply type information via `__typename` field

jensneuse:
This PR adds the capability for the HTTPJsonDataSource to select a possible __typename based on the response http statusCode from the upstream.
Closes #171


jensneuse:
Hey @spetrunin please checkout this branch and test the feature, thanks.

PR #184, opened 2020-02-14
implement arg templating with custom directives

### LINKED ISSUE #156: Add ability to convert input type into a query args for json data source
At the moment we passing each query arg separately
```
    query(arg1: String, arg2: Int! = 1): Response!
    @HttpJsonDataSource(
        host: "localhost:8080"
        url: "/api/query?arg1={{ .arguments.arg1 }}&arg2={{ .arguments.arg2 }}"
    )
```
I think it would be more convinient to use query type for that and build query args from it
```
input QueryInput {
    arg1: String
    arg2: Int!
    
}
type Query {
   query(q: QueryInput!): Response!
    @HttpJsonDataSource(
        host: "http://localhost:8080"
        url: "/api/query?{{ URLEncode .arguments.q }}"
    )
}
```

For example, we could use fasthttp args builder
https://godoc.org/github.com/valyala/fasthttp#Args.QueryString

jensneuse:
This PR will give the library user the possibility to supply custom directives for argument resolution to the execution.Handler.
This closes #156

### DIFF
--- a/pkg/execution/handler.go
+++ b/pkg/execution/handler.go
@@ -9,6 +9,7 @@ import (
 	"github.com/cespare/xxhash"
 	"github.com/gobuffalo/packr"
 	"github.com/jensneuse/abstractlogger"
+	"github.com/jensneuse/byte-template"
 	"github.com/jensneuse/graphql-go-tools/pkg/ast"
 	"github.com/jensneuse/graphql-go-tools/pkg/astnormalization"
 	"github.com/jensneuse/graphql-go-tools/pkg/astparser"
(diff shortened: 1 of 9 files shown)


pkg/execution/handler.go
```diff
@@ -25,9 +26,10 @@ type Handler struct {
 	log                abstractlogger.Logger
 	definition         ast.Document
 	graphqlDefinitions *packr.Box
+	templateDirectives []byte_template.DirectiveDefinition
 }
 
-func NewHandler(schema []byte, logger abstractlogger.Logger) (*Handler, error) {
+func NewHandler(schema []byte, templateDirectives []byte_template.DirectiveDefinition, logger abstractlogger.Logger) (*Handler, error) {
```
devsergiy, line 32: do you think we should always supply directives?

PR #224, opened 2020-06-11
Fixes build for less than 64bit architectures

excieve:
On CPU architectures, where data units are less than 64bit wide builds currently fail with an overflow. This is due to implicit type in the for loop iterator `i` in `Document.GenerateUnusedVariableDefinitionName` function, which the compiler guesses as just `int`. This type is architecture dependent, so that on 32bit arch it's an `int32`, while on 64bit one it's an `int64`.

When setting the loop's upper bound at `math.MaxInt64` and trying to compile it with `GOARCH=386` variable `i` would be of `int32` type and thus cause an overflow (which is detected by the compiler). This change makes loop iterator types explicit as `int64`.

### DIFF
--- a/pkg/ast/ast_operation_definition.go
+++ b/pkg/ast/ast_operation_definition.go
@@ -66,17 +66,20 @@ const (
 )
 
 func (d *Document) GenerateUnusedVariableDefinitionName(operationDefinition int) []byte {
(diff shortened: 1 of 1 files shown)


excieve:
Oops, wrong repo :)

jensneuse:
Good catch!

PR #250, opened 2021-04-11
[Federation] Merging SDLs into base schema and creating engineV2 configuration from SDLs

chedom:
1. Merging SDLs into base schema
2. Creating engineV2 from SDLs (and base schema)
3. Adding example of creating and configuring federation gateway server

PR #251, opened 2021-04-13
Feature/add subscription to execution engine v2

chedom:
1. Add Subscription operation to execution engine V2


jensneuse:
Can you extend the example with a federation API so that it can be tested?
Here's an example with a subscription upstream you could use:
https://github.com/wundergraph/wundergraph-demo/tree/main/federation/services/products

chedom:
@jensneuse I have added test case for federation API.
Is that what you meant ?

jensneuse:
> @jensneuse I have added test case for federation API.
> Is that what you meant ?

I would like to manually test federated subscriptions using a gqlgen example subgraph so that you can open the playground and see it in action.

chedom:
Example has been added.

jensneuse:
> Example has been added.

When I merge all 3 branches together, I get a merge conflict.
Can you create a clean PR with all code together without conflicts?
I'd really appreciate it.

chedom:
> > Example has been added.
> 
> When I merge all 3 branches together, I get a merge conflict.
> Can you create a clean PR with all code together without conflicts?
> I'd really appreciate it.

Done. All code is in initial PR #250

chedom:
Changes will be  in #250

PR #252, opened 2021-04-13
fix code formatting

chedom:
1. Commit result of `make prepare-merge`


chedom:
Changes will be merged in #250

Issue #276, opened 2021-08-23
panic occurred when gateway access union types


vvakame:
Hi, We tried to use federation gateway for our PoC project.
We met the panic with union types.
I tried to investigate why this problems are happens. but I couldn't... 😿 

reproduction code.
https://github.com/vvakame/graphql-go-tools/tree/feat-panic

If this problem is solved, my use case may work, so I hope it will be fixed. 😸 

<details>

<summary>example schema</summary>

schema A

```
extend type Query {
    me: User
}

type User @key(fields: "id") {
    id: ID!
}
```

schema B

```
extend type User @key(fields: "id") {
    id: ID! @external
    histories(
        first: Int = 10
        after: String
    ):HistoryConnection!
}

type HistoryEdge {
    node: History!
    cursor: String
}

type HistoryConnection {
    edges: [HistoryEdge!]!
    nodes: [History!]!
}

union History =
    HistoryEntityA |
    HistoryEntityB

type HistoryEntityA {
    id: String
    textA: String
}

type HistoryEntityB {
    id: String
    textB: String
}
```

</details>

<details>

<summary>query that cause panic</summary>

```
{
  me {
    id
    histories(first: 10) {
      edges {
        node {
          ... on HistoryEntityB {
            id
            textB
          }
        }
      }
      nodes {
        ... on HistoryEntityB {
          id
          textB
        }
      }
    }
  }
}
```

```
2021/08/23 18:14:54 http: panic serving [::1]:54855: runtime error: index out of range [5] with length 5
goroutine 47 [running]:
net/http.(*conn).serve.func1(0xc0012f6140)
	/usr/local/Cellar/go/1.16.6/libexec/src/net/http/server.go:1804 +0x153
panic(0x170d140, 0xc0001553e0)
	/usr/local/Cellar/go/1.16.6/libexec/src/runtime/panic.go:971 +0x499
github.com/jensneuse/graphql-go-tools/pkg/ast.(*Document).AddSelection(...)
	/Users/vvakame/work/gopath/pkg/mod/github.com/jensneuse/graphql-go-tools@v1.20.2/pkg/ast/ast_selection.go:107
github.com/jensneuse/graphql-go-tools/pkg/engine/datasource/graphql_datasource.(*Planner).EnterInlineFragment(0xc0005c6d00, 0x1)
[...]

PR #283, opened 2021-08-28
fix Union and Interface type for federation

chedom:
Fix issue #276


spetrunin:
@chedom We already have a basic integration test here https://github.com/jensneuse/graphql-go-tools/blob/master/pkg/graphql/execution_engine_v2_test.go#L521

Do you think you need a separate one?

chedom:
> @chedom We already have a basic integration test here https://github.com/jensneuse/graphql-go-tools/blob/master/pkg/graphql/execution_engine_v2_test.go#L521
> 
> Do you think you need a separate one?

According to [comment](https://github.com/TykTechnologies/graphql-go-tools/pull/155#issuecomment-908270679) its required to add e2e tests for federation.
TestExecutionEngineV2_FederationAndSubscription_IntegrationTest skips important step - creating federation schema from SDLs.

spetrunin:
@chedom it is not a hard requirement it is only nice to have

Also, I believe we could use existing subgraphs from examples to not copy over them - this will make a code a bit simpler

chedom:
@spetrunin  I don't consider it as copying/duplicating.  In my opinion `examples` and `tests` exist for different purposes and, more importantly, they have different reason to change that's why I decided not to bind them.  But if you have a different opinion, I can redo.

PR #284, opened 2021-08-31
Feature/add operation middleware

chedom:
- add operation middleware, it allows to inject user defined code before execution of a query (it knows nothing about transport layer implementation). It will be useful for any operation that must be done before/after execution of the query. Possible use cases (at least I'll use it for that cases): query authorization (there is no need to implement separate http middleware and webscoket hook), tracing, logging. Middleware can break call chain (it will be useful for authorization).
- add example with authorization operation middleware


chedom:
@jensneuse conflicts have been merged.

csilvers:
@dnerdy I know you were thinking about how useful it would be for Khan Academy to have hooks around the upstream requests.  Does this API satisfy the needs that you have?

PR #285, opened 2021-09-10
Fix/context is missing in the arguments of subscription.NewExecutorV2Pool in the federation example

bcgodev:
fix(example): add context.Backgrount() to the arguments of subscription.NewExecutorV2Pool

It's a regression bug introduced int 83a5ad29 of #148.


bcgodev:
After running `go get -u` the federation package, it's fixed.

Issue #298, opened 2021-11-01
Unable to resolve an array of strings when it contains a stringified  number


chedom:
In case when an array of strings contains a stringified  number `resolveString` doesnt recognize an element as `jsonparser.String` (it recognizes element as jsonparser.Number).

You can confirms the problem by running test (sub test of TestResolver_ResolveNode):
```
t.Run("resolve array of strings", testFn(func(t *testing.T, r *Resolver, ctrl *gomock.Controller) (node Node, ctx Context, expectedOutput string) {
		return &Object{
			Fetch: &SingleFetch{
				BufferId:              0,
				DataSource:FakeDataSource(`{"strings": ["Alex", "true", "123"]}`),
			},
			Fields: []*Field{
				{
					BufferID: 0,
					HasBuffer: true,
					Name: []byte("strings"),
					Value: &Array{
						Path: []string{"strings"},
						Item: &String{
							Nullable: false,
						},
					},
				},
			},
		},Context{Context: context.Background()}, `{"strings":["Alex","true","123"]}`
	}))
```

Here is `resolveString` method:
```
func (r *Resolver) resolveString(str *String, data []byte, stringBuf *BufPair) error {
	var (
		value     []byte
		valueType jsonparser.ValueType
		err       error
	)
	if len(data) != 0 && str.Path == nil {
		_, valueType, _, _ = jsonparser.Get(data)
		if valueType == jsonparser.String || unicode.IsLetter(rune(data[0])) {
			value = data
		} else if !str.Nullable {
			return errNonNullableFieldValueIsNull
		} else {
			r.resolveNull(stringBuf.Data)
			return nil
		}
	}
	if value == nil {
		value, valueType, _, err = jsonparser.Get(data, str.Path...)
		if err != nil || valueType != jsonparser.String {
			if !str.Nullable {
				return errNonNullableFieldValueIsNull
			}
			r.resolveNull(stringBuf.Data)
			return nil
		}
	}
	if value == nil && !str.Nullable {
		return errNonNullableFieldValueIsNull
	}
	stringBuf.Data.WriteBytes(quote)
	stringBuf.Data.WriteBytes(value)
	stringBuf.Data.WriteBytes(quote)
	return nil
}
```

It seems as in case if `str.Path == nil` there is no need to check the type of the `value`

PR #299, opened 2021-11-01
fix resolving array of strings with stringified number

chedom:
Fix an issue #298 .

`jsonparser.Get` [strips](https://github.com/buger/jsonparser/blob/master/parser.go#L974) quotes from string value, in case if `jsonparser.Get` is called outside of `resolveString` method, there is no way to distinguish if value is stringified number ("123") or number (123).

### DIFF
--- a/pkg/engine/resolve/resolve.go
+++ b/pkg/engine/resolve/resolve.go
@@ -13,7 +13,6 @@ import (
 	"strconv"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


jensneuse:
seems legit, I'll verify once I'm done with the batch PR (probably in the next couple of days, I slightly improved it as it didn't cover all edge cases)

Issue #294, opened 2021-10-22
The EngineV2Configuration function is very slow for large schema


dnerdy:
Our service schemas are 1k-4k lines each, and producing an engine configuration for all services (around 35k lines total) takes about 15-20 seconds on a 2.7 GHz Quad-Core Intel Core i7.

The bottleneck appears to be the `baseSchema` method on `LocalTypeFieldExtractor` which prints and parses the schemas multiple times for each type. Engine construction time is important to us because we construct multiple engines and replace engines as schemas updated.

#293 is also related to the federation config factory. It could make sense to address both issues at the same time since node traversal needs to be updated for #293.

Would you be up for us optimizing construction of the engine config?

jensneuse:
@dnerdy How long does it take with your updated branch with the improvements? Is the processing time now acceptable?

dnerdy:
Yes! Constructing an engine configuration now takes around 60ms for our schema instead of 15-20 seconds, which is great.

PR #300, opened 2021-11-02
Fix bugs in and improve performance of LocalTypeFieldExtractor

### LINKED ISSUE #293: Federation config factory builds an invalid configuration when local type extensions are present in an SDL
Type extensions are often used to extend a federated type owned by another service, but that need not be the case!

It is perfectly valid to extend a locally owned type like this:
```
type User {
    name: String
}

extend type User {
    email: String!
}
```
Local type extensions can also appear before the corresponding type definition, e.g.
```
extend type User {
    email: String!
}

type User {
    name: String
}
```
Local type extensions are a useful way to organize related fields when a schema is spread across multiple `.graphql` files (our schemas has many such extensions).

Local type extensions confuse the federation config factory. Specifically the `EngineV2Configuration` function produces plan data sources with root and child nodes that don't contain a complete set of fields. This leads to the planner producing invalid downstream queries like `query { user } # user doesn't have a selection set!` when a query like `{ user { email } }` is processed.

dnerdy:
This PR fixes several bugs in the LocalTypeFieldExtractor GetAllNodes
method. It also drastically improves the performance of the code. Tests
and a benchmark for the updated code were added in #297. Details of the
node extraction algorithm are described inline.

See the updated and uncommented tests for details about the fixed bugs
(and well as the fixed issues linked below).

Since this is a rewrite, the diff for the local_type_field_extractor.go file isn't
very useful. I recommend reviewing the file as if it was entirely new. The
tests are meant to demonstrate that the behavior of the existing code is
preserved.

The benchmark for the updated code is as follows:
```
$ cd pkg/engine/plan
$ go test -bench=. -run=xxxx
goos: darwin
goarch: amd64
pkg: github.com/jensneuse/graphql-go-tools/pkg/engine/plan
BenchmarkGetAllNodes-8  53476  22360 ns/op  10804 B/op  173 allocs/op
PASS
ok  github.com/jensneuse/graphql-go-tools/pkg/engine/plan  1.872s
```

This is a 500x speedup over the previous implementation (the prior
benchmark results can be found in #297).

Fixes #293 and #294.


pkg/engine/plan/local_type_field_extractor.go
```diff
@@ -27,173 +27,263 @@ func NewLocalTypeFieldExtractor(document *ast.Document) *LocalTypeFieldExtractor
...
-		return
-	}
-
-	fieldsRefs := e.document.NodeFieldDefinitions(node)
+		nodeInfo.isInterface = isInterface
+		// A node may be a local extension of a root node. The node is
+		// considered a root node if ANY node related to the type is a root
+		// node.
+		nodeInfo.isRoot = nodeInfo.isRoot || isRootNode
+		nodeInfo.concreteTypeNames = concreteTypeNames
```
csilvers: I'm not really knowledgeable about any of the semantics here, but was excited to read this diff.  I was wondering: do you need to do a set-union here like you do on line 149?  I don't know what the semantics of unionMemberRefs is in the presence of union extensions.

jensneuse:
Can you please merge master, I'll then take a look.

dnerdy:
I've rebased the changes on upstream master (e307e557). No changes to the content of the PR were made.

pkg/engine/plan/local_type_field_extractor.go
dnerdy: Yes, you're right. Here's the spec: https://spec.graphql.org/October2021/#sec-Union-Extensions. Union extensions DO NOT reiterate existing union members and can add new members. (This means a union is needed, but a set-union isn't necessary.) I'll update this code to collect the member types instead.

dnerdy: There aren't many examples of union extensions. Here are a couple from the Apollo federation code base:
- https://github.com/apollographql/federation/blob/c7907dcecbb308565ca1ff370d70cd4ec1649c9d/internals-js/src/__tests__/definitions.test.ts#L494
- https://github.com/apollographql/federation/blob/c7907dcecbb308565ca1ff370d70cd4ec1649c9d/federation-js/src/composition/validate/postComposition/__tests__/providesFieldsSelectInvalidType.test.ts#L182

jensneuse:
@dnerdy I'm sorry, this is now on top of my list. Can you rebase it again. -_-

pvormste:
I have refactored and merged this PR here: https://github.com/TykTechnologies/graphql-go-tools/pull/177
We will now need backport it here @jensneuse

Issue #303, opened 2021-11-04
Graphql datasource planner builds wrong ContextVariable when input variable is Array


chedom:
Graphql datasource planner builds wrong ContextVariable when input variable is Array.
`Document.ResolveUnderlyingType` [skips](https://github.com/jensneuse/graphql-go-tools/blob/master/pkg/ast/ast_type.go#L205)  `TypeKindList` types, therefore, [this](https://github.com/jensneuse/graphql-go-tools/blob/master/pkg/engine/resolve/resolve.go#L1609) condition will never be met.

Here is a [test](https://github.com/chedom/graphql-go-tools/blob/test%2Farray-context-variables/pkg/graphql/execution_engine_v2_test.go#L448) that confirms the issue (there is `"` before `[` and after `]`).

```
-{"query":"query($heroNames: [String!]!){heroes(names: $heroNames)}","variables":{"heroNames":["Luke Skywalker","R2-D2"]}}
+{"query":"query($heroNames: [String!]!){heroes(names: $heroNames)}","variables":{"heroNames":"["Luke Skywalker","R2-D2"]"}}
```

PR #304, opened 2021-11-04
Fix input Array variables  

chedom:
Fix issue #303


pkg/ast/ast_type.go
```diff
@@ -209,3 +209,14 @@ func (d *Document) ResolveUnderlyingType(ref int) (typeRef int) {
 	}
 	return
 }
+
+func (d *Document) ResolveListOrNameType(ref int) (typeRef int) {
```
dnerdy, line 213: This method clicked for me once I realized there are three types--TypeKindNamed, TypeKindList and TypeKindNonNull--and what the method is doing is unwrapping the non-null. Since non-nulls shouldn't be nested, I think this could be a little simpler (unlike ResolveUnderlyingType above which needs to handle arbitrarily nested lists). Like so:
```
graphqlType := d.Types[ref]
if graphqlType.TypeKind == TypeKindNonNull {
    return graphqlType.OfType
}
return ref
```
This also makes me think that perhaps the method should be named something UnwrapNonNull, though that would break with the "Resolve" naming.

pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
```diff
@@ -250,6 +250,280 @@ func TestGraphQLDataSource(t *testing.T) {
...
+				}
+			}
+		}
+	`, "", &plan.SynchronousResponsePlan{
+		Response: &resolve.GraphQLResponse{
+			Data: &resolve.Object{
+				Fetch: &resolve.SingleFetch{
+					DataSource: &Source{},
+					BufferId:   0,
+					Input:      `{"method":"POST","url":"https://subgraph-reviews/query","body":{"query":"query($representations: [_Any!]!){_entities(representations: $representations){__typename ... on Product {reviews {body author {username id}}}}}","variables":{"representations":$$0$$}}}`,
```
dnerdy, line 273: For a moment I thought that existing entity queries might somehow be broken, but that didn't seem right. This test appears to be testing a scenario where the engine itself receives an entity query. I don't think this is realistic since the [federation spec](https://www.apollographql.com/docs/federation/federation-spec/#query_service) indicates that federation modifications SHOULD NOT be included in service SDLs. That means _entites and the related types would never appear in a plan data source configuration.

I think this also means the _Any scalar wouldn't necessarily be the federation _Any type (though I suppose if someone defined a type with the same name but with different semantics that would conflict with federation and be really confusing... and why would someone define a custom underscore type anyway?).

dnerdy (commented):
Nice work finding these edge cases!

This PR appears to be addressing two issues. The first is the incorrect `JsonValueType` due to the use of ResolveUnderlyingType. The second is the incorrect `ArrayJsonValueType` value.

I looked into how `ArrayJsonValueType` is used, and it appears to be used for [rendering arrays as CSVs](https://github.com/jensneuse/graphql-go-tools/blob/e307e557cab98679e6b7ddcde6e07bd6ab20050d/pkg/engine/resolve/resolve.go#L1321) in query templates.

It isn't clear to me how non-scalar values should be rendered as CSVs. Also, an issue with the `JsonValueType` and `ArrayJsonValueType` construct is it can't represent arbitrarily nested lists, though I don't know if nested lists make any sense in this context. It seems like there should maybe be validation for datasources that use CSV rendering to ensure that input types are always either scalars or arrays of scalars. At the moment it looks like the CSV code will just skip values if the value in the array doesn't match the ArrayJsonValueType, and I'm not sure it outputs anything sensical for object or list types when the value does match.

What do you think about removing the ArrayJsonValueType changes from this PR and addressing the CSV/ArrayJsonValueType issues in another change? (I also left a comment inline about how the _Any type might not necessarily need special handling in this context.)

pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
jensneuse: We actually have a use case for this. If all subgraphs are a managed API themselves, we use the engine as a "proxy" to route traffic to the subgraph. In this scenario, we have one + N engines. One represents the supergraph, N the subgraphs. This way, you can manage supergraphs and subgraphs independently.

jensneuse: Adding "_Any" is a good shortcut until we find a better solution I guess, e.g. allowing a datasource to add their own scalars or schema extensions, including validation hooks..

pkg/ast/ast_type.go
jensneuse: I like UnwrapNonNullType and I think you're right with the simpler solution. That said, doesn't make much difference.

jensneuse:
> Nice work finding these edge cases!
> 
> This PR appears to be addressing two issues. The first is the incorrect `JsonValueType` due to the use of ResolveUnderlyingType. The second is the incorrect `ArrayJsonValueType` value.
> 
> I looked into how `ArrayJsonValueType` is used, and it appears to be used for [rendering arrays as CSVs](https://github.com/jensneuse/graphql-go-tools/blob/e307e557cab98679e6b7ddcde6e07bd6ab20050d/pkg/engine/resolve/resolve.go#L1321) in query templates.
> 
> It isn't clear to me how non-scalar values should be rendered as CSVs. Also, an issue with the `JsonValueType` and `ArrayJsonValueType` construct is it can't represent arbitrarily nested lists, though I don't know if nested lists make any sense in this context. It seems like there should maybe be validation for datasources that use CSV rendering to ensure that input types are always either scalars or arrays of scalars. At the moment it looks like the CSV code will just skip values if the value in the array doesn't match the ArrayJsonValueType, and I'm not sure it outputs anything sensical for object or list types when the value does match.
> 
> What do you think about removing the ArrayJsonValueType changes from this PR and addressing the CSV/ArrayJsonValueType issues in another change? (I also left a comment inline about how the _Any type might not necessarily need special handling in this context.)

I actually want to get rid of JsonValueType and ArrayJsonValueType entirely as they are incomplete.
Instead, I'd like to refactor it so that we can simply have one single (nested) JSON-Schema definition for the variables. Could be a single scalar, object or nested array or a mix, JSON-Schema can represent all of them in a clean way.

PR #307, opened 2021-11-05
add supporting multiple key directives

chedom:
Add supporting multiple key directives.
According to [federation-spec](https://www.apollographql.com/docs/federation/federation-spec/#key) multiple keys can be defined on a single object type:
```
type Product @key(fields: "upc") @key(fields: "sku") {
  upc: UPC!
  sku: SKU!
  name: String
}

```


jensneuse:
I vaguely remember that the value could also be a  selection set, e.g. `@key(fields: "foo bar { baz }")`.
Not sure if this is still valid but it could also be possible to treat the string as a selection set, parse all fields, maybe even nested and use them as keys. What do you think?

chedom:
@jensneuse  you are right, it's called a [compound](https://www.apollographql.com/docs/federation/entities/#defining-a-compound-primary-key) key.
Compound key could be addressed in separate PR.
A lack of multiple key blocks my work as [graphql-transform-federation](https://github.com/0xR/graphql-transform-federation) build federated schema with multiple key (`@key(fields: "prop1") @key(fields: "prop2")` instead of `@key(fields: "prop1 prop2")`).

jensneuse:
Hey @chedom, can you please fix the merge conflicts. I'll then add this PR.

chedom:
@jensneuse  conflicts have been fixed.

PR #315, opened 2021-12-29
Include `@requires` fields when constructing entity representations

### LINKED ISSUE #314: `@requires` fields not included in entity representations
graphq-go-tools doesn't not include `@requires` fields in representations when making entity queries.

Consider the following schema:
```
# service one

extend type Query {
    serviceOne: ServiceOneType
}

type ServiceOneType @key(fields: "id"){
    id: ID!
    serviceOneFieldOne: String!
    serviceOneFieldTwo: String!
}

# service two

extend type ServiceOneType @key(fields: "id") {
    id: ID! @external
    serviceOneFieldOne: String! @external
    serviceOneFieldTwo: String! @external
    serviceTwoFieldOne: String! @requires(fields: "serviceOneFieldOne")
    serviceTwoFieldTwo: String! @requires(fields: "serviceOneFieldTwo")
}
```
When performing the following query:
```
{
    serviceOne {
        serviceTwoFieldOne
        serviceTwoFieldTwo
    }
}
```
the following upstream queries are expected:
```
# service one
{"query": "{ serviceOne { id serviceOneFieldOne serviceOneFieldTwo } }"}

#service two
{
    "query": "query($representations: [_Any!]!) { _entities(representations: $representations) {... on ServiceOneType { serviceTwoFieldOne serviceTwoFieldTwo } } }",
    "variables": {
        "representations": [
            {"id": 1, "serviceOneFieldOne": "someValue", "serviceOneFieldTwo": "someOtherValue"}
        ]
    }
}
```
This isn't what graphql-go-tools sends. Instead, it sends representations that don't include the required fields, i.e. `{"representations": [{"id": 1}]}`.

See the attached file for a failing test case. [testcase.txt](https://github.com/jensneuse/graphql-go-tools/files/7786893/testcase.txt)

Also, while investigating this issue I noticed that the Planner.handleFederation function adds values for each primary key to representations.

This doesn't align with the [GraphQL federation specification](https://www.apollographql.com/docs/federation/entities/#referencing). The specification states that keys for types extended by a service "must match the value of exactly one @key defined in the entity's originating subgraph".

I mention this because the representation logic will need to be updated to fix the `@requires` issue, and the tests in graphql_datastore_test.go related to multiple federated keys should probably be updated to align with the spec as well, e.g. https://github.com/jensneuse/graphql-go-tools/blob/master/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go#L2744 should not include "name" (and it isn't valid for the product service schema to specify "name" as an additional key).

dnerdy:
This PR fixes #314.

graphql-go-tools currently doesn't include `@requires` fields in entity
representations. This PR updates the GraphQL datasource planner to
include `@requires` fields when constructing representations for an
upstream operation.

The representations logic has been updated to use RequiresFields from
plan.FieldConfiguration. This field already includes the necessary
federation keys and required fields.

Due to the order the Planner visitor calls LeaveDocument, the
representations are updated each time a field is visited instead of
being set once during LeaveDocument. This is because ConfigureFetch is
called before LeaveDocument. It would be possible to move the code to
LeaveDocument if the visit order was reversed for the Leave* methods,
but that is a large change so it wasn't made in this PR.

Note that the use of RequiresFields also fixes a bug where
graphql-go-tools didn't align with the federation spec with respect to
federation keys on extended types. The existing representations code
includes all the keys defined on a type, but the spec states that
exactly one key must match in a subgraph[1]. The RequiresFields logic
aligns with this requirement. The GraphQL datasource tests have been
updated accordingly.

[1] https://www.apollographql.com/docs/federation/entities/#referencing

### DIFF
--- a/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
+++ b/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"strings"
 
 	"github.com/jensneuse/graphql-go-tools/pkg/asttransform"
 	"github.com/tidwall/sjson"
@@ -30,6 +29,7 @@ type Planner struct {
 	config                             Configuration
 	upstreamOperation                  *ast.Document
 	upstreamVariables                  []byte
+	representationsJson                []byte
 	nodes                              []ast.Node
 	variables                          resolve.Variables
 	lastFieldEnclosingTypeName         string
@@ -86,7 +86,6 @@ func (p *Planner) addDirectiveToNode(directiveRef int, node ast.Node) {
 }
 
 func (p *Planner) DownstreamResponseFieldAlias(downstreamFieldRef int) (alias string, exists bool) {
-
 	// If there's no alias but the downstream Query re-uses the same path on different root fields,
 	// we rewrite the downstream Query using an alias so that we can have an aliased Query to the upstream
 	// while keeping a non aliased Query to the downstream but with a path rewrite on an existing root field.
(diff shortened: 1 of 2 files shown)


jensneuse:
Really nice contribution! Will look into it in Jan but from the description I think we want to merge it.

jensneuse:
Thanks again @dnerdy, great work!

PR #317, opened 2022-01-14
Fix federation directive variables

dnerdy:
Fixes jensneuse/graphql-go-tools#316.

When an entity query contains a field with a directive that includes a
variable, the variable isn't added to constructed upstream operation.
This PR makes it so variables used with directives are added as needed.

### DIFF
--- a/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
+++ b/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
@@ -81,8 +81,82 @@ func (p *Planner) addDirectiveToNode(directiveRef int, node ast.Node) {
 	if p.upstreamDefinition != nil && !p.upstreamDefinition.DirectiveIsAllowedOnNodeKind(upstreamDirectiveName, node.Kind, operationType) {
 		return
 	}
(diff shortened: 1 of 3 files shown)


jensneuse:
Great work! Very much impressed that you understand all the low level concepts of the library to solve this problem.

dnerdy:
I've pulled this change into our service, and I'm seeing a panic when printing the upstream operation. The problematic query contains several directives with variable arguments, so that could be the difference (compared to the test). I'll investigate further and report back.

dnerdy:
Ah, so close! The issue was I wasn't importing the variable into the upstream operation before adding it to the operation definition. I variable ref from the planner operation was being used. Apparently the refs just happened to match. Do you have any suggestions about how to suss out such mistakes in the tests?

jensneuse:
> Ah, so close! The issue was I wasn't importing the variable into the upstream operation before adding it to the operation definition. The variable ref from the planner operation was being used. Apparently the refs just happened to match. Do you have any suggestions about how to suss out such mistakes in the tests?

We have been thinking about this as well. We think one solution could be to close down the AST package and add safe helper methods so that you cannot add a variable from outside of it doesn't exist on the AST etc.. what do you think? On the tests, I guess an end to end test with planning and execution with a fake upstream can catch this.

jensneuse:
Thanks, @dnerdy for your contribution, great work!

PR #327, opened 2022-01-27
Fix data loader heterogeneous types

### LINKED ISSUE #326: Data loader should skip inputs that don't contain the field being loaded.
When a query contains an interface field that contains an array of heterogeneous types, the data loader returns the error "Key path not found". This is happening here: https://github.com/jensneuse/graphql-go-tools/blob/47621af0e2831f40171c681f77777da2ead0eefa/pkg/engine/resolve/dataloader.go#L367

Consider the following query:
```
{
  someArrayWithInterfaceItem {
    ... on A {
      aField {
        id
        fieldFromAnotherService  # <- this is federated
      }
    }
    ... on B {
      someOtherField
    }
  }
}
```
The result after fetching someArrayWithInterfaceItem might be:
```
{
  "data": {
    "someArrayWithInterfaceItem": [
      {"__typename": "A", "aField": {"id": 1}},
      {"__typename": "B", "someOtherField": "hello"},
      {"__typename": "A", "aField": {"id": 2}}
    ]
}
```
In this case, the data loader currently attempts to get the "aField" key from all the "someArrayWithInterfaceItem" objects, which results in the error.

dnerdy:
This PR fixes #326.

See the in-code comment for a description of when this issue occurs.

I also fixed up some typos in the data loader tests when "name" was being selected instead of "username".

### DIFF
--- a/pkg/engine/datasource/graphql_datasource/batch.go
+++ b/pkg/engine/datasource/graphql_datasource/batch.go
@@ -102,7 +102,6 @@ func (b *BatchFactory) multiplexBatch(out *fastbuffer.FastBuffer, inputs [][]byt
 		}
 
 		_, err = jsonparser.ArrayEach(inputVariables, func(value []byte, dataType jsonparser.ValueType, offset int, err error) {
-
 			for j := range responseMappings {
 				existing := responseMappings[j].originalInput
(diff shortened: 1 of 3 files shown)


csilvers:
I just rebased on upstream master, so this should be mergeable now.

PR #329, opened 2022-01-27
Avoid printing the same variable twice for a query.

csilvers:
Before this change, if a variable was used in a query both at the top
level and inside a recursive structure, it would be included twice in
the query:
```
query($name: String!, $name: String!){user(name: $name){normalized(data: {name: $name})
```

After this change, it only prints it once.  I think the previous code
location was just a mistake.


csilvers:
It looks like this is already in master, at https://github.com/jensneuse/graphql-go-tools/commits/27e533ea7f9fcb7aac1556e2d7b14fdbd549e9c8

PR #330, opened 2022-01-28
Fix the json variable parser to properly handle null and non-null fields

csilvers:
The previous logic was saying that null fields could be omitted from
the input, but not that they could be set explicitly to null.  Now
both are allowed, as per the graphql spec.

For example, if the input is `$a: String`, it is now legal to pass any of:
```
{"a": "some string"}  // previously allowed
{}                    // previously allowed
{"a": null}           // newly allowed
```

The new tests turned up a few other bugs, most notably misspelling the
jsonschema keyword `items` as `item`.


pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
```diff
@@ -2457,15 +2457,15 @@ func TestGraphQLDataSource(t *testing.T) {
...
 							},
 							&resolve.ContextVariable{
 								Path:     []string{"completed"},
-								Renderer: resolve.NewJSONVariableRendererWithValidation(`{"type":"boolean"}`),
+								Renderer: resolve.NewJSONVariableRendererWithValidation(`{"type":["boolean"]}`),
 							},
 							&resolve.ContextVariable{
 								Path:     []string{"name"},
-								Renderer: resolve.NewJSONVariableRendererWithValidation(`{"type":"string"}`),
+								Renderer: resolve.NewJSONVariableRendererWithValidation(`{"type":["string","null"]}`),
```
csilvers, line 3168: @jensneuse this is one of the places I was confused on what the code was supposed to be doing.  I made the test match the current functionality, but maybe it's the functionality that needs to be changed:

The issue here is that the AddTask mutation says that `name` is a non-nullable string.  But `$name` is _used_ in `input:[{user:{name: $name}}]`, and AddTaskInput.UserRef.name is a nullable string.  The code that constructs the renderer (`resolveNestedArgumentType`) is looking for the type of the input-object and getting the nullable string.

I suspect that this really should be a non-nullable string, and the logic in `resolveNestedArgumentType` is not actually right in this case.  But I don't really understand what's going on here.  I'm hoping you do, and know what the right fix is! (If indeed something needs to be fixed.)

csilvers: (There are a few other tests that have the same issue, but this one is as clear as any of them.)

pkg/engine/datasource/rest_datasource/rest_datasource_test.go
```diff
@@ -259,11 +259,11 @@ func TestFastHttpJsonDataSourcePlanning(t *testing.T) {
 						Variables: resolve.NewVariables(
 							&resolve.ContextVariable{
 								Path:     []string{"idVariable"},
-								Renderer: resolve.NewPlainVariableRendererWithValidation(`{"type":"string"}`),
+								Renderer: resolve.NewPlainVariableRendererWithValidation(`{"type":["string"]}`),
 							},
 							&resolve.ContextVariable{
 								Path:     []string{"a"},
-								Renderer: resolve.NewPlainVariableRendererWithValidation(`{"type":"string"}`),
+								Renderer: resolve.NewPlainVariableRendererWithValidation(`{"type":["string","null"]}`),
```
csilvers, line 266: @jensneuse I couldn't figure out how nullability is determined for rest parameters.  It seems like the last param is nullable and the other ones aren't?  I modified all the tests with that rule in mind, in any case.

pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -771,7 +771,7 @@ func (p *Planner) resolveNestedArgumentType(fieldName []byte) (fieldTypeRef int)
 
 	for _, inputFieldDefRef := range p.visitor.Definition.InputObjectTypeDefinitions[argTypeNode.Ref].InputFieldsDefinition.Refs {
 		if bytes.Equal(p.visitor.Definition.InputValueDefinitionNameBytes(inputFieldDefRef), fieldName) {
-			return p.visitor.Definition.ResolveListOrNameType(p.visitor.Definition.InputValueDefinitions[inputFieldDefRef].Type)
+			return p.visitor.Definition.InputValueDefinitions[inputFieldDefRef].Type
```
csilvers, line 779: @jensneuse I wanted to point out this change in particular.  The old code was "looking through" the non-null type, which I think was wrong.  All tests pass with this taken out, but I don't really understand why it was in there in the first place, so I'm not 100% confident in this change.

PR #334, opened 2022-02-03
Implement federation of abstract fields

### LINKED ISSUE #328: Impelement federation of interface fields
This is a big one. If a field has an interface type and the concrete types are federated, the query planner either ignores the field or generates an upstream operation that isn't valid.

There appear to be (at least) three areas that need to be updated:
- The EnterField method in the configuration visitor
- The GraphQL data source code that generates the representations variable
- The GraphQL data source code that adds fields to upstream operations

The EnterField method -- non-child interface fields are skipped during query planning. This is because interface types aren't listed in root nodes; entity queries must be on concrete types (and therefore interfaces can't be entry points). (Note that interface fields owned by the same service work fine because interfaces appear in the list of child nodes.) This will be tricky. See "Interfaces and root fields" below.

Representations -- `__typename` in the ["representations" variable in the upstream operation is static](https://github.com/jensneuse/graphql-go-tools/blob/47621af0e2831f40171c681f77777da2ead0eefa/pkg/engine/datasource/graphql_datasource/graphql_datasource.go#L540). `__typename` needs to instead be set to the type of the input object.

Upstream operation -- the inline fragment in an upstream operation is assumed to have [same type as the parent object](https://github.com/jensneuse/graphql-go-tools/blob/47621af0e2831f40171c681f77777da2ead0eefa/pkg/engine/datasource/graphql_datasource/graphql_datasource.go#L581) of the enclosing field. This isn't the case for interfaces! There needs to be a separate inline fragment for each concrete type that implements the interface. Note this also means that [when fields are added](https://github.com/jensneuse/graphql-go-tools/blob/47621af0e2831f40171c681f77777da2ead0eefa/pkg/engine/datasource/graphql_datasource/graphql_datasource.go#L1126) they need to be added to each of the inline fragments. Furthermore, if the selection in the original operation contains inline fragments, the fields in these fragments need to be added to the appropriate concrete inline fragments.

Interfaces and root fields:

Although this doesn't align exactly with the spec, let's make the assumption that the same field on all the concrete types that implement an interface are resolved by the same service. For example, if the "id" field of  `type Dog implements Pet` is resolved by the "pets" data source, then the "id" field of `type Cat implements Pet` is resolved by the "pets" data source as well. The federation spec is a bit vague about how federated interfaces work; it isn't clear how one would construct a set of schemas where the above wasn't true (where would the `@external` directive on the interface go?), so this solution seems good enough--or a least a good first step. It's certainly better than the current situation where federated interface fields don't work at all.

How would this work?

During planning (in the [configurationVisitor EnterField method](https://github.com/jensneuse/graphql-go-tools/blob/47621af0e2831f40171c681f77777da2ead0eefa/pkg/engine/plan/plan.go#L1400)), if typeName is an interface and a planner isn't found with a child node for the typeName, fieldName pair, look up the first concrete type for the interface and set typeName to that. Then, a planner will be constructed based on the service that owns the field of the concrete type.

dnerdy:
Fixes #328.

(Note: this change looks huge, but approximately 1800 of the lines are
tests and 150 of the lines are comments).

This PR implements federation of abstract fields. Specifically, it adds
support for the situation where a field with an interface or union type
has concrete types that are owned by another service (or at least have
selected fields owned by another service).

Take, for example, the query:
```
query {
  user {
    username
    pets {  # interface Pet
      name
      ... on Dog {
        dogField
      }
      ... on Cat {
        catField
      }
    }
  }
}
```

In this example "user" and "pets" are both resolved by the "user"
service and "name", "dogField" and "catField" are resolved by the "pet"
service. The issue is that the "pets" field is owned by one service but
the fields of the Pet type are owned by another service.

The major changes are:
- During planning, the first concrete type of an interface is used to make planning decisions. (This requires that a field with a given name is owned by the same service for all concrete types.)
- Also during planning, the first concrete type of an interface is used for field configuration. The same requirement as above applies.
- The federation root is added to the upstream opertion the first time an inline fragment, selection set or field is visited below the planner parent path.
- A "__typename" selection is always added to interface field selection sets, it case it's needed for federation (this isn't necessary for unions because union selection sets have explicit inline fragments).
- The "__typename" variable is now dynamic in representations when the root type is abstract.
- Multiple inline fragments are added to the federation root, one for each concrete type (this only applies to interface, since unions come with their own inline fragments).
- All sub-selections are added to the concrete inline fragments, and then fragments with invalid type conditions or empty fragments are pruned during upstream operation normalization.

### DIFF
--- a/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
+++ b/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
@@ -283,7 +283,7 @@ func (p *Planner) ConfigureFetch() plan.FetchConfiguration {
 			ExtractGraphqlResponse:    true,
 			ExtractFederationEntities: p.extractEntities,
 		},
-		BatchConfig:      batchConfig,
+		BatchConfig: batchConfig,
 	}
 }
 
@@ -323,6 +323,10 @@ func (p *Planner) LeaveOperationDefinition(ref int) {
 }
 
 func (p *Planner) EnterSelectionSet(ref int) {
+	// If this is the first node below the parent path being visited by this
+	// planner, set up federation if needed.
+	p.handleFederation(nil)
+
 	parent := p.nodes[len(p.nodes)-1]
 	set := p.upstreamOperation.AddSelectionSet()
 	switch parent.Kind {
@@ -335,21 +339,39 @@ func (p *Planner) EnterSelectionSet(ref int) {
 	case ast.NodeKindInlineFragment:
 		p.upstreamOperation.InlineFragments[parent.Ref].HasSelections = true
 		p.upstreamOperation.InlineFragments[parent.Ref].SelectionSet = set.Ref
+	default:
+		panic(fmt.Sprintf("unexpected parent kind: %v", parent.Kind))
 	}
 	p.nodes = append(p.nodes, set)
-	for _, selectionRef := range p.visitor.Operation.SelectionSets[ref].SelectionRefs {
-		if p.visitor.Operation.Selections[selectionRef].Kind == ast.SelectionKindField {
-			if p.visitor.Operation.FieldNameUnsafeString(p.visitor.Operation.Selections[selectionRef].Ref) == "__typename" {
-				field := p.upstreamOperation.AddField(ast.Field{
-					Name: p.upstreamOperation.Input.AppendInputString("__typename"),
-				})
-				p.upstreamOperation.AddSelection(set.Ref, ast.Selection{
-					Ref:  field.Ref,
-					Kind: ast.SelectionKindField,
-				})
+
+	switch p.visitor.Walker.EnclosingTypeDefinition.Kind {
+	case ast.NodeKindInterfaceTypeDefinition, ast.NodeKindUnionTypeDefinition:
+		// Always include __typename in abstract type selection sets. This is
+		// done because child fields may be federated and __typename will be
+		// needed for representations. While it would be possible to determine
+		// exactly when __typename is needed, there's no harm in just always
+		// including it.
+		p.addTypenameToSelectionSet(set.Ref)
+	default:
+		// Otherwise, include __typename if explicitly requested.
+		for _, selectionRef := range p.visitor.Operation.SelectionSets[ref].SelectionRefs {
+			if p.visitor.Operation.Selections[selectionRef].Kind == ast.SelectionKindField {
+				if p.visitor.Operation.FieldNameUnsafeString(p.visitor.Operation.Selections[selectionRef].Ref) == "__typename" {
+					p.addTypenameToSelectionSet(set.Ref)
+				}
 			}
 		}
 	}
+}
+
+func (p *Planner) addTypenameToSelectionSet(selectionSet int) {
+	field := p.upstreamOperation.AddField(ast.Field{
+		Name: p.upstreamOperation.Input.AppendInputString("__typename"),
+	})
+	p.upstreamOperation.AddSelection(selectionSet, ast.Selection{
+		Ref:  field.Ref,
+		Kind: ast.SelectionKindField,
+	})
 }
 
 func (p *Planner) LeaveSelectionSet(ref int) {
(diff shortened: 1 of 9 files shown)


pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -411,20 +433,23 @@ func (p *Planner) EnterField(ref int) {
...
 	p.handleFederation(fieldConfiguration)
+
+	// Then add the field.
 	p.addField(ref)
 
 	upstreamFieldRef := p.nodes[len(p.nodes)-1].Ref
 
+	if fieldConfiguration == nil {
+		return
+	}
```
csilvers, line 451: It makes no difference, but this could go above line 447.

pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -527,19 +552,40 @@ func (p *Planner) updateRepresentationsVariable(fieldConfig *plan.FieldConfigura
 		return
 	}
 
+	if len(p.representationsJson) == 0 {
+		isAbstract := false
+		switch p.visitor.Walker.LastFieldTypeDefinition.Kind {
```
csilvers, line 557: I don't know if you need this logic anywhere else, but it may make sense to have an `IsAbstract` method on `Kind`.

pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -576,22 +622,55 @@ func (p *Planner) fieldDefinition(fieldName, typeName string) *ast.FieldDefiniti
...
-			Type: typeRef,
-		},
-	})
-	p.upstreamOperation.AddSelection(p.nodes[len(p.nodes)-1].Ref, ast.Selection{
-		Kind: ast.SelectionKindInlineFragment,
-		Ref:  inlineFragment,
-	})
-	p.nodes = append(p.nodes, selectionSet)
+
+	p.addTypenameToSelectionSet(p.nodes[len(p.nodes)-1].Ref)
```
csilvers, line 657: Do you need to do this always? or only for abstract types?

pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -1068,6 +1147,7 @@ func (p *Planner) normalizeOperation(operation, definition *ast.Document, report
 		astnormalization.WithExtractVariables(),
 		astnormalization.WithRemoveFragmentDefinitions(),
 		astnormalization.WithRemoveUnusedVariables(),
+		astnormalization.WithDeleteInvalidInlineFragments(),
```
csilvers, line 1150: Do you want a comment here pointing to your comment on lines 647-654?

pkg/engine/plan/plan.go
```diff
@@ -402,6 +418,79 @@ func (v *Visitor) AllowVisitor(kind astvisitor.VisitorKind, ref int, visitor int
 		return true
 	}
 	for _, config := range v.planners {
+		// In the following query:
```
csilvers, line 421: This is a super-detailed comment!  I'm still not sure I entirely understand it.  (I wonder if it would benefit from saying *why* the pets planner needs to iterate as deep as it does.)  But the length of the comment itself is a useful signal that something tricky is going on. :-)

csilvers (commented):
Obviously I'm not qualified to review this, but I read it with great interest.  Amazing how (relatively) little work was required to implement all this!  I didn't look at the tests except to see how extensive they were; hopefully they've captured all the corner cases!

csilvers:
(Also, I noticed there are conflicts; you might want to pull from upstream-master into the khan repo and resolve those.)

dnerdy:
I'm going to pursue an approach that uses operation normalization. There are edge cases this approach doesn't handle, and there are other bug with fixes that would also benefit from further normalization. For additional details see https://github.com/jensneuse/graphql-go-tools/issues/328#issuecomment-1030296919.

PR #336, opened 2022-02-05
Implement copying of query document AST nodes

dnerdy:
This PR implements copying of AST nodes used in query documents. Copying
will be used when normalizing operations for federation. Selection sets
will be copied when "expanding" inline fragments and inserting inline
fragments for fields with interface types. See the plan in #328 for
details.

I'm not sure everything that's being copied needs to be copied. For
example, it seems like it would be odd for someone to modify a value,
but I copy everything except byte slice references to be one the safe
side.

Copying could also be utilized when inlining fragments, if we wanted.
Right now there's a big warning on ReplaceFragmentSpread about field
aliasing (and it isn't clear to be that field de-duplication fully fixes
the aliasing).

### DIFF
--- a/pkg/ast/ast.go
+++ b/pkg/ast/ast.go
@@ -161,6 +161,10 @@ func (d *Document) NextRefIndex() int {
 	return d.RefIndex
 }
 
+func (d *Document) NewEmptyRefs() []int {
+	return d.Refs[d.NextRefIndex()][:0]
+}
+
 func (d *Document) AddRootNode(node Node) {
 	d.RootNodes = append(d.RootNodes, node)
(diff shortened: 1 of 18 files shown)


jensneuse:
I'd suggest that you make a copy of the ByteSlice and not just copy the "pointer". With your implementation, if I copy a node with a ByteSliceReference, then rename the original, that would also rename the copy. I don't think this would be the expected behaviour from a copy. What you can do is read the Bytes from the document.Input, append them so that you get a new ByteSliceReference, and then use that new ref. You could also add a helper function to document.Input to "copy" a ByteSliceReference. I think this is a safer way.

PR #338, opened 2022-02-05
Don't include all external fields in a type's child fields.

csilvers:
The local type field extractor is responsible for saying what fields
of what types can be resolved by a given datasource (upstream server).
This includes all the fields owned by the datasource, but it also
includes `@key` fields for federated types, since if you are extending
a type you know the value of its key field by definition, and
`@provides` fields.

The logic in local-type-field extractor was using `@external` as a
synonym for "is in a @key or @provides field".  Unfortunately, there's
a third reason a field might be declared `@external`: it's `@require`d
somewhere.  And such fields should definitely *not* be considered to
be owned by datasource.  (In fact, here "external" is saying exactly
the opposite of that!)

This PR extends the logic of the local-type-field extractor to ignore
external fields that are also marked `@required` in the same
type-extension.  This doubles down on the "negative" approach
(assuming everything external is actually provided except for explicit
exceptions) rather than the "positive" approach of actually
enumerating the instances where the schema reports a field is provided
(namely `@provides` and `@key` directives).  This is due to the
difficulty of parsing non-trivial key/provides fields like `@key:
"child { address { city } }"`.  It should work well unless there's a
fourth reason to mark something `@external` that we're missing!

(Of course, someone could add an `@external` field just for the heck
of it and things might break then.  So a "positive" approach would
still be better one day.)

One could imagine legitimate, but baroque, corner cases where this
fails, for instance something like:
```
extend type User {
    name: String @external
    fullName: String @requires("name")
}

extend type Review {
    reviewer: User @provides("name")
}
```
where `User.name` is locally providable in some contexts but not
others.  In that case we'll basically ignore the `@provides` here,
which is sub-optimal but not, technically, incorrect.

### DIFF
--- a/pkg/engine/plan/local_type_field_extractor.go
+++ b/pkg/engine/plan/local_type_field_extractor.go
@@ -52,6 +52,7 @@ type nodeInformation struct {
 	concreteTypeNames []string
 	localFieldRefs    []int
 	externalFieldRefs []int
+	requiredFields    map[string]struct{}
 }
 
 type rootNodeNamesMap struct {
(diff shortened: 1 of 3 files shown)


csilvers:
I just merged in the latest upstream master, so this should be mergeable now.

PR #339, opened 2022-02-10
WIP: Fix OnTypeName resolve panic

dnerdy:
There's a panic that occurs when resolving an object field and 1) the
object has an associated fetch, and 2) the field being resolved is
skipped because OnTypeName is set. In this case the context
`responseElements` are reset, but they aren't restored when skipping the
field. This results in a panic later when attemping to remove the
existing path elements from the context.

This PR is a work-in-progress because I still need to construct a test
case to demonstrate the issue.

### DIFF
--- a/pkg/engine/resolve/resolve.go
+++ b/pkg/engine/resolve/resolve.go
@@ -81,8 +81,10 @@ type Node interface {
 	NodeKind() NodeKind
 }
 
-type NodeKind int
-type FetchKind int
+type (
+	NodeKind  int
+	FetchKind int
+)
 
 const (
 	NodeKindObject NodeKind = iota + 1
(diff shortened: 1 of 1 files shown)


pkg/engine/resolve/resolve.go
```diff
@@ -1075,6 +1071,8 @@ func (r *Resolver) resolveObject(ctx *Context, object *Object, data []byte, obje
 			typeName, _, _, _ := jsonparser.Get(fieldData, "__typename")
 			if !bytes.Equal(typeName, object.Fields[i].OnTypeName) {
 				typeNameSkip = true
+				// Restore the response elements that may have been reset above.
+				ctx.responseElements = responseElements
```
csilvers, line 1078: I don't understand this code at all, but do you need to restore ctx.lastFetchID too?

dnerdy: Indeed, it looks like that's the case.

Issue #290, opened 2021-10-11
Pass WS initial payload from gateway to federated services


chedom:
Hi, for my use case it's required to pass user related information (token/id/roles) from gateway to federated services.
There are two possible options:

1. Use payload in `connection_init` message  (receiver must handle initPayload from `connection_init` message) (I prefer this option)
2. Use `Header` field from `FetchConfiguration`  (receiver must handle http header (before upgrading))

Both options aren't ready to use, they require some contribution.

@jensneuse which option would you prefer ?

chedom:
@jensneuse I checked all my cases, and it turned out that in all cases `ws initial payload` is almost exactly the same as `Header` for the same upstream. 
In my fork, I just added sending `Header` as` ws initial Payload`. If necessary, I can create a PR, otherwise theissue can be closed.

jensneuse:
I'm not sure I understand 100%. How would Header as ws initial payload look like?

spetrunin:
closed with https://github.com/wundergraph/graphql-go-tools/pull/425

PR #275, opened 2021-08-20
Feature/ws init func

chedom:
Add possibility to check initial payload to see whether to accept the websocket connection.
Example: authenticate user from websocket connection and save session of user in context.

PR #343, opened 2022-02-16
Propagate headers to WS upstream using connection_init

v0id3r:
Solves #290 

We already have `http.Header` and include it when hashing.
However, we do not propagate the actual value during the init+ack handshake.

With this change, it will send headers as payload.
Works perfectly together with client-facing WS init hook from #275 and makes it possible to propagate init payload from user WS to federation WS. Entities requests also follow and headers are attached to outgoing requests to data sources.

`playground` sends headers as payload instead of regular headers for subscriptions.
`gqlgen` has a special hook on WebSocket for init.

Going to add specific tests, but it is at least backward-compatible with existing suite.


v0id3r:
@chedom Hi, you mentioned that you've also implemented headers propagation at your fork. May I ask you to provide some feedback here, if any?

chedom:
Hi @v0id3r , my [solution](https://github.com/chedom/graphql-go-tools/commit/afb1f7a98bd206f9292a73daf8b476fbff99e4ec#diff-264442409bb80cf3c5d65dfc33eb05352fa6e349e56d7ca111a9e8dc476f4736) is very close to your implementation.

bailycase:
Is there any update on this being merged?

spetrunin:
closed in favor of https://github.com/wundergraph/graphql-go-tools/pull/425

PR #347, opened 2022-03-03
treat a union intersecting an interface symmetrically from interface | union

pvormste:
```
When expanding a named fragment, one of the cases to consider is that
the named fragment has union type, but the enclosing type is an
interface type.  (e.g. `user { ...data } fragment data on SomeUnion {...}`,
where `user` has an interface type.  fragment_spread_inlining.go already
treats this case correctly.

However, there's a symmetrical case where the enclosing type is a
union, and the naamed fragment has an interface type.  According to
the spec, this should be treated identically, but
fragment_spread_inlining.go was not considering this case.  With this
PR it is.

Fixing this turned up another problem in ast_node.go:
NodeFragmentIsAllowedOnInterfaceTypeDefinition properly allowed the
fragment on union types when appropriate, but in the symmetrical case,
NodeFragmentIsAllowedOnUnionTypeDefinition _never_ allowed the
fragment on interface types.  Now it uses the same condition as the
other case, allowing it when there's a non-zero intersection of the
two types.
```
from @csilvers

### DIFF
--- a/pkg/ast/ast_node.go
+++ b/pkg/ast/ast_node.go
@@ -499,7 +499,7 @@ func (d *Document) NodeFragmentIsAllowedOnUnionTypeDefinition(fragmentNode, unio
 	case NodeKindObjectTypeDefinition:
 		return d.NodeIsUnionMember(fragmentNode, unionTypeNode)
 	case NodeKindInterfaceTypeDefinition:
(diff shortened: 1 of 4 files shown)


pvormste:
wrong repository

PR #362, opened 2022-04-15
Decoupled GraphQL DataSource Planner from HTTP

clayne11:
It now uses a GraphqlDatasource interface to configure the
fetch and subscription. This will enable us to call into a local
GraphQL executable schema as well as a remote HTTP GraphQL
server.

I haven't solved how to handle configuration yet. Since config is different for different types of GraphQL data sources I need to figure out how to support this.


jensneuse:
After our last discussion, we've concluded that currently there's no need for a non-http implementation.

PR #367, opened 2022-05-03
fix broken object variable rendering when handling strings

pvormste:
This PR solves a problem with rendering object variables when the content is a string.
A test has been added in the `graphql` package.

**Problem**
When using object variable rendering strings are always being rendered with quotes, e.g. `{{ .object.name }}` => `"lucky"`

This behavior results in a problem when being used in a URL:
`https://petlibrary.fake/{{ .object.name }}` => `https://petlibrary.fake/"lucky"`

### DIFF
--- a/pkg/engine/datasource/introspection_datasource/input.go
+++ b/pkg/engine/datasource/introspection_datasource/input.go
@@ -33,7 +33,7 @@ var (
 	rBrace                 = []byte("}")
 	comma                  = []byte(",")
 	requestTypeField       = []byte(`"request_type":`)
-	onTypeField            = []byte(`"on_type_name":{{ .object.name }}`)
+	onTypeField            = []byte(`"on_type_name":"{{ .object.name }}"`)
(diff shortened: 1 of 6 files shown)


pvormste:
@jensneuse As far as I remember it is already using a PlainVariableRenderer in this case - it just wasn't stripping the quotes again which leads to a lot of strange behavior - see the changed code for the introspection data source.

I think it's rather a bug in the PlainVariableRenderer than a missing feature like a missing URLEncodedRenderer (which doesn't means it won't make sense to add it)

jensneuse:
> This is exactly what this PR fixes: It always omits the Quotes for the PlainVariableRenderer. In the current state this example: `"{{.object.name}}"` is not possible, because it will result in `""lucky""` and this PR changes it to `"lucky"`.
> 
> Or maybe I'm not understanding the issue

Got it, makes sense. Thanks for clarification!

jensneuse:
merged with #369

pvormste:
Thanks! Sorry for the confusion! :-)

PR #380, opened 2022-06-28
Input field default values

kofoworola:
This fixes an issue in the astnormalization where input field default values are not used

Example

```
enum MyEnum {
VALUE_A
VALUE_B
}

input MyInput {
fieldA: MyEnum! = VALUE_A
fieldB: String
}

extend type Mutation {
myMutation(input: MyInput!): Boolean!
}
```


jensneuse:
great work @kofoworola!

Issue #277, opened 2021-08-23
refactor: ExternalError and RequestError should be merged and cleaned up


jensneuse:
We should consider to if the ast.Path wrapper is needed.

PR #388, opened 2022-07-15
subscription: add support for ka while waiting for ack (#277)

kofoworola:
## Issue
When running a GQL subscription (and thus using a WSS connection) and upstream does not return connection_ack as its first response, the subscription fails and "message": "expected connection_ack, got: ka” is shown. Also, logs are spammed with messages

## Solution
While waiting for a connection_ack message also listen and accept ka messages

[Spec here](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md)


spetrunin (approved):

spetrunin:
merged via https://github.com/wundergraph/graphql-go-tools/pull/390

PR #391, opened 2022-07-25
check errors before marking a query as parsed

pvormste:
Fix authored by: @buraksezer 

This PR fixes an issue where a query is being marked as parsed even if parsing fails.
This behavior results in a misbehaving subscription.


spetrunin:
merged via https://github.com/wundergraph/graphql-go-tools/pull/393

PR #392, opened 2022-07-25
remove circular dependencies

pvormste:
This PR removes the circular dependency between `examples/federation` and the library itself.
It will reduce the chance of breaking the go mod CLI caused by not being able to resolve the circular dependency properly.


spetrunin:
merged via https://github.com/wundergraph/graphql-go-tools/pull/394

PR #402, opened 2022-08-15
fix: return after parsing arg fields of type ast.ValueKindList

maronavenue:
### Summary 

Suppose we have this graphql query request with the following query variables:

#### Sample request
```graphql
query book_sn($id: String) {
  book_attic {
    oreilly_books(
      where: {
        AND: [ # Parsed ast ast.ValueKindList since it contains query variables, instead of the usual ast.ValueKindVariable
          { python: { tag: { eq: "beginner" } } }
          { golang: { serial_number: { eq: $id } } }
          {
            OR: [
              {
                AND: [
                  {
                    python: {
                      date_published: { eq: "2022-07-29" }
                    }
                  }
                  {
                    python: { date_purchased: { eq: "2022-07-29" } }
                  }
                ]
              }
              {
                AND: [
                  {
                    golang: {
                      date_published: { eq: "2022-08-01" }
                    }
                  }
                  {
                    golang: { date_purchased: { eq: "2022-08-01" } }
                  }
                ]
              }
            ]
          }
        
      }
    ) {
      jrr_tolkien_books
      george_rr_martin_books
    }
  }
}

```

```json
{"id": "1801071039"}
```

This will be stored into an `ast.Document` which then gets parsed during validation in https://github.com/wundergraph/graphql-go-tools/blob/v1.51.0/pkg/astvalidation/operation_validation.go#L68.

It then fails within an implementation of `EnterArgument()` specifically for [allVariablesUsedVisitor](https://github.com/wundergraph/graphql-go-tools/blob/master/pkg/astvalidation/operation_rule_all_variables_used.go#L76) because it attempts to continue processing a field of type `ast.ValueKindList` but we only resolve `ast.ValueKindVariable` fields to its actual `ast.VariableValue` mapping. Thus, the out of bound index error as we can see from the screenshot below:

graphql-go-tools-bug

### Changelog

- Break case condition after iterating all elements of an `ast.ValueKindList` within the recursive function, `veifyValue()`.
- Add corresponding test case for under `5.8.4 All Variables Used` grouping to capture this case:

**Before:**

```bash
--- FAIL: TestExecutionValidation (0.09s)
    --- FAIL: TestExecutionValidation/5.8_Variables (0.01s)
        --- FAIL: TestExecutionValidation/5.8_Variables/5.8.4_All_Variables_Used (0.00s)
            --- FAIL: TestExecutionValidation/5.8_Variables/5.8.4_All_Variables_Used/variables_in_nested_array_object (0.00s)
panic: runtime error: index out of range [2] with length 2 [recovered]
	panic: runtime error: index out of range [2] with length 2
``` 

**After:**

```bash
ok  	github.com/wundergraph/graphql-go-tools/pkg/astvalidation	0.940s

$ go test -count=1 ./pkg/astvalidation/...
ok      github.com/wundergraph/graphql-go-tools/pkg/astvalidation       0.959s
?       github.com/wundergraph/graphql-go-tools/pkg/astvalidation/reference     [no test files]
ok      github.com/wundergraph/graphql-go-tools/pkg/astvalidation/reference/testsgo     0.157s
```

### Tests

- [x] Manual testing
- [x] Functional testing

@jensneuse

### DIFF
--- a/pkg/astvalidation/operation_rule_all_variables_used.go
+++ b/pkg/astvalidation/operation_rule_all_variables_used.go
@@ -69,6 +69,7 @@ func (a *allVariablesUsedVisitor) verifyValue(value ast.Value) {
 		for _, i := range a.operation.ListValues[value.Ref].Refs {
 			a.verifyValue(a.operation.Values[i])
 		}
+		return
 	default:
 		return // skip all others
 	}
--- a/pkg/astvalidation/operation_validation_test.go
+++ b/pkg/astvalidation/operation_validation_test.go
@@ -3524,6 +3524,48 @@ func TestExecutionValidation(t *testing.T) {
 								}`,
 					AllVariablesUsed(), Valid)
 			})
+			t.Run("variables in nested array object", func(t *testing.T) {
+				run(`query variableUnused($name: String) {
+					dog(where: {
+						AND: [
+						  { nestedDog: { nickname: { eq: "Scooby Doo" } } }
+						  { nestedDog: { name: { eq: $name } } }
+						  {
+							OR: [
+							  {
+								AND: [
+								  {
+									nestedDog: {
+									  birthday: { eq: "2021-07-29" }
+									}
+								  }
+								  {
+									nestedDog: { barkVolume: { eq: 20 } }
+								  }
+								]
+							  }
+							  {
+								AND: [
+								  {
+									nestedDog: {
+										birthday: { eq: "2021-08-02" }
+									}
+								  }
+								  {
+									nestedDog: { barkVolume: { eq: 35 } }
+								  }
+								]
+							  }
+							]
+						  }
+						]
+					  }) {
+						id
+						name
+					}
+				}`,
+					AllVariablesUsed(), Valid)
+			})
 		})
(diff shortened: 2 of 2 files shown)


jensneuse:
Can you add a test?

maronavenue:
Please note that this does not normally occur and commonly caught because the package only instantiates arg fields as `ast.ValueKindList` or `ast.ValueKindObject` only if they contain query variables that need further resolution.

Arg fields that does not contain query variables are instantiated as `ast.VaueKindVariable` which is resolved to its `ast.VariableValue` mapping.

As such, this bug occurs specifically if the request ast document has both an arg field list and query variable within.

maronavenue:
@jensneuse - Added test to capture this. It an also be used for reproducer. Let me know if this looks good.

maronavenue:
Let me know if there's anything else needed here. Thanks @jensneuse. I'd be happy to merge after getting approvals and passing workflows.

maronavenue:
Any updates here? @jensneuse

jensneuse:
I'm on vacation this week. Will probably review and merge next week.

maronavenue:
Sounds good, @jensneuse. Have a great trip!

maronavenue:
Can you give us a rough ETA when you could merge this in? @jensneuse

PR #406, opened 2022-09-04
Feature/ws init func (rebased on master)

argoyle:
This is a rebased version of #275

### DIFF
--- a/pkg/subscription/init.go
+++ b/pkg/subscription/init.go
@@ -1,7 +1,5 @@
 package subscription
 
-type key string
-
 // InitPayload is a structure that is parsed from the websocket init message payload.
 type InitPayload map[string]interface{}
 
--- a/pkg/subscription/init.go
+++ b/pkg/subscription/init.go
@@ -1,0 +1,35 @@
+package subscription
+
+type key string
+
+// InitPayload is a structure that is parsed from the websocket init message payload.
+type InitPayload map[string]interface{}
+
+// GetString safely gets a string value from the payload. It returns an empty string if the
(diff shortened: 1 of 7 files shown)


pkg/subscription/init.go
```diff
@@ -0,0 +1,33 @@
+package subscription
+
+// InitPayload is a structure that is parsed from the websocket init message payload.
+type InitPayload map[string]interface{}
```
jensneuse: This assumes that InitPayload is a map. What if it's any other valid JSON? Could it be an Array or a string, bool or number?

I see no other issues with this PR, but would suggest that you rewrite this to accept any kind of JSON, or maybe even a binary payload. What are your thoughts on this? (just making it a bit more generic) Nothing against the Authorization helper method. It can live in some way, but the init handler should be unopinionated.

jensneuse (approved):

pkg/subscription/init.go
argoyle: Looking at the type of connectionParams in `subscription-transport-ws` it seems to always be a map:
```
export declare type ConnectionParams = {
    [paramName: string]: any;
};
```

argoyle: Ok, coming back to this one finally. I changed InitPayload to be a json.RawMessage and then trying to unmarshal it when getting a string value. That should make it handle any kind of value, right?

jensneuse:
@spetrunin can you give this PR a review? I think it looks fine.

argoyle:
🎉 Thanks!

PR #410, opened 2022-09-19
Yuri/eng 356 improve subscription implementation

YuriBuerov:


pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -215,236 +144,89 @@ func (c *WebSocketGraphQLSubscriptionClient) generateHandlerIDHash(options Graph
...
-		return
-	}
-	sub, ok := h.subscriptions[id]
-	if !ok {
-		return
-	}
-	payload, _, _, err := jsonparser.Get(data, "payload")
-	if err != nil {
-		return
+	if err := waitForAck(reqCtx, conn); err != nil {
```
jensneuse, line 202: what if there's no ack?

jensneuse (commented):
seems ok so far

pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
YuriBuerov: we have ackWaitTimeout

jensneuse (approved):

PR #432, opened 2022-10-28
feat: make error message for missing fields more detailed

fiam:
Include the available fields in the type


spetrunin:
solution is too generic and lists all fields from a type

won't fix for now

PR #433, opened 2022-10-28
fix: don't swallow error from jsonparser.ArrayEach

fiam:
Fixes https://github.com/wundergraph/wundergraph/issues/278


spetrunin:
discarded

Issue #431, opened 2022-10-23
[Bug] Queries containing Boolean fail validation


alasdairtran:
Consider the following test cases, which I added in my fork: https://github.com/mochilabs/graphql-go-tools/commit/25ef05295385683ba7df24c7f9bea638c860c84f

I think both of them are valid queries, but they both fail the validation test, giving us fairly opaque error such as:

```
--- FAIL: TestExecutionValidation (0.16s)
    --- FAIL: TestExecutionValidation/5.8_Variables (0.07s)
        --- FAIL: TestExecutionValidation/5.8_Variables/5.8.5_All_VariableValue_Usages_are_Allowed (0.06s)
            --- FAIL: TestExecutionValidation/5.8_Variables/5.8.5_All_VariableValue_Usages_are_Allowed/with_boolean_input (0.00s)
panic: want expectation: Valid, got: Invalid
reason: external: value: {shared: {equals: $a}} doesn't satisfy inputType: nodepoolWhereInput, locations: [], path: [query,findFirstnodepool]
operation:
query QueryWithBooleanInput($a: Boolean){findFirstnodepool(where: {shared: {equals: $a}}){id}}
 [recovered]
        panic: want expectation: Valid, got: Invalid
reason: external: value: {shared: {equals: $a}} doesn't satisfy inputType: nodepoolWhereInput, locations: [], path: [query,findFirstnodepool]
operation:
query QueryWithBooleanInput($a: Boolean){findFirstnodepool(where: {shared: {equals: $a}}){id}}

```

I came across this as I was integrating Hasura as a backend for WunderGraph. Queries involving boolean inputs/arguments are common in the Hasura API.

I don't know Go or this codebase well enough to dig any deeper. Hope those test cases will allow you guys to reproduce the error.

PR #437, opened 2022-11-04
Add failed test cases with boolean args

alasdairtran:
Hi @jensneuse 

Just following up on #431, here's a PR with two failed test cases containing queries with boolean args. I don't know Go that well, so I might not be able to provide a fix, but hopefully this will point you guys in the right direction.


jensneuse:
@alasdairtran thank you! I'll add it to my backlog and will fix it.

Issue #310, opened 2021-11-10
Oct 2021 Spec


mentat:
Is there plans to support the Oct. 2021 specification? https://spec.graphql.org/October2021/

Specifically repeatable directives are a key feature.

jensneuse:
Yeah, absolutely! Do you want to step ahead?

mentat:
I'm actually not even sure how this library can support Apollo Federation as Fed 1 includes this directive on the supergraph:

```
directive @join__type(graph: join__Graph!, key: join__FieldSet) repeatable on OBJECT | INTERFACE
```
I'd be happy to test out Fed1 pre-composed supergraphs with this library when that is supported.

PR #439, opened 2022-11-04
starting a subscription without attached data source crashes (#310)

pvormste:
authored by @buraksezer:

Now the library returns the following error if the data source has not been defined:

```json
{"errors":[{"message":"no data source found"}]}
```

I investigated the possibility of controlling this before calling `ResolveGraphQLSubscription` but it seems that there is no way to do it. See the following:

```go
func (v *Visitor) LeaveDocument(operation, definition *ast.Document) {
	for _, config := range v.fetchConfigurations {
		if config.isSubscription {
			v.configureSubscription(config)
		} else {
			v.configureObjectFetch(config)
		}
	}
}
```

`v.fetchConfigurations` was empty our case. Possibly, we need to spend more time implementing a more general solution. The solution proposed here just works.

### DIFF
--- a/pkg/engine/resolve/resolve.go
+++ b/pkg/engine/resolve/resolve.go
@@ -508,6 +508,15 @@ func (r *Resolver) ResolveGraphQLResponse(ctx *Context, response *GraphQLRespons
 	return writeGraphqlResponse(buf, writer, ignoreData)
 }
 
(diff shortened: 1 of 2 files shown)


spetrunin:
Superseded by https://github.com/wundergraph/graphql-go-tools/pull/442

Issue #429, opened 2022-10-17
[Feature Request] Support POST method in SSE requests


qiangfoo:
The SSE connection is hardcoded to use `GET` method (https://github.com/wundergraph/graphql-go-tools/blob/master/pkg/engine/datasource/graphql_datasource/graphql_sse_handler.go#L188). 

But for my use case, the graphql server only takes `POST` requests. I believe this is a very common practice in graphql servers. Can we make the http method configurable?

jensneuse:
We don't have a use case for this, but you can open a PR if you're interested.

spetrunin:
fixed by https://github.com/wundergraph/graphql-go-tools/pull/449

PR #449, opened 2022-11-16
Add Support for Post Requests in SSE Subscriptions

jplee0324:
Hi @jensneuse!

This PR adds support for sending POST requests in subscriptions using SSE
Made everything default to sending a GET request unless POST is explicitly defined in the field

Addresses Issue [#429](https://github.com/wundergraph/graphql-go-tools/issues/429)

### DIFF
--- a/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
+++ b/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
@@ -51,6 +51,9 @@ type Planner struct {
 	upstreamDefinition                 *ast.Document
 	currentVariableDefinition          int
 	addDirectivesToVariableDefinitions map[int][]int
+
+	insideCustomScalarField bool
+	customScalarFieldRef    int
 }
 
 func (p *Planner) EnterVariableDefinition(ref int) {
@@ -198,10 +201,16 @@ func (p *Planner) DataSourcePlanningBehavior() plan.DataSourcePlanningBehavior {
 }
 
 type Configuration struct {
-	Fetch          FetchConfiguration
-	Subscription   SubscriptionConfiguration
-	Federation     FederationConfiguration
-	UpstreamSchema string
+	Fetch                  FetchConfiguration
+	Subscription           SubscriptionConfiguration
+	Federation             FederationConfiguration
+	UpstreamSchema         string
+	CustomScalarTypeFields []SingleTypeField
+}
+
+type SingleTypeField struct {
+	TypeName  string
+	FieldName string
 }
 
 func ConfigJson(config Configuration) json.RawMessage {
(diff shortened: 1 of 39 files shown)


pkg/engine/datasource/graphql_datasource/graphql_sse_handler.go
```diff
@@ -185,6 +186,35 @@ func trim(data []byte) []byte {
 }
 
 func (h *gqlSSEConnectionHandler) performSubscriptionRequest(ctx context.Context) (*http.Response, error) {
+
+	var req *http.Request
+	var err error
+
+	// default to GET requests when POST is not specified in the SubscriptionConfiguration
+	if h.options.SSEMethod == http.MethodPost {
```
jensneuse: Overall, the PR looks good to me.
But I think, as we're only supporting GET & POST, the field SSEMethod should just be a boolean that defaults to false (GET) or can be set to true (POST).
E.g. `SSEMethodPOST`.
Otherwise, you could set SSEMethod to `PUT` but it would still use `GET`, a bit confusing.

pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -1262,10 +1264,11 @@ type GraphQLSubscriptionClient interface {
 }
 
 type GraphQLSubscriptionOptions struct {
-	URL    string      `json:"url"`
-	Body   GraphQLBody `json:"body"`
-	Header http.Header `json:"header"`
-	UseSSE bool        `json:"use_sse"`
+	URL       string      `json:"url"`
+	Body      GraphQLBody `json:"body"`
```
jensneuse: In conjunction with renaming `SSEMethod` to `SSEMethodPost bool`, I'd change `Body` to `PostBody`.

jensneuse (changes requested):

jensneuse:
Hey, @jplee0324! Thank you for this PR. I've requested 2 minor changes. Please let me know if you're ok, and want to make the changes. Happy to merge then.

pkg/engine/datasource/graphql_datasource/graphql_datasource.go
devsergiy: I believe it is not required, cause this body is a graphql payload with query, variables etc.

jplee0324:
> Hey, @jplee0324! Thank you for this PR. I've requested 2 minor changes. Please let me know if you're ok, and want to make the changes. Happy to merge then.

Awesome! Thanks @jensneuse! Just made the change to change the field to a `bool` since I agree that makes more sense :D For the `Body` -> `PostBody` comment, I'm leaning towards what @spectrunin said and think it'd be nice to have to name a bit more general since this `Body` object is used in other places than the method for `POST` requests. Either way I don't have any strong opinions so I can still change it if you'd like!

jensneuse (approved):
@spetrunin please have another look and merge if you're happy. I think it's fine this way.

spetrunin (approved):

Issue #459, opened 2022-11-30
graphql.Request.SetHeader doesn't work


daniil-kzn:
I set a new header to the graphql.Request but there is no the header in a subgraph service.
```
var gqlRequest graphql.Request
gqlRequest.SetHeader(http.Header{"Some-Header": []string{"some value"}})
```
[https://github.com/wundergraph/graphql-go-tools/blob/master/examples/federation/gateway/http/http.go#L21](https://github.com/wundergraph/graphql-go-tools/blob/master/examples/federation/gateway/http/http.go#L21)

Maybe there is an another way to sent a custom header to subgraph service.

roderm:
You could add a `http.RoundTripper` to the [httpClient passed into DataSourcePoller](https://github.com/wundergraph/graphql-go-tools/blob/master/examples/federation/gateway/main.go#L50)
```go
type AddHeaderTransport struct {
	T      http.RoundTripper
}

func (t *AddHeaderTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	req.Header.Add("Some-Header", "some value")
	return t.T.RoundTrip(req)
}
```
Creating the client:
```go
httpClient := http.DefaultClient
httpClient.Transport = &AddHeaderTransport{T: http.DefaultTransport}

datasourceWatcher := NewDatasourcePoller(httpClient, DatasourcePollerConfig{
	Services: []ServiceConfig{
		{Name: "accounts", URL: "http://localhost:4001/query"},
		{Name: "products", URL: "http://localhost:4002/query", WS: "ws://localhost:4002/query"},
		{Name: "reviews", URL: "http://localhost:4003/query"},
	},
	PollingInterval: 30 * time.Second,
})
```

daniil-kzn:
It works but I explained my case not so clear.
The gateway receives some header from a frontend.
How can I pass the header to a subgraph?

jensneuse:
You can pass client request headers like this to a subgraph:
https://github.com/wundergraph/graphql-go-tools/blob/607a7ec39dbf195cfe1ccf8b8c1f837f4670bbb0/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go#L180

daniil-kzn:
c
https://github.com/wundergraph/graphql-go-tools/blob/607a7ec39dbf195cfe1ccf8b8c1f837f4670bbb0/examples/federation/gateway/http/http.go#L18

mcondie:
[...]

PR #460, opened 2022-12-03
Ability to send custom headers to subservices

daniil-kzn:


jensneuse:
Can you add any sort of test to make it clear what this PR changes/fixes?

daniil-kzn:
Added simple test to check that request from context will pass to the final request to subservice.

jensneuse:
I see what you're trying to do.
Do you think the following existing solution already solves your problem?
https://github.com/wundergraph/graphql-go-tools/blob/607a7ec39dbf195cfe1ccf8b8c1f837f4670bbb0/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go#L180

Or is an additional solution required?

daniil-kzn:
@jensneuse I am not sure the solution helps me because I have no idea how to inject plan.Configuration to g.engine in here
[plan.Configuration](https://github.com/wundergraph/graphql-go-tools/blob/master/examples/federation/gateway/http/http.go#L30)

daniil-kzn:
@jensneuse Any hints how can I configurate Headers from the method ?
https://github.com/wundergraph/graphql-go-tools/blob/607a7ec39dbf195cfe1ccf8b8c1f837f4670bbb0/examples/federation/gateway/http/http.go#L18

spetrunin:
@daniil-kzn you don't have to inject plan configuration on the execution stage

You have a place where you defining your datasource configuration and passing it to execution engine constructor
So you could add similar configuration as in example propose by @jensneuse 

After that you could send you headers along with request and they will be forwarded to upstream

daniil-kzn:
@spetrunin I have the same code as in the gateway example https://github.com/wundergraph/graphql-go-tools/tree/master/examples/federation/gateway and there is the datasource configuration only in the datasouirce_poller
I tried to add Authorization header here: 
https://github.com/wundergraph/graphql-go-tools/blob/d2fb53dada07e7defacb7b36b7c223d6b3750f2d/examples/federation/gateway/datasource_poller.go#L156

but it didn't help to send the header within an regulal request to upstream.

daniil-kzn:
@jensneuse @spetrunin So, guys, let's merge it or fix the issue https://github.com/wundergraph/graphql-go-tools/issues/459

PR #463, opened 2022-12-07
Fix normalization when variables extraction is disabled

spetrunin:
astnormalization: skip applying variables normalization when variables extraction is disabled
default values extraction normalizer: make an argument required when it is optional and has default value

fixes ENG-737

### DIFF
--- a/pkg/astnormalization/variables_default_value_extraction_test.go
+++ b/pkg/astnormalization/variables_default_value_extraction_test.go
@@ -54,7 +54,7 @@ func TestVariablesDefaultValueExtraction(t *testing.T) {
 			mutation simple($in: String = "bar" ) {
 			  simple(input: $in)
 			}`, "", `
-			mutation simple($in: String) {
+			mutation simple($in: String!) {
 			  simple(input: $in)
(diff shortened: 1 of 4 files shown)


pkg/astnormalization/variables_default_value_extraction_test.go
```diff
@@ -7,75 +7,263 @@ import (
...
 		t.Run("value provided", func(t *testing.T) {
 			runWithVariablesDefaultValues(t, extractVariablesDefaultValue, variablesDefaultValueExtractionDefinition, `
-			mutation simple {
-			  simple(input: "bazz")
-			}`, "", `
-			mutation simple {
-			  simple(input: "bazz")
-			}`, ``, ``)
+				mutation simple {
+					simple(input: "bazz")
```
jensneuse, line 61: Why do we not extract inline values? Is this another handler?

devsergiy: yeah, it is handled by VariablesExtraction normalizer

this one handles only default values

jensneuse: ok

PR #464, opened 2022-12-08
fix planning variable path

spetrunin:
regression from https://github.com/wundergraph/graphql-go-tools/pull/373

fixes ENG-907


spetrunin:
@pvormste FYI

PR #469, opened 2022-12-12
Fix: Cannot specify extend after blockstring

### LINKED ISSUE #468: Cannot specify extend after blockstring
Cannot specify extend after blockstring

Example
```
""" comment ""
extend type Foo {
  var: Int
}
```

Error
```
parse graphql document string: external: unexpected literal - got: EXTEND want one of: [TYPE INPUT SCALAR INTERFACE UNION ENUM DIRECTIVE], locations: [{Line:3 Column:1}], path: []
```

I tried to fix it.
https://github.com/wundergraph/graphql-go-tools/commit/391b818eca3da057c736b432a114ed8153546408

Mistat:
fixes #468

### DIFF
--- a/pkg/astparser/parser.go
+++ b/pkg/astparser/parser.go
@@ -715,6 +715,8 @@ func (p *Parser) parseRootDescription() {
 		p.parseEnumTypeDefinition(&description)
 	case identkeyword.DIRECTIVE:
 		p.parseDirectiveDefinition(&description)
+	case identkeyword.EXTEND:
+		p.parseExtension()
 	default:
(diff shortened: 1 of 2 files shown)


jensneuse:
@Mistat can you add a test to cover this new code? the solution seems to be fine

Mistat:
I wrote a test code and confirmed that the "Block string" before "extend" is ignored.

jensneuse:
> I wrote a test code and confirmed that the "Block string" before "extend" is ignored.

thank you

spetrunin (approved):

PR #472, opened 2022-12-21
feat: add removeNullVariables directive

spetrunin:
- feat: add removeNullVariables directive
- fix: ignore null in variable names when we unnull variables


pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -1307,7 +1307,7 @@ func (s *Source) unNullVariables(input []byte) ([]byte, bool) {
 		}
 		return append(input[:startQuote], input[end:]...), true
 	}
-	if i := bytes.Index(input, []byte("null")); i != -1 {
+	if i := bytes.Index(input, []byte("null")); i != -1 && input[i-1] == ':' {
```
jensneuse: can you make this check more explicit? I feel like this is still error-prone.

jensneuse (changes requested):

pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -67,9 +70,15 @@ func (p *Planner) LeaveVariableDefinition(_ int) {
 
 func (p *Planner) EnterDirective(ref int) {
 	parent := p.nodes[len(p.nodes)-1]
-	if parent.Kind == ast.NodeKindOperationDefinition && p.currentVariableDefinition != -1 {
-		p.addDirectivesToVariableDefinitions[p.currentVariableDefinition] = append(p.addDirectivesToVariableDefinitions[p.currentVariableDefinition], ref)
-		return
+	if parent.Kind == ast.NodeKindOperationDefinition {
+		if p.visitor.Operation.DirectiveNameString(ref) == "removeNullVariables" {
```
devsergiy: remove?

pkg/asttransform/baseschema.go
```diff
@@ -166,6 +166,9 @@ directive @deprecated(
     reason: String = "No longer supported"
 ) on FIELD_DEFINITION | ENUM_VALUE
 
+"Enables removing variables with null values from the operation."
```
jensneuse: Can you make this a multi line comment explaining with a small example how it will affect the request?

pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -1279,48 +1292,60 @@ func (s *Source) compactAndUnNullVariables(input []byte) []byte {
...
+
+	removeNullVariables := httpclient.IsInputFlagSet(input, httpclient.UNNULLVARIABLES)
+	variables = s.cleanupVariables(variables, removeNullVariables)
+
+	input, _ = jsonparser.Set(input, variables, "body", "variables")
+	return input
+}
+
+func (s *Source) cleanupVariables(variables []byte, removeNullVariables bool) []byte {
 	cp := make([]byte, len(variables))
```
jensneuse, line 1304: if second arg is false, we can skip this completely, no?

jensneuse (changes requested):

pkg/asttransform/baseschema.go
devsergiy: yep

pkg/engine/datasource/graphql_datasource/graphql_datasource.go
devsergiy: This method is also cleaning up empty objects `{}`

so without directive we skip removing only top level variables with `null` value, but still removing `{}`

pkg/asttransform/baseschema.go
devsergiy: updated

jensneuse (approved):

Issue #473, opened 2022-12-23
Field transformations


bakjos:
First at all, thank you so much for this wonderful library. I was wondering if there’s a way to apply field transformations when using the execution engine v2, by looking at the code seems like it’s only supported in v1. What I like to achieve is to expose a field by composing to other fields `name = firstName + lastName` and I’d like to contribute this feature, but not sure if it makes sense and if the right approach would be to add a new option to the `FieldConfiguration` for the transformation using https://github.com/jensneuse/pipeline/pkg/pipe

PR #474, opened 2022-12-23
feat: Add transformation option to the field configuration

bakjos:
This PR adds the option for applying a transformation using the [pipeline library](https://github.com/jensneuse/pipeline) to fix https://github.com/wundergraph/graphql-go-tools/issues/473

### DIFF
--- a/pkg/engine/plan/plan.go
+++ b/pkg/engine/plan/plan.go
@@ -9,6 +9,8 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/jensneuse/pipeline/pkg/pipe"
+
 	"github.com/wundergraph/graphql-go-tools/pkg/ast"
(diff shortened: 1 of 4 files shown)


pkg/engine/plan/plan.go
```diff
@@ -933,6 +1008,25 @@ func (v *Visitor) resolveFieldPath(ref int) []string {
 	return []string{fieldName}
 }
 
+func (v *Visitor) resolveTransformation(ref int) *pipe.Pipeline {
```
bakjos: I put it as an independent function, but for performance could be done as part of the `resolveFieldPath` function

devsergiy:
Hi @bakjos 

v1 is no longer under active development
it will receive only critical bug fixes

what should we do with this PR?

PR #476, opened 2022-12-27
fix: export directive for id scalar

spetrunin:


jensneuse (changes requested):
How can we make this work for any custom scalar?

jensneuse (approved):

PR #492, opened 2023-01-26
Yuri/eng 1110 resolving json scalars handle plain types

YuriBuerov:
Addressed the following issues:
- https://github.com/wundergraph/wundergraph/issues/448
-  https://github.com/wundergraph/wundergraph/issues/562

### DIFF
--- a/pkg/engine/resolve/resolve.go
+++ b/pkg/engine/resolve/resolve.go
@@ -954,7 +954,9 @@ func (r *Resolver) resolveString(ctx *Context, str *String, data []byte, stringB
 	if str.UnescapeResponseJson {
 		value = bytes.ReplaceAll(value, []byte(`\"`), []byte(`"`))
 
-		// if value is a string
+		// When the original value from upstream response was a string `"hello"`
+		// after getting it via jsonparser.Get we will return unquoted value `hello`,
+		// which is invalid json, so we need to quote it again
 		if !gjson.ValidBytes(value) {
 			// wrap value in quotes to make it valid json
 			value = append(literal.QUOTE, append(value, literal.QUOTE...)...)
--- a/pkg/engine/resolve/resolve.go
+++ b/pkg/engine/resolve/resolve.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/buger/jsonparser"
 	"github.com/cespare/xxhash/v2"
+	"github.com/tidwall/gjson"
 	errors "golang.org/x/xerrors"
 
 	"github.com/wundergraph/graphql-go-tools/internal/pkg/unsafebytes"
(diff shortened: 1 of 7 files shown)


jensneuse:
@YuriBuerov sorry but I feel a bit lost. Can you explain in simple terms how this solves the 2 linked issues?
I think we've come to a place where the functions used in `Load()` might need some comments and descriptions to make sense to an outsider. Why do we use which function and in which case? Would be nice if we could make this a bit more clear.

pkg/engine/resolve/resolve.go
```diff
@@ -952,6 +953,13 @@ func (r *Resolver) resolveString(ctx *Context, str *String, data []byte, stringB
 
 	if str.UnescapeResponseJson {
 		value = bytes.ReplaceAll(value, []byte(`\"`), []byte(`"`))
+
+		// if value is a string
```
devsergiy: could we extend comment to smth like that?

```go
// When the original value from usptream response was a string `"hello"`
// after getting it via jsonparser.Get we will get unquoted value `hello`,
// which is invalid json and we need to quote it again
```

pkg/engine/resolve/resolve_test.go
```diff
@@ -1514,6 +1514,106 @@ func TestResolver_ResolveNode(t *testing.T) {
 			},
 		}, Context{Context: context.Background()}, `{"id":1,"name":"Jens","pet":{"name":"Woofie"}}`
 	}))
+	t.Run("json object within a string", testFn(false, false, func(t *testing.T, ctrl *gomock.Controller) (node Node, ctx Context, expectedOutput string) {
```
devsergiy: we could add an additional wrapping test group "with unescape json enabled"

pkg/engine/resolve/resolve_test.go
```diff
@@ -1514,6 +1514,106 @@ func TestResolver_ResolveNode(t *testing.T) {
...
+					HasBuffer: true,
+				},
+			},
+		}, Context{Context: context.Background()}, `{"data":[1, 2, 3]}`
+	}))
+	t.Run("json boolean within a string", testFn(false, false, func(t *testing.T, ctrl *gomock.Controller) (node Node, ctx Context, expectedOutput string) {
+		return &Object{
+			Fetch: &SingleFetch{
+				BufferId:   0,
+				DataSource: FakeDataSource(`{"data": "true"}`),
```
devsergiy: hmm, looks like an edge case

upstream is returning a string `"true"` not a boolean `true`
wdt should we treat it as a string?

devsergiy: disregard, it is desired with unescapeJson

YuriBuerov: I think we need to be consistent. If we treat an object encoded in a string as a JSON object, I would expect similar behavior for boolean/array etc. What do you think?

pkg/engine/resolve/resolve_test.go
YuriBuerov: it's enabled, check line 1530, is that what you mean?

spetrunin (approved):

greemo:
Hi guys, thanks for the effort on this. Is there a way to disable the unescapeResponseJson functionality?

We're using Wundergraph in a very generic platform, and unescaping a JSON response will not always be appropriate.

For me, it would work best as part of the data source configuration. I've looked through the code and could not find any config to enable/disable it. Please excuse me if it's there, and point me in the right direction.

Cheers!

spetrunin:
@greemo Could I ask you to open a discussion/issue in the Wundergraph repo and describe your use case?

PR #494, opened 2023-02-06
feat: engine, distinguish between null and omitted input variables

YuriBuerov:


pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
```diff
@@ -8146,6 +8146,26 @@ func TestSource_Load(t *testing.T) {
...
+		var (
+			src       = &Source{httpClient: &http.Client{}}
+			serverUrl = ts.URL
+			variables = []byte(`{"a":null,"b":"null"}`)
+		)
+		t.Run("should remove undefined variables and leave null variables", func(t *testing.T) {
+			var input []byte
+			input = httpclient.SetInputBodyWithPath(input, variables, "variables")
+			input = httpclient.SetInputURL(input, []byte(serverUrl))
+			input = httpclient.SetInputFlag(input, httpclient.UNNULLVARIABLES)
```
devsergiy: I believe `UNNULLVARIABLES` flag should not be set for this test

pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
```diff
@@ -8146,6 +8146,26 @@ func TestSource_Load(t *testing.T) {
 			assert.Equal(t, `{"variables":{"a":null,"b":"b","c":{}}}`, buf.String())
 		})
 	})
+	t.Run("remove undefined variables", func(t *testing.T) {
+		var (
+			src       = &Source{httpClient: &http.Client{}}
+			serverUrl = ts.URL
+			variables = []byte(`{"a":null,"b":"null"}`)
```
devsergiy: `"b":"null"` is a string `"null"` not the null value
I believe it should be `"b": null`

In case you will set it like this and set a flag `UNNULLVARIABLES` both `a` and `b` will be removed

pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -1342,31 +1343,33 @@ func (s *Source) compactAndUnNullVariables(input []byte) []byte {
...
 	copy(cp, variables)
 
-	if removeNullVariables {
-		// remove null variables from JSON: {"a":null,"b":1} -> {"b":1}
-		err := jsonparser.ObjectEach(variables, func(key []byte, value []byte, dataType jsonparser.ValueType, offset int) error {
-			if dataType == jsonparser.Null {
+	// remove null variables from JSON: {"a":null,"b":1} -> {"b":1}
+	err := jsonparser.ObjectEach(variables, func(key []byte, value []byte, dataType jsonparser.ValueType, offset int) error {
+		if dataType == jsonparser.Null {
+			if removeNullVariables || slices.Contains(undefinedVariables, string(key)) {
```
devsergiy: as linter fails with generics we could use instead smth like this

```
type UndefinedVars []string

func(vars UndefinedVars) Contains(name string) (bool) {
  for i := 0; i < len(vars); i++ {
    if vars[i] == name {
      return true
    }
  }
  return false
}

**Review by spetrunin — CHANGES_REQUESTED** [review]

File: pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go (hunk 2 above)
Comment (YuriBuerov) [reply pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go]: typo

File: pkg/engine/datasource/graphql_datasource/graphql_datasource.go (hunk 3 above)
Comment (YuriBuerov) [reply pkg/engine/datasource/graphql_datasource/graphql_datasource.go]: fixed by updating CI

File: pkg/graphql/execution_engine_v2_test.go [hunk 4]
```diff
@@ -887,6 +887,68 @@ func TestExecutionEngineV2_Execute(t *testing.T) {
 		expectedResponse: `{"data":{"heroes":["Human","Droid"]}}`,
 	}))
 
+	t.Run("execute operation with null and omitted input variables", runWithoutError(ExecutionEngineV2TestCase{
+		schema: func(t *testing.T) *Schema {
+			t.Helper()
+			schema := `
+             type Query {
```
devsergiy: spacing issue

pkg/graphql/execution_engine_v2_test.go
```diff
@@ -887,6 +887,68 @@ func TestExecutionEngineV2_Execute(t *testing.T) {
...
+		     }`
+			parseSchema, err := NewSchemaFromString(schema)
+			require.NoError(t, err)
+			return parseSchema
+		}(t),
+		operation: func(t *testing.T) Request {
+			return Request{
+				OperationName: "MyHeroes",
+				Variables:     []byte(`{"height": null}`),
+				Query: `query MyHeroes($heroNames: [String!]!, $height: String){
```
devsergiy: probably will be better to have heroNames variable with nullable list as you have defined in schema `$heroNames: [String!]`

For the operation validation, it will be correct, but real upstream will return an error if required list will be null or undefined

spetrunin (approved):

pkg/graphql/execution_engine_v2_test.go
```diff
@@ -887,6 +887,68 @@ func TestExecutionEngineV2_Execute(t *testing.T) {
 		expectedResponse: `{"data":{"heroes":["Human","Droid"]}}`,
 	}))
 
+	t.Run("execute operation with null and omitted input variables", runWithoutError(ExecutionEngineV2TestCase{
```
jensneuse, line 890: Can you also add a test where we omit `{"height":null}` ? Or does this test already exist?

pkg/astvalidation/operation_rule_field_selection_merging.go
```diff
@@ -142,13 +142,10 @@ func (f *fieldSelectionMergingVisitor) EnterField(ref int) {
...
 			path = f.pathCache[f.pathCacheIndex][:len(f.Path)]
 			f.pathCacheIndex++
-			for i := 0; i < len(f.Path); i++ {
-				path[i] = f.Path[i]
-			}
 		} else {
 			path = make(ast.Path, len(f.Path))
-			copy(path, f.Path)
 		}
+		copy(path, f.Path)
```
jensneuse, line 148: can you explain this change?

pkg/astvalidation/operation_rule_field_selection_merging.go
```diff
@@ -197,13 +194,10 @@ func (f *fieldSelectionMergingVisitor) EnterField(ref int) {
...
 		path = f.pathCache[f.pathCacheIndex][:len(f.Path)]
 		f.pathCacheIndex++
-		for i := 0; i < len(f.Path); i++ {
-			path[i] = f.Path[i]
-		}
 	} else {
 		path = make(ast.Path, len(f.Path))
-		copy(path, f.Path)
 	}
+	copy(path, f.Path)
```
jensneuse, line 200: can you explain this change?

pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -1327,7 +1328,7 @@ type Source struct {
 	httpClient *http.Client
 }
 
-func (s *Source) compactAndUnNullVariables(input []byte) []byte {
+func (s *Source) compactAndUnNullVariables(input []byte, undefinedVariables ...string) []byte {
```
jensneuse: can we just make `undefinedVariables ` a slice of string instead of variadic?

pkg/engine/datasource/graphql_datasource/graphql_datasource.go
jensneuse: Is it possible to implement this function without casting from `[]byte` to `string`?

pkg/engine/resolve/inputtemplate.go
```diff
@@ -32,11 +33,14 @@ type InputTemplate struct {
 	// This is the case, e.g. when using batching and one sibling is null, resulting in a null value for one batch item
 	// Returning null in this case tells the batch implementation to skip this item
 	SetTemplateOutputToNullOnVariableNull bool
+	UndefinedVariables                    []string
 }
 
 var setTemplateOutputNull = errors.New("set to null")
 
 func (i *InputTemplate) Render(ctx *Context, data []byte, preparedInput *fastbuffer.FastBuffer) (err error) {
+	i.UndefinedVariables = make([]string, 0)
```
jensneuse: This InputTemplate is used concurrently. It's not possible to allocate a slice on it.
Can you move UndefinedVariables to the function scope?
Or outside of the function.
But I think it cannot stay on the "shared" InputTemplate.

jensneuse (changes requested):

pkg/graphql/execution_engine_v2_test.go
YuriBuerov: we omit `names` in this test, and `height` is set to `null` so it remains

pkg/astvalidation/operation_rule_field_selection_merging.go
YuriBuerov: not related to the task itself, but after updating the linter (to support generics), had to make this change

pkg/engine/datasource/graphql_datasource/graphql_datasource.go
YuriBuerov: sure, we will need to update all the data source tests after that

pkg/engine/datasource/graphql_datasource/graphql_datasource.go
YuriBuerov: do you mean `key`?  we need it to be a string to delete it (line 1362)

jensneuse: yes, but it should be possible without allocating, as we just read the value
e.g. you can use unsafebytes package to translate byte slice to string without allocation

pkg/engine/datasource/graphql_datasource/graphql_datasource.go
jensneuse: ok

jensneuse (approved):

PR #497, opened 2023-02-14
Fix validation for unknown non-nill field types

ikawaha:
Validation did not work for invalid schemas with fields of undefined type when the type was specified as non-null. 
see. [0a97177](https://github.com/wundergraph/graphql-go-tools/pull/497/commits/0a971770faa3fc0d06550c87b978c70ef367f499)

Field types specified as non-null are registered as TypeKindNonNull. The field type names of these fields must be referenced by ofType. However, this process was missing from the collection of  referenced types, so non-null types passed validation even though there were inconsistencies in the references.


ikawaha:
[0a97177](https://github.com/wundergraph/graphql-go-tools/pull/497/commits/0a971770faa3fc0d06550c87b978c70ef367f499) will fail.

jensneuse:
@ikawaha hey, thank you for the contribution! Are you able to add a test to prove the change works as described?

ikawaha:
I have noticed another problem with schema validation, so I would like to close this pr and submit an issue to report the problem. After that I will try to fix it again.

Thank you.

PR #500, opened 2023-02-15
feat: add support for declaring directives as repeatable

fiam:


spetrunin:
@fiam why do we need this?

The directive is repeatable in case it is defined as repeatable https://spec.graphql.org/October2021/#example-67869
We have already support for this feature
https://github.com/wundergraph/graphql-go-tools/blob/master/pkg/ast/ast_directive_definition.go#L22

PR #509, opened 2023-03-01
feat: Use complete channel for callbacks

bakjos:
The GraphQL SSE spec includes a complete event when the stream finishes, the current implementation keeps running even when the event is received. 

This PR adds support for receiving a callback when the complete event is received. I changed the API to include a channel to receive the complete event, wasn't sure if it was the right way or if we should add another method to the interfaces including the channel


bakjos:
@jensneuse Sure, this is the link https://github.com/graphql/graphql-over-http/blob/main/rfcs/GraphQLOverSSE.md#complete-event

StarpTech:
Closing due to inactivity.

PR #522, opened 2023-04-22
refactor: remove embedding of context.Context in resolve.Context

fiam:
Instead, use an approach very similar to the one used by net/http.Request


pkg/engine/resolve/inputtemplate.go
```diff
@@ -37,23 +36,28 @@ type InputTemplate struct {
...
 var setTemplateOutputNull = errors.New("set to null")
 
-func (i *InputTemplate) Render(ctx *Context, data []byte, preparedInput *fastbuffer.FastBuffer) (err error) {
-	undefinedVariables := make([]string, 0)
+func (i *InputTemplate) Render(ctx *Context, data []byte, preparedInput *fastbuffer.FastBuffer) error {
+	var undefinedVariables []string
 
-	for j := range i.Segments {
-		switch i.Segments[j].SegmentType {
+	for _, segment := range i.Segments {
```
jensneuse, line 43: The reason we've used index access here is to not create copies of the segments, which your rewrite now does. Effects might be minimal, but not zero. Do you have a good reason to use copies here?

I'm not opposed to copies, just want to understand why this refactor was necessary.

pkg/engine/resolve/resolve.go
```diff
@@ -174,24 +191,27 @@ func (c *Context) Clone() Context {
...
-		pathElements:    pathElements,
-		patches:         patches,
-		usedBuffers:     make([]*bytes.Buffer, 0, 48),
-		currentPatch:    c.currentPatch,
-		maxPatch:        c.maxPatch,
-		pathPrefix:      pathPrefix,
-		beforeFetchHook: c.beforeFetchHook,
-		afterFetchHook:  c.afterFetchHook,
-		position:        c.position,
+		ctx:                c.ctx,
```
jensneuse: Should we actually allow this? Might be another possible memory leak.
Wouldn't it be better if we change it to `Clone(ctx context.Context)` to force the user to explicitly set the context?
This way, they could create a new context with the content they want and pass it on.
Alternatively, we could not clone context.Context at all, leave Clone() as is, and provide CloneWithContext(ctx context.Context). WDYT?

pkg/engine/resolve/resolve.go
```diff
@@ -309,7 +338,7 @@ type DataSourceBatch interface {
 }
 
 type DataSource interface {
-	Load(ctx context.Context, input []byte, w io.Writer) (err error)
+	Load(ctx *Context, input []byte, w io.Writer) (err error)
```
jensneuse: You're aware that this will break all dependent DataSources?
Why is it necessary? Can we not just pass context.Context to DataSources like before?
Is there a reason why we need *resolve.Context in the Load function?

pkg/subscription/context.go
```diff
@@ -17,27 +18,51 @@ func NewInitialHttpRequestContext(r *http.Request) *InitialHttpRequestContext {
...
 	return true
 }
 
-func (sc subscriptionCancellations) CancelAll() {
-	for _, cancelFunc := range sc {
+func (sc *subscriptionCancellations) CancelAll() {
+	// Copy the values to temporary storage to avoid calling arbitrary functions
+	// with the lock held
+	sc.mu.RLock()
+	funcs := make([]context.CancelFunc, 0, len(sc.cancellations))
```
jensneuse: Why is this needed?

jensneuse (changes requested):

pkg/engine/resolve/resolve.go
fiam: Oh, good idea. Maybe we can even entirely remove `Clone()` in favour of `WithContext()` let me look into it and get back to you in a bit

pkg/subscription/context.go
fiam: Oh, I see! At the moment I was rewriting that I didn't realize we have full control of the cancellation functions, so my reasoning was that some function might call into one of the other mutex guarded methods and cause a deadlock, but now I see that's not possible. I'll remove the copy.

pkg/engine/resolve/resolve.go
fiam: I am aware that's an incompatible change. The reason I had to change the signature is because graphql_datasource.go requires state from the `resolve.Context` to remove the undefined variables https://github.com/wundergraph/graphql-go-tools/pull/522/files#diff-62c4a9dc18bfa9b5576bd76534407a914233956d05ed14837342302dd17168beL1411 - Can we store additional state in `input`parameter? That would allow us to carry the information about the undefined variables from here https://github.com/wundergraph/graphql-go-tools/pull/522/files#diff-b4d0466beb12697fa426019dad1c410175f54bf1466b252c6ac18033f402d266R74 and then we can pass just a `context.Context`

pkg/engine/resolve/resolve.go
fiam: The only usage of `.Clone()` in the whole package is in https://github.com/wundergraph/graphql-go-tools/blob/f29b89066be95f556ef7acdc1e67ba027d3a6d5f/pkg/engine/resolve/resolve.go#L826 - which looks fine to me since the contexts are cloned, executed in parallel and then we wait for all to finish before returning. However, this API is so easy to misuse that I would make it private by renaming it to `.clone()` and I would also add a comment to the method clarifying that it needs to be used with care. WDYT?

pkg/engine/resolve/resolve.go
fiam: I've just realized `dataLoader` also uses state from `resolve.Context` https://github.com/wundergraph/graphql-go-tools/blob/9fe3016fb1a70badd51f67b6923c4389b42a62df/pkg/engine/resolve/dataloader.go#L153 ~and this one would be way more complicated to refactor away~

fiam: `dataLoader` is easy to refactor because it does not need to conform to `DataSource`

jensneuse: Keep in mind that other data sources, outside of our repo, might have implemented this interface.

pkg/engine/resolve/resolve.go
fiam: I've renamed `.Clone()` to `.clone()`

pkg/engine/resolve/resolve.go
fiam: Since we were already changing `Context` and doing a breaking change, I assumed more breaking changes were fair game. However, thinking about this a bit more I realized we can carry the required state in the rendered data instead of passing it via the context, hence not needing a full `resolve.Context` anymore. PTAL

pkg/engine/resolve/inputtemplate.go
fiam: It is the other way around. With the range using the index you're making a new copy every time you read from the slice:

- 44:`switch i.Segments[j].SegmentType`
- 48:`switch i.Segments[j].VariableKind `
- 54:`i.Segments[j].VariableSourcePath`
- 56:`i.Segments[j].VariableKind`

(I'm not counting copies that pass  i.Segments[j] by value to another function because we cannot elide those)

Ranging with a value instead we make a copy at the start of the loop and then we use that across the whole iteration, creating 3 less copies.

In fact, the generated assembly seems to support this ('-' is ranging with value, '+' is ranging with index):

```diff
-github.com/wundergraph/graphql-go-tools/pkg/engine/resolve.(*InputTemplate).Render STEXT size=1504 args=0x30 locals=0x188 funcid=0x0 align=0x0
-       0x0000 00000 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   TEXT    github.com/wundergraph/graphql-go-tools/pkg/engine/resolve.(*InputTemplate).Render(SB), ABIInternal, $400-48
+github.com/wundergraph/graphql-go-tools/pkg/engine/resolve.(*InputTemplate).Render STEXT size=1680 args=0x30 locals=0x1e8 funcid=0x0 align=0x0
+       0x0000 00000 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   TEXT    github.com/wundergraph/graphql-go-tools/pkg/engine/resolve.(*InputTemplate).Render(SB), ABIInternal, $496-48
        0x0000 00000 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   MOVD    16(g), R16
        0x0004 00004 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   PCDATA  $0, $-2
-       0x0004 00004 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   SUB     $272, RSP, R17
+       0x0004 00004 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   SUB     $368, RSP, R17
        0x0008 00008 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   CMP     R16, R17
-       0x000c 00012 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   BLS     1436
+       0x000c 00012 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   BLS     1612
        0x0010 00016 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   PCDATA  $0, $-1
-       0x0010 00016 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   SUB     $400, RSP, R20
+       0x0010 00016 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   SUB     $496, RSP, R20
        0x0014 00020 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   STP     (R29, R30), -8(R20)
        0x0018 00024 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   PCDATA  $0, $-2
        0x0018 00024 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   MOVD    R20, RSP
        0x001c 00028 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   PCDATA  $0, $-1
        0x001c 00028 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   SUB     $8, RSP, R29
-       0x0020 00032 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   FUNCDATA        $0, gclocals·IL4UPPBcICQL/TGWCkBKmw==(SB)
-       0x0020 00032 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   FUNCDATA        $1, gclocals·sbbtSWTKH8TzPzMpN5aMwg==(SB)
+       0x0020 00032 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   FUNCDATA        $0, gclocals·DTMxmu7RUpolfv2Y2eSf3w==(SB)
+       0x0020 00032 (/Users/fiam/Source/WunderGraph/graphql-go-tools/pkg/engine/resolve/inputtemplate.go:40)   FUNCDATA        $1, gclocals·6Nlgdw1ZQ7pFG8AxIyzd6w==(SB)
```

Notice how with these changes the code size decreases from 1680 to 1504 and the locals siz decreases from 0x1e8 to 0x188. That's 96 bytes less, matching the 3 less copies of `InputTemplate` that we counted by hand, which is 32 bytes.

pkg/execution/datasource/datasource_http_polling_stream.go
```diff
@@ -110,11 +110,27 @@ type HttpPollingStreamDataSource struct {
...
 	ch       chan []byte
-	closed   bool
 	Delay    time.Duration
 	client   *http.Client
 	request  *http.Request
 	lastData []byte
+
+	// The mutex guards the fields following it. Use the
+	// accessor methods to read/write them.
+	mu     sync.RWMutex
```
jensneuse, line 120: Why is this needed? Do we access the field from multiple goroutines?

jensneuse (changes requested):

pkg/execution/datasource/datasource_http_polling_stream.go
fiam: `Resolve()` starts `startPolling()` on a different goroutine. The former reads and writes to the `closed`  field while the latter writes to it

jensneuse (approved):

PR #524, opened 2023-05-06
fix: if subgraph does not have type then ignore `__typename` field

Mistat:
This occurs when using Federation and Fragment with the `__typename` problem.

Example

SubGraphA
```
type Group {
  :
}
type Query {
  group(id: ID!) Group
}
```

SubGraphB
```
type Employee {
  :
}
type Query {
  employee(id: ID!) Group
}
```

Query
```
{
  group(id: "xxx") {
     :
  }
  employee(id: "xxx") {
     ... EmployeeItem
    __typename // <-- This would be an error
  }
}

fragment EmployeeItem on Employee {
 :
}
```

Then, Federation queries SubGraphA for the Employee type, which results in an error.


jensneuse:
Why closed?

Mistat:
> Why closed?

This change was a bad idea. I created another issue
https://github.com/wundergraph/graphql-go-tools/issues/525

PR #534, opened 2023-05-15
fix: replace jsonschema library

fiam:
github.com/qri-io/jsonschema has race conditions, bugs and seems to not
 be maintained anymore. Use github.com/santhosh-tekuri/jsonschema/v5 instead


devsergiy (approved):

devsergiy (changes requested):
tests are failing with this change

fiam:
> tests are failing with this change

Yep, on it

jensneuse (approved):

pkg/graphqljsonschema/jsonschema_test.go
```diff
@@ -10,6 +10,18 @@ import (
 	"github.com/wundergraph/graphql-go-tools/internal/pkg/unsafeparser"
 )
 
+func prettyPrint(s string) string {
+	var v interface{}
```
devsergiy, line 14: it could be simplified to
```
func prettyPrint(s string) string {
	var out bytes.Buffer
	if err := json.Indent(&out, []byte(s), "", "  "); err  != nil {
		panic(err)
	}
	return out.String()
}
```

fiam: Sorry, I missed this. I'll submit it as a separate PR.

PR #535, opened 2023-05-16
refactor: simply prettyPrint for JSON

fiam:
As suggested by @devsergiy


fiam:
@devsergiy I've realized this is not equivalent, because the decoding/encoding does normalize things like the field order and whitespace, which simplifies writing the test data while this approach doesn't. Closing.

PR #538, opened 2023-06-05
fix: provide more accurate error for invalid fragment spread

Aenimus:
Spreading a fragment on an invalid type should provide a more accurate error.

Also revert an accidental change of port for the federation example.


devsergiy (approved):

devsergiy (changes requested):
tests are failing

pkg/astvalidation/operation_rule_fragments.go
```diff
@@ -29,13 +29,18 @@ type fragmentsVisitor struct {
 }
 
 func (f *fragmentsVisitor) EnterFragmentSpread(ref int) {
-	spreadName := f.operation.FragmentSpreadNameBytes(ref)
+	fragmentName := f.operation.FragmentSpreadNameBytes(ref)
+	defer func() {
```
fiam: Would it be possible to avoid causing the panic without too big of a refactor?

devsergiy (approved):

fiam (approved):
LGTM

PR #550, opened 2023-06-30
fix for handling non existing field of input type in astnormalization

phob0s-pl:
This fix is for normalization of request where one of nested subfields of input type does not exist.

This can be checked by running such a query in input_coercion_for_list_test.go

```
query {
  inputWithListNestedList(input: {
    nested: {non_existing_field: true},
  }) {
    id
    name
  }
}
```

fixes https://github.com/wundergraph/graphql-go-tools/pull/559

### DIFF
--- a/pkg/astnormalization/astnormalization_test.go
+++ b/pkg/astnormalization/astnormalization_test.go
@@ -460,6 +460,42 @@ var run = func(normalizeFunc registerNormalizeFunc, definition, operation, expec
 	}
 }
 
+var runWithExpectedErrors = func(t *testing.T, normalizeFunc registerNormalizeVariablesFunc, definition, operation, expectedError string, additionalNormalizers ...registerNormalizeFunc) {
+	t.Helper()
+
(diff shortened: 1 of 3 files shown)


jensneuse:
Thanks for the contribution.
Will this issue not be caught by validating the Operation?

phob0s-pl:
@jensneuse  , no because flow of engine is following normalization->then validate against schema. 

```go
func (e *ExecutionEngineV2) Execute(ctx context.Context, operation *Request, writer resolve.FlushWriter, options ...ExecutionOptionsV2) error {
	if !operation.IsNormalized() {
		result, err := operation.Normalize(e.config.schema)
		if err != nil {
			return err
		}

		if !result.Successful {
			return result.Errors
		}
	}

	result, err := operation.ValidateForSchema(e.config.schema)
```

jensneuse:
I see, makes sense.
Can you add a test to cover this code change?

phob0s-pl:
@jensneuse Hi, I have added the test for that fix

devsergiy (approved):

PR #556, opened 2023-07-14
improve performance of batch fetching

phob0s-pl:
This fix improves drastically performance of enriching data in federated services. The inner for loop in jsonparser.ArrayEach is bottleneck if response to be enriched is large and has unique keys. 

The reason is size of responseMapping will grow to size of input, making computation n^2.
In our case the fix reduced time from 20minutes to 16 seconds for response with hundred thousand objects.


devsergiy (approved):

devsergiy:
@phob0s-pl thanks for the contribution!

PR #557, opened 2023-07-18
chore: improve error messages when base schema is not merged

fiam:
Now the error message says "operation type %s is not defined; did you forget to merge the base schema?"
where the placeholder is replaced by query/mutation/subscription

### DIFF
--- a/pkg/ast/ast_operation_definition.go
+++ b/pkg/ast/ast_operation_definition.go
@@ -17,6 +17,8 @@ const (
 	OperationTypeSubscription
 )
 
+// Name returns a human-readable operation name for the given OperationType.
+// If the operation is not one of the OperationType constants, it panics.
 func (t OperationType) Name() string {
 	switch t {
 	case OperationTypeUnknown:
@@ -28,7 +30,7 @@ func (t OperationType) Name() string {
 	case OperationTypeSubscription:
 		return "subscription"
 	}
-	return fmt.Sprintf("unknown operation type %d", int(t))
+	panic(fmt.Errorf("unknown operation type %d", int(t)))
 }
 
 type OperationDefinition struct {
--- a/pkg/ast/ast_operation_definition.go
+++ b/pkg/ast/ast_operation_definition.go
@@ -1,6 +1,7 @@
 package ast
 
 import (
+	"fmt"
 	"math"
 
 	"github.com/wundergraph/graphql-go-tools/internal/pkg/unsafebytes"
@@ -15,6 +16,20 @@ const (
 	OperationTypeMutation
 	OperationTypeSubscription
 )
+
+func (t OperationType) Name() string {
+	switch t {
+	case OperationTypeUnknown:
+		return "unknown"
+	case OperationTypeQuery:
+		return "query"
+	case OperationTypeMutation:
+		return "mutation"
+	case OperationTypeSubscription:
+		return "subscription"
+	}
+	return fmt.Sprintf("unknown operation type %d", int(t))
+}
 
 type OperationDefinition struct {
 	OperationType          OperationType      // one of query, mutation, subscription
--- a/pkg/astvisitor/visitor.go
+++ b/pkg/astvisitor/visitor.go
@@ -1348,6 +1348,7 @@ func (w *Walker) appendAncestor(ref int, kind ast.NodeKind) {
 				FieldName: literal.SUBSCRIPTION,
 			})
 		default:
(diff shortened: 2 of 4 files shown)


ggeijteman-wg (approved):

pkg/ast/ast_operation_definition.go
```diff
@@ -16,6 +17,20 @@ const (
 	OperationTypeSubscription
 )
 
+func (t OperationType) Name() string {
+	switch t {
+	case OperationTypeUnknown:
```
StarpTech, line 24: Why would someone use `OperationTypeUnknown` as OperationType?

pkg/ast/ast_operation_definition.go
```diff
@@ -16,6 +17,20 @@ const (
...
+		return "unknown"
+	case OperationTypeQuery:
+		return "query"
+	case OperationTypeMutation:
+		return "mutation"
+	case OperationTypeSubscription:
+		return "subscription"
+	}
+	return fmt.Sprintf("unknown operation type %d", int(t))
```
StarpTech: I'd expect an empty string or panic. It looks like a misuse to communicate an error.

pkg/ast/ast_operation_definition.go
fiam: To initialize a variable with an invalid value or to return it e.g.

```go
op := OperationTypeUnknown
switch operationString {
    case "query":
    op = OperationTypeQuery
    ....
    default:
         return nil, fmt.Errorf("unknown operation %s", operationString)
```

Additionally, having an invalid value as the zero-value is also useful because if you forget to initialize a variable/field of `OperationType` it doesn't end up with a valid value.

pkg/ast/ast_operation_definition.go
fiam: A `panic()` seemed a bit too harsh, but I have no strong feelings about it. I've changed it to panic, PTAL

pkg/ast/ast_operation_definition.go
StarpTech: That would imply that `OperationTypeUnknown` is a valid value but it's not. There is only `query`, `mutation`, `subscription`. For your other arguments. You can always start with `query` and return an error if nothing matches.

pkg/ast/ast_operation_definition.go
```diff
@@ -16,6 +17,22 @@ const (
...
+	case OperationTypeUnknown:
+		return "unknown"
+	case OperationTypeQuery:
+		return "query"
+	case OperationTypeMutation:
+		return "mutation"
+	case OperationTypeSubscription:
+		return "subscription"
+	}
+	panic(fmt.Errorf("unknown operation type %d", int(t)))
```
StarpTech, line 33: Let's return an empty string. You already tracking this as an error.

pkg/astvisitor/visitor.go
```diff
@@ -1349,6 +1350,10 @@ func (w *Walker) appendAncestor(ref int, kind ast.NodeKind) {
 		default:
 			return
 		}
+		if len(typeName) == 0 {
```
StarpTech, line 1354: Can this even happen? If `operationType` does not match with any case, it returns. Could you add a test?

StarpTech (changes requested):

pkg/ast/ast_operation_definition.go
fiam: An empty string is a bad idea. If it gets propagated to other places then it's very difficult to track where the unknown value came from. We can either panic or return a string that's easy to search for.

StarpTech: ok

pkg/astvisitor/visitor.go
fiam: Maybe we should return an error in the `default:` case. I'll take a look around. I'll add a test too.

fiam: I've looked around and it seems we can safely change that to stop parsing with an error if the operation is not valid in that switch. I've also added a test for the new error that this PR introduces. PTAL

StarpTech: Much better. Thanks for addressing it.

StarpTech (approved):
LGTM

pkg/astvisitor/visitor.go
fiam: 🫡

PR #559, opened 2023-07-21
Fix potential panic

ikawaha:
This PR fixes a potential panic that exists in the specification of variables to the query.

For example, in cases where a field that does not exist in the definition, but does exist in a variable.

Details:

If an undefined field of input type is in the given variable, `walkJsonObject()` returns -1.
```Go
func (i *inputCoercionForListVisitor) walkJsonObject(inputObjDefTypeRef int, data []byte) {
... snip
		inputValueDefRef := i.definition.InputObjectTypeDefinitionInputValueDefinitionByName(inputObjDefTypeRef, key)
		typeRef := i.definition.ResolveListOrNameType(i.definition.InputValueDefinitionType(inputValueDefRef))

```

This causes a panic by referencing the `InputValueDefinitions` array with a index of -1.

```Go
func (d *Document) InputValueDefinitionType(ref int) int {
	return d.InputValueDefinitions[ref].Type     // ← ★ ref=-1 !!! cause panic.
}
```


devsergiy:
Hi @ikawaha we will fix this issue by merging another contribution

It has a bit more clear error message
thanks a lot for double-checking this!

PR #567, opened 2023-07-31
feat: add support for BigInt

fiam:


pkg/engine/resolve/resolve.go
```diff
@@ -955,6 +958,27 @@ func (r *Resolver) resolveFloat(ctx *Context, floatValue *Float, data []byte, fl
...
+	case err != nil, valueType == jsonparser.Null:
+		if !bigIntValue.Nullable {
+			return errNonNullableFieldValueIsNull
+		}
+		r.resolveNull(bigIntBuf.Data)
+		return nil
+	case valueType == jsonparser.Number:
+		bigIntBuf.Data.WriteBytes(value)
+	case valueType == jsonparser.String:
+		bigIntBuf.Data.WriteBytes(value)
```
devsergiy, line 974: in case of string, you have to quote it when writing to buffer

value from json parser comes unquoted

```
	bigIntBuf.Data.WriteBytes(quote)
	bigIntBuf.Data.WriteBytes(value)
	bigIntBuf.Data.WriteBytes(quote)
```

fiam: Thanks! I've updated it, PTAL

devsergiy (approved):

PR #569, opened 2023-08-04
Bump dependencies

aerfio:
- github.com/cespare/xxhash/v2 and github.com/tidwall/sjson are not bumped because it breaks golden master tests
- github.com/imdario/mergo can be bumped to v1.0.0 if some dependency switched to new import path, for now v0.3.16 is as high as it can go


aerfio:
closing as it became stale

PR #603, opened 2023-09-18
Fix/planner fixes

devsergiy:


v2/pkg/ast/ast_selection.go
```diff
@@ -179,3 +179,26 @@ func (d *Document) SelectionSetHasFieldSelectionWithNameOrAliasBytes(set int, na
...
+		if d.Selections[i].Kind != SelectionKindField {
+			continue
+		}
+		fieldRef := d.Selections[i].Ref
+		fieldName := d.FieldNameBytes(fieldRef)
+		if !bytes.Equal(fieldName, name) {
+			continue
+		}
+
+		if d.FieldAliasIsDefined(fieldRef) {
```
Aenimus: how about this, to reduce nesting and remove an else?
```
if !d.FieldAliasIsDefined(fieldRef) {
  return true, fieldRef
  }

// other stuff
```

PR #613, opened 2023-09-19
Fix/typename alias and bypass scalars

devsergiy:
ENG-4066 fix aliasing __typename
ENG-4032 do not check scalars


v2/pkg/engine/resolve/resolve.go
```diff
@@ -415,6 +417,26 @@ func (r *Resolver) resolveBigInt(ctx *Context, bigIntValue *BigInt, data []byte,
...
+			return errNonNullableFieldValueIsNull
+		}
+		r.resolveNull(scalarBuf.Data)
+		return nil
+	case valueType == jsonparser.String:
+		scalarBuf.Data.WriteBytes(quote)
+		scalarBuf.Data.WriteBytes(value)
+		scalarBuf.Data.WriteBytes(quote)
+	default:
+		scalarBuf.Data.WriteBytes(value)
```
Aenimus, line 434: Is there a world where this is dangerous?

Aenimus (approved):

PR #624, opened 2023-09-29
Fix/planning interfaces with entity implementations

devsergiy:


v2/pkg/engine/resolve/load.go
```diff
@@ -451,7 +462,54 @@ func (l *Loader) resolveFetch(ctx *Context, fetch Fetch, res *resultSet) (err er
...
+		if fetch.Input.SkipErrItem {
+			err = nil
+			// skip fetch on render item error
+			return
+		}
+		return errors.WithStack(err)
+	}
+
+	switch {
+	case itemBuf.Len() == 4 && bytes.Equal(itemBuf.Bytes(), null):
```
jensneuse, line 488: Is it possible that we would want to fetch with null as item?

v2/pkg/engine/resolve/load.go
```diff
@@ -451,7 +462,54 @@ func (l *Loader) resolveFetch(ctx *Context, fetch Fetch, res *resultSet) (err er
...
+	}
+
+	_, _ = input.Write(itemBuf.Bytes())
+
+	err = fetch.Input.Footer.Render(ctx, nil, input)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	res.mergePath = fetch.PostProcessing.MergePath
```
jensneuse, line 503: Is this the right place to set mergePath?
I guess yes, just asking.
Seems better to abstract it away somehow.

v2/pkg/engine/plan/interface_selection_rewriter.go
```diff
@@ -0,0 +1,359 @@
...
+func (r *interfaceSelectionRewriter) isFieldReturnsInterface(fieldRef int, enclosingNode ast.Node) (interfaceDefRef int, isInterface bool) {
+	var (
+		fieldDefRef int
+		hasField    bool
+	)
+
+	interfaceDefRef = ast.InvalidRef
+
+	switch enclosingNode.Kind {
+	case ast.NodeKindObjectTypeDefinition:
```
jensneuse, line 42: Is this even necessary? You dismiss object types in line 51.

v2/pkg/engine/plan/interface_selection_rewriter.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	if node.Kind != ast.NodeKindInterfaceTypeDefinition {
+		return
+	}
+
+	return node.Ref, true
+}
+
+func (r *interfaceSelectionRewriter) objectTypesImplementingInterface(interfaceDefRef int) (typeNames []string, ok bool) {
```
jensneuse: this should be on ast.Document, I guess we already have it?

v2/pkg/engine/plan/interface_selection_rewriter.go
```diff
@@ -0,0 +1,359 @@
...
+type inlineFragmentSelection struct {
+	selectionRef int
+	typeName     string
+	fieldNames   []string
+}
+
+func (r *interfaceSelectionRewriter) collectFieldInformation(fieldRef int, entityNames []string) interfaceFieldSelectionInfo {
+	fieldSelectionSetRef, ok := r.operation.FieldSelectionSet(fieldRef)
+	if !ok {
+		panic("unexpected error: field does not have a selection set")
```
jensneuse: return and handle err?

v2/pkg/engine/plan/interface_selection_rewriter.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	fieldSelections := r.operation.SelectionSetFieldSelections(fieldSelectionSetRef)
```
jensneuse: Does this recursively find all nested fields?

v2/pkg/engine/plan/interface_selection_rewriter.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	inlineFragmentSelectionRefs := r.operation.SelectionSetInlineFragmentSelections(fieldSelectionSetRef)
+	inlineFragmentSelections := make([]inlineFragmentSelection, 0, len(inlineFragmentSelectionRefs))
+	for _, inlineFragmentSelectionRef := range inlineFragmentSelectionRefs {
+		inlineFragmentRef := r.operation.Selections[inlineFragmentSelectionRef].Ref
+		typeCondition := r.operation.InlineFragmentTypeConditionNameString(inlineFragmentRef)
+		inlineFragmentSelectionSetRef, ok := r.operation.InlineFragmentSelectionSet(inlineFragmentRef)
+		if !ok {
+			panic("unexpected error: inline fragment does not have a selection set")
```
jensneuse: handle err properly?

v2/pkg/engine/plan/interface_selection_rewriter.go
devsergiy: field could be on an interface or an object
```
type EnclosingObject {
  field: PossibleInterfaceType
}

type EnclosingInterface {
  field: PossibleInterfaceType
}
```

on line 51 I'm checking if a field type e.g. `PossibleInterfaceType` is an interface

v2/pkg/engine/plan/interface_selection_rewriter.go
devsergiy: no we don't, we have only `InterfaceTypeDefinitionImplementedByRootNodes` used on line 58 or we could check that particular object ref implements a particular interface

devsergiy: will move to ast

v2/pkg/engine/plan/interface_selection_rewriter.go
devsergiy: no, only top level

we do not need to go deep

PR #647, opened 2023-10-27
feat: take client request headers into account when deciding whether to reuse WS conns

fiam:
Otherwise it's not possible for the client to route headers through the engine because
a connection with different headers might be reused instead of creating a new one.
Instead, now we only reuse connections if the headers except Sec-Websocket-Key
and Content-Length match.

### DIFF
--- a/v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
+++ b/v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
@@ -11,6 +11,7 @@ import (
 	"github.com/buger/jsonparser"
 	"github.com/cespare/xxhash/v2"
 	"github.com/jensneuse/abstractlogger"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/resolve"
 	"nhooyr.io/websocket"
 )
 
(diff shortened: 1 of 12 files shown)


v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -181,22 +182,38 @@ func (c *SubscriptionClient) subscribeWS(reqCtx context.Context, options GraphQL
...
-	if err != nil {
+	if err := options.Header.Write(xxh); err != nil {
 		return 0, err
 	}
+	// The client request headers might be propagated to the outgoing
+	// request by the caller. We need to take them into account as well.
+	// However, to improve the changes of connection reuse we do skip
+	// some headers that are very likely to be different but very unlikely
+	// to cause different responses, otherwise no connections would be reused.
+	for key, values := range ctx.Request.Header {
```
jensneuse: I think we should have an allow list of headers that we want to add to the hash.
E.g. in most cases I think we want to limit this to just "Authorization".
This allow list could be configurable at the resolver level, and then we expose it through the cosmo router yaml.
WDYT?

I can also imagine use cases where we want the user to be authenticated, but the origin connection should not use the Authorization header. So the behavior should be configurable and potentially opt-out.

cc @StarpTech

jensneuse (changes requested):

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
StarpTech: I agree, it should be configurable and I like the proposal to maintain a waitlist rather than taking all headers into consideration. I'd definitely add `Authorization` by default to prevent data exposure and let the rest be use case specific.

fiam: I've refactored it so the caller can pass a list of headers. This allows us to make it automatic in the router side, by passing the headers declared in the forwarding rules. PTAL

jensneuse (approved):

PR #650, opened 2023-10-30
bug: astnormalization creates query that fails field merging

soonick:
One of my queries is failing with a field merging error. Created a test to reproduce.


soonick:
Closing this one since I created another pull request with the fix: https://github.com/wundergraph/graphql-go-tools/pull/653

PR #660, opened 2023-11-06
fix: make SubscriptionConfiguration work with go 1.20

fiam:
regexp.Regexp implements text marshal/unmarshal only in go 1.21 and
newer, requiring some workarounds in previous versions. To support this,
implement custom JSON marshalling methods that get compiled only when
using go 1.20 and below.


fiam:
We've decided top drop support for go 1.20 instead, closing.

Issue #495, opened 2023-02-07
Arguments on interface fields aren't passed to service


roderm:
If an interface has a field with arguments, a request to the interface "directly" always gets `nil` arguments. If queried via an inline fragment the parameters are passed to the subgraph.

example Schema:
```graphql
interface User {
	id: ID!
	email: String
	roles(tenant: String): [String]
}

type BaseUser implements User {
	id: ID!
	email: String
	roles(tenant: String): [String]
}
type Query {
	user: User
}
```

Working example inline fragment:
```graphql
query UserRoles($tenant: String){
	user {
		... on BaseUser {
			roles(tenant: $tenant)
		}
	}
}
```
The `tenant`-string is correctly passed to the resolver.

**not** working example:
```graphql
query UserRoles($tenant: String){
	user {		
		roles(tenant: $tenant)
	}
}
```
The tenant-string is not passed and will always be `nil` or `""`.

### Tests
Branch: https://github.com/roderm/graphql-go-tools/tree/roderm/abstract-field-argument

The problem seems to be in the [execution engine v2](https://github.com/roderm/graphql-go-tools/commit/3639fc39fbbaf0317a9e467e252a1417b3791b48#diff-318fda8fb3d273cbbbde93201ad6e54bf4774bd7f4522b19f323e884f2c7e82e):
* `query interface-field with argument and type specification` mimiks the inline fragment and works properly
* `query interface-field with argument` queries the fields on the interface and fails on `graphql_datasource.Factory`/`expectedBody`

I tried creating test-cases in [graphql-datasource](https://github.com/roderm/graphql-go-tools/commit/3639fc39fbbaf0317a9e467e252a1417b3791b48#diff-a9a261cb37ea0ad0b90f7e0fefa3e29adf43b0cd23ca1c3f7b528e34fba83834), which work as expected.

roderm:
As I can see now, a lot has changed and the test-cases work now. Unfortunately, it still doesn't work as an application. I tracked a few things down:
[...]

PR #662, opened 2023-11-07
feat: add interface-arguments

roderm:
Regarding to this issue here: https://github.com/wundergraph/graphql-go-tools/issues/495

* [ ] testing planner


roderm:
Hey there,
I tried to add some tests and implement this in the main package as well as in `v2`. In `v2` I was not able to run the tests in `schema_test.go`. I also created and tested this feature/bug in an own repository: https://github.com/roderm/graphql-gateway-examples/tree/master/example-1
* Most of the code is just copy-paste from `ObjectType` to `InterfaceType`. Would you prefer more generic functions and also are generic types okay?
* Are the tests sufficient or would you like some more tests?
* Are there any docs that needs to be updated?

devsergiy:
Hi @roderm 
are u still interested in this functionality? Once tests are green we could merged into v1 package
For the v2 your changes are not required, this code in v2 will be removed soon, as we handle it differently

Sorry for such a big delay, v1 branch is no longer under active support we are fully focused on v2

Please let us know what to do with your branch

Thanks,
Wundergraph Team

PR #679, opened 2023-11-23
fix duplicated path item in errors when object is null

devsergiy:


v2/pkg/astjson/astjson.go
```diff
@@ -374,7 +374,7 @@ func (j *JSON) AppendNonNullableFieldIsNullErr(fieldPath string, errorPath []Pat
 		ObjectFields: j.getIntSlice(),
 	})
 	errMessageStart, errMessageEnd := j.appendString("message")
-	errMessageValueStart, errMessageValueEnd := j.appendString(fmt.Sprintf("Cannot return null for non-nullable field %s.", fieldPath))
+	errMessageValueStart, errMessageValueEnd := j.appendString(fmt.Sprintf("Cannot return null for non-nullable field '%s'", fieldPath))
```
jensneuse: Add a "." after '%s'.
As:
```
"Cannot return null for non-nullable field '%s'".
```

jensneuse (changes requested):

PR #682, opened 2023-11-28
feat: add variablesvalidation package

jensneuse:


v2/pkg/variablesvalidation/variablesvalidation.go
```diff
@@ -0,0 +1,392 @@
...
+	var path string
+	if len(v.path) > 1 {
+		path = fmt.Sprintf(` at "%s"`, v.renderPath())
+	}
+	v.err = &InvalidVariableError{
+		Message: fmt.Sprintf(`Variable "$%s" got invalid value %s%s; Value "%s" does not exist in "%s" enum.`, variableName, invalidValue, path, string(enumValue), string(typeName)),
+	}
+}
+
+func (v *variablesVisitor) renderVariableEnumValueInvalidTypeError(typeName []byte, enumValue []byte) {
```
devsergiy: should this method be used anywhere?

v2/pkg/variablesvalidation/variablesvalidation.go
```diff
@@ -0,0 +1,392 @@
...
+		v.err = err
+		return
+	}
+	typeName := buf.String()
+	v.err = &InvalidVariableError{
+		Message: fmt.Sprintf(`Variable "$%s" got invalid value null; Expected non-nullable type "%s" not to be null.`, string(variableName), typeName),
+	}
+}
+
+func (v *variablesVisitor) traverseNode(jsonNodeRef int, fieldName, typeName, parentTypeName []byte) {
```
devsergiy: `fieldName` and `parentTypeName` is never used

v2/pkg/variablesvalidation/variablesvalidation.go
```diff
@@ -0,0 +1,392 @@
...
+		}
+		for _, field := range v.variables.Nodes[jsonNodeRef].ObjectFields {
+			inputFieldName := v.variables.ObjectFieldKey(field)
+			inputValueDefinitionRef := v.definition.InputObjectTypeDefinitionInputValueDefinitionByName(fieldTypeDefinitionNode.Ref, inputFieldName)
+			if inputValueDefinitionRef == -1 {
+				v.renderVariableFieldNotDefinedError(inputFieldName, typeName)
+				return
+			}
+		}
+	case ast.NodeKindScalarTypeDefinition:
```
devsergiy, line 351: if the object has a field with custom scalar `json: Scalar!` will we catch that field was null?

v2/pkg/variablesvalidation/variablesvalidation_test.go
```diff
@@ -0,0 +1,565 @@
...
+func TestVariablesValidation(t *testing.T) {
+
+	t.Run("required field argument not provided", func(t *testing.T) {
+		tc := testCase{
+			schema:    `type Query { hello(arg: String!): String }`,
+			operation: `query Foo($bar: String!) { hello }`,
+			variables: `{}`,
+		}
+		err := runTest(t, tc)
+		if err == nil {
```
devsergiy: could we use 
```
require.NoError(t, err)
or
require.Error(t, err)
```
instead of
```
	if err != nil {
		t.Fatal(err)
	}
```

devsergiy (changes requested):

devsergiy (approved):

PR #684, opened 2023-11-28
fix: redact headers in trace input

thisisnithin:


v2/pkg/engine/resolve/loader_test.go
```diff
@@ -592,6 +593,99 @@ func BenchmarkV2Loader_LoadGraphQLResponseData(b *testing.B) {
...
+	productsService := mockedDS(t, ctrl,
+		`{"method":"POST","url":"http://products","header":{"Authorization":["$$4$$"]},"body":{"query":"query{topProducts{name __typename upc}}"},"__trace__":true}`,
+		`{"topProducts":[{"name":"Table","__typename":"Product","upc":"1"},{"name":"Couch","__typename":"Product","upc":"2"},{"name":"Chair","__typename":"Product","upc":"3"}]}`)
+
+	response := &GraphQLResponse{
+		Data: &Object{
+			Fetch: &SingleFetch{
+				InputTemplate: InputTemplate{
+					Segments: []TemplateSegment{
+						{
```
jensneuse, line 899: Can you split up this template to use an actual header?

hints:

header variable:
https://github.com/wundergraph/graphql-go-tools/blob/9a35167d7f01ad6c90441e690d67152862106e64/v2/pkg/engine/resolve/resolve_test.go#L3761C39-L3761C39

example test with Authorization Header:
https://github.com/wundergraph/graphql-go-tools/blob/9a35167d7f01ad6c90441e690d67152862106e64/v2/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go#L125

Keep in mind that for $$0$$ to resolve, you need to use the postprocessing function on a plan, it turns the input into a template with multiple segments.

example of this:

before:
https://github.com/wundergraph/graphql-go-tools/blob/9a35167d7f01ad6c90441e690d67152862106e64/v2/pkg/engine/postprocess/datasourceinput_test.go#L24

after:
https://github.com/wundergraph/graphql-go-tools/blob/9a35167d7f01ad6c90441e690d67152862106e64/v2/pkg/engine/postprocess/datasourceinput_test.go#L173

without this postprocessing step, the resolver cannot properly resolve

jensneuse (changes requested):

jensneuse (approved):

PR #697, opened 2023-12-13
Handle three character strings

rwest202:
To reproduce, create a BigInt primary key column in a postgres database and query it using a graphql operation for findUnique. If the id is less than 10 and therefore a single character, ex: "1". This function fails to parse it.


fiam:
@rwest202 It seems to me this additional check shouldn't be necessary, removing the first and last characters works just fine for strings of 3 characters. See https://go.dev/play/p/7a2pmKrGFPN

rwest202:
You're right!

PR #729, opened 2024-01-25
Add go work

clayne11:
There are actually multiple go roots in this repo. Set up go work so the IDE can pick them all up and compile them properly.


jensneuse:
Btw have you tried to just open the v2 module (folder) in a separate instance of vs code? Maybe this solves the problem as well.

clayne11:
That partially helps, but then the examples won't compile. Each module would need to be opened independently which is a PITA. This solves the problem holistically and allows you to open the root and have every submodule work as expected.

clayne11:
Thanks!

Issue #618, opened 2023-09-26
Custom scalar cannot be nullable and cannot resolve null.


LimhorngSok:
Hi
I ran into a problem where I define a custom scalar called `Uint` and I have a schema like this

Game{
  playerOneID: Uint
  playerTwoID:  Uint
  WinnerID: Uint
}

When I tried to query Game and the WinnerID is null it always return `unable to resolve`
As I checked the code in lib, I found that my `CustomNode.Nullable is false`.

Here is my customScalar

type UintResolver struct {
	Nullable bool
}

// UintResolver implements resolve.CustomResolve.
func (UintResolver) Resolve(data []byte) ([]byte, error) {
	return data, nil
}

// Create resolver map
var customResolveMap = map[string]resolve.CustomResolve{
	"Uint": &UintResolver{Nullable: true},
}

or is there a way to make this CustomNode.Nullable = true?

BenjaminYong:
Did you ever find out how to do this @LimhorngSok ? Running into a similar issue with custom scalar is null, the entire object response becomes null with it:


```
query getScopes {
  getScopes {
    scope
    type
    jsonScopeArgs   # <-- JSON custom scalar (same resolver as yours)
    __typename
  }
}
```

```
{
  "data": {
    "getScopes": [
      null,
      null,
      null,
      null,
      null,
      null,
      null,
      null,
      null
    ]
  }
}
```

If I dont return the custom scalar, things return fine. Otherwise, if the custom scalar is not null, it will also return fine

Issue #179, opened 2020-02-11
There should be a way of defining custom scalars


jensneuse:
Custom scalars, e.g. JSON, Date etc. should be able to be defined and registered to the engine from outside the engine itself. That is, possible scalars should be externalized and able to be configured.

lyueyang:
Hello! Firstly, I'd like to start by saying thank you for writing such a brilliant library!

I would just like to know if there are any updates regarding this as I'm currently faced with the following situation:

I've written 2 custom scalars in gqlgen, 1 which unmarshals to a string (named Point) and another for uint64 for my subgraphs. The gateway has no issues handling the Point scalar but it has issues resolving fields which use uint64 as its type.

I've spent some time looking at how the nodes are resolved and so far I've observed that the uint64 node is resolved as a String when it should be an Integer. This issue doesn't occur when I update the schemas in my subgraphs to use Int instead of Uint64.

Thank you!

jensneuse:
It's definitely something that needs a solution at some point in the future.
Could you explain a bit more about your use case? I like to understand how devs use the library.

On the concrete example:
How do we communicate between GraphQL Schema (SDL) and engine what parser (e.g. uint64) to use for a field?
Do we simply assume that "uint64" and "Uint64" should be resolved by the corresponding resolver?

If you want to speed up the implementation, you could add failing tests and I'll have a look at implementing it.
That said, I cannot guarantee any timeline as this feature is currently no high priority.

Obviously, you're free to raise a PR with the full implementation but I assume it's hard for an "outsider" to find the right place and the right level of abstraction. It needs changing of the planner as well as the execution engine which means finding a good implementation is a bit complex if you're not familiar with the codebase. However, writing a failing test should be doable.

lyueyang:
Thank you for the reply! 

In my current use case, I have my scalars written as a separate library all together, similar to how gqlgen has their scalars [here](https://github.com/99designs/gqlgen/tree/master/graphql). In such a case, would it be possible to have the gateway to be aware of these scalars?

jensneuse:
> Thank you for the reply!
> 
> In my current use case, I have my scalars written as a separate library all together, similar to how gqlgen has their scalars [here](https://github.com/99designs/gqlgen/tree/master/graphql). In such a case, would it be possible to have the gateway to be aware of these scalars?

So if I understand you correctly it would be sufficient if we do the following:
- create an interface to marshal & unmarshal a scalar
[...]

PR #741, opened 2024-02-01
fix: allow resolve of nullable custom resolvers

BenjaminYong:
Hi there @devsergiy,

I am working with custom resolvers for custom scalars and found that they cannot be nullable on resolve. If nullable, the entire object becomes null with it or there may be an error (unable to resolve).

I can see the nullable attribute is not getting passed into the creation of resolve.CustomNode.
When adding this in, it inherits the nullable attribute from the schema and works as intended.

This should fix issue: https://github.com/wundergraph/graphql-go-tools/issues/618
And address the latest comment here: https://github.com/wundergraph/graphql-go-tools/issues/179

Thanks,
Ben


jensneuse:
Hey, I think this looks fine. Can you add 2 tests that prove functionality? nullable and non-nullable?

BenjaminYong:
> Hey, I think this looks fine. Can you add 2 tests that prove functionality? nullable and non-nullable?

Hey @jensneuse , that's been added now. Thanks for checking!

jensneuse (approved):

BenjaminYong:
Thanks for the approval @jensneuse , any ideas about the failing CI for windows test? Something going on with subscriptions in github.com/wundergraph/graphql-go-tools/pkg/engine/datasource/graphql_datasource. A breaking change in dependencies?

Seems to raise the same errors for me in master and my fix branch

PR #750, opened 2024-02-22
Sergiy/eng 4788 improve query planning time

devsergiy:


v2/pkg/ast/path.go
```diff
@@ -96,26 +97,42 @@ func (p Path) String() string {
 }
 
 func (p Path) DotDelimitedString() string {
-	out := ""
+	builder := strings.Builder{}
```
jensneuse, line 100: What if you use a bytes buf pool with a pre-defined size (toGrow) or a string.Builder pool that you grow to the required size? Currently, you're creating one buffer per func call.

v2/pkg/engine/plan/datasource_filter_collect_nodes_visitor.go
```diff
@@ -0,0 +1,132 @@
...
+			itemIds = append(itemIds, len(f.nodes.items)-1)
+		}
+	}
+
+	parentNodeId := f.currentParentID()
+	currentNodeId := TreeNodeID(ref)
+
+	added, exists := f.nodes.responseTree.Add(currentNodeId, parentNodeId, itemIds)
+	f.parentNodeIds = append(f.parentNodeIds, currentNodeId)
+	_, _ = added, exists
```
jensneuse: What's this for?

v2/pkg/engine/plan/datasource_filter_collect_nodes_visitor.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func (f *collectNodesVisitor) currentParentID() uint {
+	return f.parentNodeIds[len(f.parentNodeIds)-1]
+}
+
+func TreeNodeID(fieldRef int) uint {
+	return uint(100 + fieldRef)
```
jensneuse, line 126: What does 100 + mean?

v2/pkg/engine/plan/datasource_filter_node_suggestions.go
```diff
@@ -0,0 +1,304 @@
...
+
+type NodeSuggestions struct {
+	items           []*NodeSuggestion
+	pathSuggestions map[string][]*NodeSuggestion
+	seenFields      map[int]struct{}
+	responseTree    tree.Tree[[]int]
+}
+
+func NewNodeSuggestions() *NodeSuggestions {
+	return NewNodeSuggestionsWithSize(32)
```
jensneuse, line 67: Why 32?

v2/pkg/engine/plan/datasource_filter_node_suggestions.go
```diff
@@ -0,0 +1,304 @@
...
+	for _, parentIdx := range parentNodeIndexes {
+		if f.items[parentIdx].DataSourceHash == f.items[idx].DataSourceHash {
+			return parentIdx, true
+		}
+	}
+
+	return -1, false
+}
+
+func (f *NodeSuggestions) printNodes(msg string) {
```
jensneuse, line 236: Is this for debugging purposes?

jensneuse (changes requested):

v2/pkg/ast/path.go
devsergiy: toGrow could not be predefined, it varies depending on a path size

reusing Builder is not possible, it uses unsafe string
https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/strings/builder.go;l=49

v2/pkg/engine/plan/datasource_filter_node_suggestions.go
devsergiy: some not zero size for queries with a small amount of nodes
we could not predict how many nodes we will have, so just some not zero default to start with

v2/pkg/engine/plan/datasource_filter_node_suggestions.go
devsergiy: yep, part of debug framework, used in a couple of places in planner and nodes filter

v2/pkg/engine/plan/datasource_filter_collect_nodes_visitor.go
devsergiy: removed and added comment why do we ignore them

v2/pkg/engine/plan/datasource_filter_collect_nodes_visitor.go
devsergiy: added comment

jensneuse (approved):

PR #757, opened 2024-03-19
feat: support multiple edfs sources

Aenimus:


v2/pkg/engine/datasource/pubsub_datasource/pubsub_datasource.go
```diff
@@ -209,7 +210,7 @@ func (p *Planner) ConfigureSubscription() plan.SubscriptionConfiguration {
 		Input:     fmt.Sprintf(`{"topic":"%s"}`, p.current.topic),
 		Variables: p.variables,
 		DataSource: &SubscriptionSource{
-			pubSub: p.pubSub,
+			pubSub: p.pubSubBySourceName[p.current.config.SourceName],
 		},
```
devsergiy, line 222: We probably should have an additional check here that we have found source name in a map

If no, we should do p.Visitor.StopWithError

same relates to other panics, better have stop with error than panic

devsergiy (changes requested):

v2/pkg/engine/datasource/pubsub_datasource/pubsub_datasource.go
jensneuse: agreed

jensneuse (changes requested):

jensneuse (approved):

devsergiy (approved):

PR #758, opened 2024-03-20
feat: support file upload in router

pedraumcosta:
This PR makes possible to support a file upload according to spec:

https://github.com/jaydenseric/graphql-multipart-request-spec

With this PR merged to main and released we can discuss and work on PR to support the feature in cosmo router:

https://github.com/wundergraph/cosmo/pull/652

Basically, this PR makes possible to pass a temporary file information stored in file system to structures responsible for graphql resolve operation. Additionally, introduces a method in nethttpclient.go to actually perform the multipart http request.


v2/pkg/engine/datasource/httpclient/nethttpclient.go
```diff
@@ -219,3 +224,247 @@ func respBodyReader(res *http.Response) (io.Reader, error) {
 		return res.Body, nil
 	}
 }
+
+func DoMultipartForm(
```
jensneuse, line 231: Can the Do func and DoMultipartForm func share more code?
They look like they have a lot in common.

v2/pkg/engine/datasource/introspection_datasource/source.go
```diff
@@ -16,7 +17,7 @@ type Source struct {
 	introspectionData *introspection.Data
 }
 
-func (s *Source) Load(ctx context.Context, input []byte, w io.Writer) (err error) {
+func (s *Source) Load(ctx context.Context, input []byte, files []httpclient.File, w io.Writer) (err error) {
```
jensneuse: Should we introduce "LoadWithFiles" which will only implemented by the GraphQL DataSource?
This would make the change less breaking and most DS don't have and will never have to worry about files.

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
```diff
@@ -9101,12 +9105,157 @@ func TestSource_Load(t *testing.T) {
...
+		}
+		assert.Equal(t, values[0].Filename, expectedFile.Name)
+		assert.Equal(t, values[0].Size, expectedFile.Size)
+	}
+}
+
+func TestLoadFiles(t *testing.T) {
+	t.Run("single file", func(t *testing.T) {
+		queryString := `mutation($file: Upload!){singleUpload(file: $file)}`
+		variableString := `{"file":null}`
```
jensneuse, line 9200: Can you explain why file us null?

v2/pkg/engine/resolve/loader.go
```diff
@@ -1395,13 +1395,13 @@ func (l *Loader) executeSourceLoad(ctx context.Context, source DataSource, input
 
 		// Prevent that the context is destroyed when the loader hook return an empty context
 		if res.loaderHookContext != nil {
-			res.err = source.Load(res.loaderHookContext, input, res.out)
+			res.err = source.Load(res.loaderHookContext, input, l.ctx.Files, res.out)
```
jensneuse: Here, we could call source.LoadWithFiles in case the source supports files and the client supplied files in the ctx, otherwise we just do source.Load. WDYT?

jensneuse (changes requested):

v2/pkg/engine/datasource/introspection_datasource/source.go
thisisnithin: I too think we should go that route. Too many data sources affected. Will make the change

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
pedraumcosta: Hey @thisisnithin , please let me jump and explain this one. @jensneuse This is the usual thing in GQL libraries. We bring this from our previous experience with file uploads without federation. The appollo client generate requests like this.

v2/pkg/engine/datasource/httpclient/nethttpclient.go
thisisnithin: Done

v2/pkg/engine/datasource/introspection_datasource/source.go
thisisnithin: Done

v2/pkg/engine/resolve/loader.go
thisisnithin: Done

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
```diff
@@ -9093,6 +9096,7 @@ func runTestOnTestDefinition(t *testing.T, operation, operationName string, expe
 func TestSource_Load(t *testing.T) {
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		body, _ := io.ReadAll(r.Body)
+		fmt.Println(string(body))
```
jensneuse: Is this obsolete?

jensneuse (approved):

PR #762, opened 2024-03-31
feat: loader hooks

StarpTech:
This is an implementation of loader hooks that are executed before every call to the downstream service and after the final response has been merged. This allows us to instrument the fetch call, including all errors in the Cosmo Router.

Rate limit or authorization errors won't trigger the hooks because no request is instantiated. It needs to be guaranteed that when `OnLoad` is called, also `OnFinished` is called, otherwise hooks might leak memory when resources are managed across them. As a Developer, you get access to the final `SubgraphError` and to all GraphQL subgraph errors. 

Usage example in the Cosmo Router: [router/core/engine_loader_hooks.go](https://github.com/wundergraph/cosmo/pull/675/files#diff-82612426f570e70d58fb5e2814d47ef0d8f6eda1b5c4928c85424f432aa1dd4b)

Example:
```go
ctx := &resolve.Context{}
ctx.SetEngineLoaderHooks(EngineLoaderHooks{})

type EngineLoaderHooks struct {}

func (f *EngineLoaderHooks) OnLoad(ctx context.Context, dataSourceID string) context.Context {

	if resolve.IsIntrospectionDataSource(dataSourceID) {
		return ctx
	}

	// E.g. create a span
    
    // Return a modified one to share data across hooks
	return ctx
}

func (f *EngineLoaderHooks) OnFinished(ctx context.Context, statusCode int, dataSourceID string, err error) {

	if resolve.IsIntrospectionDataSource(dataSourceID) {
		return
	}

   // Finish the span

	if err != nil {
		var subgraphError *resolve.SubgraphError

		if errors.As(err, &subgraphError) {
			// Extract downstream errors
			if len(subgraphError.DownstreamErrors) > 0 {
		}
	}
}
```

### DIFF
--- a/v2/pkg/engine/resolve/errors.go
+++ b/v2/pkg/engine/resolve/errors.go
@@ -1,6 +1,7 @@
 package resolve
 
 import (
+	"bytes"
 	"fmt"
 	"strings"
 )
@@ -45,24 +46,24 @@ func (e *SubgraphError) AppendDownstreamError(error *GraphQLError) {
 
 func (e *SubgraphError) Error() string {
 
-	var sb strings.Builder
+	var bf bytes.Buffer
 
 	if e.SubgraphName == "" {
-		sb.WriteString(fmt.Sprintf("Failed to fetch Subgraph at Path: '%s'", e.Path))
+		fmt.Fprintf(&bf, "Failed to fetch Subgraph at Path: '%s'", e.Path)
 	} else {
-		sb.WriteString(fmt.Sprintf("Failed to fetch from Subgraph '%s' at Path: '%s'", e.SubgraphName, e.Path))
+		fmt.Fprintf(&bf, "Failed to fetch from Subgraph '%s' at Path: '%s'", e.SubgraphName, e.Path)
 	}
 
 	if e.Reason != "" {
-		sb.WriteString(fmt.Sprintf(", Reason: %s.", e.Reason))
+		fmt.Fprintf(&bf, ", Reason: %s.", e.Reason)
 	} else {
-		sb.WriteString(".")
+		fmt.Fprintf(&bf, ".")
 	}
 
 	if len(e.DownstreamErrors) > 0 {
 
-		sb.WriteString("\n")
-		sb.WriteString("Downstream errors:\n")
+		fmt.Fprintf(&bf, "\n")
+		fmt.Fprintf(&bf, "Downstream errors:\n")
 
 		for i, downstreamError := range e.DownstreamErrors {
 			extensionCodeErrorString := ""
@@ -71,20 +72,20 @@ func (e *SubgraphError) Error() string {
 			}
 
 			if len(downstreamError.Path) > 0 {
-				sb.WriteString(fmt.Sprintf("%d. Subgraph error at Path '%s', Message: %s", i+1, strings.Join(downstreamError.Path, ","), downstreamError.Message))
+				fmt.Fprintf(&bf, "%d. Subgraph error at Path '%s', Message: %s", i+1, strings.Join(downstreamError.Path, ","), downstreamError.Message)
 			} else {
-				sb.WriteString(fmt.Sprintf("%d. Subgraph error with Message: %s", i+1, downstreamError.Message))
+				fmt.Fprintf(&bf, "%d. Subgraph error with Message: %s", i+1, downstreamError.Message)
 			}
 
 			if extensionCodeErrorString != "" {
-				sb.WriteString(fmt.Sprintf(", Extension Code: %s.", extensionCodeErrorString))
+				fmt.Fprintf(&bf, ", Extension Code: %s.", extensionCodeErrorString)
 			}
 
-			sb.WriteString("\n")
+			fmt.Fprintf(&bf, "\n")
 		}
 	}
 
-	return sb.String()
+	return bf.String()
 }
 
 func NewRateLimitError(subgraphName, path, reason string) *RateLimitError {
--- a/v2/pkg/engine/resolve/errors.go
+++ b/v2/pkg/engine/resolve/errors.go
@@ -50,13 +50,11 @@ func (e *SubgraphError) Error() string {
 	if e.SubgraphName == "" {
 		sb.WriteString(fmt.Sprintf("Failed to fetch Subgraph at path: '%s'", e.Path))
 	} else {
-		sb.WriteString(fmt.Sprintf("Failed to fetch Subgraph '%s' at path: '%s'", e.SubgraphName, e.Path))
+		sb.WriteString(fmt.Sprintf("Failed to fetch from Subgraph '%s' at path: '%s'", e.SubgraphName, e.Path))
 	}
 
 	if e.Reason != "" {
-		sb.WriteString(fmt.Sprintf(", Reason: %s.", e.Reason))
-	} else {
-		sb.WriteString(".")
+		sb.WriteString(fmt.Sprintf(", Reason: %s", e.Reason))
 	}
 
 	if len(e.DownstreamErrors) > 0 {
@@ -103,7 +101,7 @@ type RateLimitError struct {
 
 func (e *RateLimitError) Error() string {
 	if e.Reason == "" {
-		return fmt.Sprintf("Rate limit rejected for subgraph '%s' at path '%s'.", e.SubgraphName, e.Path)
+		return fmt.Sprintf("Rate limit rejected for Subgraph '%s' at path '%s'", e.SubgraphName, e.Path)
 	}
-	return fmt.Sprintf("Rate limit rejected for subgraph '%s' at path '%s'. Reason: %s.", e.SubgraphName, e.Path, e.Reason)
+	return fmt.Sprintf("Rate limit rejected for Subgraph '%s' at path '%s', Reason: %s", e.SubgraphName, e.Path, e.Reason)
 }
--- a/v2/pkg/engine/resolve/errors.go
+++ b/v2/pkg/engine/resolve/errors.go
@@ -48,13 +48,15 @@ func (e *SubgraphError) Error() string {
 	var sb strings.Builder
 
 	if e.SubgraphName == "" {
-		sb.WriteString(fmt.Sprintf("Failed to fetch Subgraph at path: '%s'", e.Path))
+		sb.WriteString(fmt.Sprintf("Failed to fetch Subgraph at Path: '%s'", e.Path))
 	} else {
-		sb.WriteString(fmt.Sprintf("Failed to fetch from Subgraph '%s' at path: '%s'", e.SubgraphName, e.Path))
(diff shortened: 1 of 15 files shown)


v2/pkg/engine/resolve/errors.go
```diff
@@ -0,0 +1,109 @@
...
+func (e *SubgraphError) AppendDownstreamError(error *GraphQLError) {
+	e.DownstreamErrors = append(e.DownstreamErrors, error)
+}
+
+func (e *SubgraphError) Error() string {
+
+	var sb strings.Builder
+
+	if e.SubgraphName == "" {
+		sb.WriteString(fmt.Sprintf("Failed to fetch Subgraph at path: '%s'", e.Path))
```
jensneuse: Can you match the wording with the existing errors?
https://github.com/wundergraph/cosmo/blob/8aa618e478fee433c6fc4d8311205b9a234bddb5/router-tests/integration_test.go#L765

v2/pkg/engine/resolve/errors.go
```diff
@@ -0,0 +1,109 @@
...
+	} else {
+		sb.WriteString(fmt.Sprintf("Failed to fetch Subgraph '%s' at path: '%s'", e.SubgraphName, e.Path))
```
jensneuse: Can you use the `fmt.Fprintf` instead?

v2/pkg/engine/resolve/errors.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	if e.Reason != "" {
+		sb.WriteString(fmt.Sprintf(", Reason: %s.", e.Reason))
```
jensneuse: fmt.Fprintf

v2/pkg/engine/resolve/errors.go
```diff
@@ -0,0 +1,109 @@
...
+		sb.WriteString("Downstream errors:\n")
+
+		for i, downstreamError := range e.DownstreamErrors {
+			extensionCodeErrorString := ""
+			if downstreamError.Extensions != nil && downstreamError.Extensions.Code != "" {
+				extensionCodeErrorString = downstreamError.Extensions.Code
+			}
+
+			if len(downstreamError.Path) > 0 {
+				sb.WriteString(fmt.Sprintf("%d. Subgraph error at path '%s', Message: %s", i+1, strings.Join(downstreamError.Path, ","), downstreamError.Message))
```
jensneuse: fmt.Fprintf

v2/pkg/engine/resolve/errors.go
```diff
@@ -0,0 +1,109 @@
...
+			} else {
+				sb.WriteString(fmt.Sprintf("%d. Subgraph error, Message: %s", i+1, downstreamError.Message))
```
jensneuse: fmt.Fprintf

v2/pkg/engine/resolve/errors.go
```diff
@@ -0,0 +1,109 @@
...
+			}
+
+			if extensionCodeErrorString != "" {
+				sb.WriteString(fmt.Sprintf(", Extension Code: %s", extensionCodeErrorString))
```
jensneuse: fmt.Fprintf

v2/pkg/engine/resolve/loader.go
```diff
@@ -182,13 +194,22 @@ func (l *Loader) resolveAndMergeFetch(fetch Fetch, items []int) error {
 	switch f := fetch.(type) {
 	case *SingleFetch:
 		res := &result{
-			out: pool.BytesBuffer.Get(),
+			out:         pool.BytesBuffer.Get(),
+			loadContext: context.Background(),
```
jensneuse: Should this wrap the engine context? (so that the engine can cancel the load call)

v2/pkg/engine/resolve/loader.go
```diff
@@ -211,7 +232,9 @@ func (l *Loader) resolveAndMergeFetch(fetch Fetch, items []int) error {
 		g, ctx := errgroup.WithContext(l.ctx.ctx)
 		for i := range f.Fetches {
 			i := i
-			results[i] = &result{}
+			results[i] = &result{
+				loadContext: context.Background(),
```
jensneuse: wrap engine context?

v2/pkg/engine/resolve/loader.go
```diff
@@ -246,7 +275,8 @@ func (l *Loader) resolveAndMergeFetch(fetch Fetch, items []int) error {
 		for i := range items {
 			i := i
 			results[i] = &result{
-				out: pool.BytesBuffer.Get(),
+				out:         pool.BytesBuffer.Get(),
+				loadContext: context.Background(),
```
jensneuse: wrap engine context?

v2/pkg/engine/resolve/loader.go
```diff
@@ -258,28 +288,41 @@ func (l *Loader) resolveAndMergeFetch(fetch Fetch, items []int) error {
...
+			}
 			if err != nil {
 				return errors.WithStack(err)
 			}
 		}
 	case *EntityFetch:
 		res := &result{
-			out: pool.BytesBuffer.Get(),
+			out:         pool.BytesBuffer.Get(),
+			loadContext: context.Background(),
```
jensneuse: wrap engine context?

v2/pkg/engine/resolve/loader.go
```diff
@@ -302,7 +345,8 @@ func (l *Loader) loadFetch(ctx context.Context, fetch Fetch, items []int, res *r
 		for i := range items {
 			i := i
 			results[i] = &result{
-				out: pool.BytesBuffer.Get(),
+				out:         pool.BytesBuffer.Get(),
+				loadContext: context.Background(),
```
jensneuse: wrap engine context?

jensneuse (changes requested):

v2/pkg/engine/resolve/errors.go
StarpTech: Could you explain? I have no stream to write into. I use a string builder instead of a string.

v2/pkg/engine/resolve/loader.go
StarpTech: I forgot to remove this. We already pass the engine ctx to the load call. This is the context returned from the hook. I see the confusion. I renamed it.

v2/pkg/engine/resolve/errors.go
StarpTech: I have aligned all errors. All end with a punctuations, keywords are uppercased and `Reason:` is set when a reason exists.

```
Failed to fetch from Subgraph '0' at Path 'query', Reason: empty response.
```

v2/pkg/engine/resolve/errors.go
jensneuse: sb is your buffer

v2/pkg/engine/resolve/errors.go
```diff
@@ -0,0 +1,109 @@
+package resolve
+
+import (
+	"fmt"
+	"strings"
+)
+
+type GraphQLError struct {
```
devsergiy, line 9: There is an [ExternalError](https://github.com/wundergraph/graphql-go-tools/blob/04d41b42261ef2e29c6857a6fe544323011444af/v2/pkg/operationreport/externalerror.go#L31) type in operationreport package, which we probably could reuse to not create the same type again

v2/pkg/engine/resolve/errors.go
```diff
@@ -0,0 +1,109 @@
...
+	Message    string          `json:"message"`
+	Locations  []Location      `json:"locations,omitempty"`
+	Path       []string        `json:"path"`
+	Extensions *ErrorExtension `json:"extensions,omitempty"`
```
devsergiy: Extension could contain any field

spec says next
```
GraphQL services may provide an additional entry to errors with key extensions. This entry, if set, must have a map as its value. 
```

So even code is not guaranteed to be present

Could we make an extension field a map?

v2/pkg/engine/resolve/loader.go
```diff
@@ -24,6 +25,17 @@ import (
...
 
+type LoaderHooks interface {
+	// OnLoad is called before the fetch is executed
+	OnLoad(ctx context.Context, dataSourceID string) context.Context
+	// OnFinished is called after the fetch has been executed and the response has been processed and merged
+	OnFinished(ctx context.Context, statusCode int, dataSourceID string, err error)
+}
+
+func IsIntrospectionDataSource(dataSourceID string) bool {
+	return dataSourceID == "introspection__schema&__type" || dataSourceID == "introspection__type__fields" || dataSourceID == "introspection__type__enumValues"
```
devsergiy: could we move these string values into constants in this package and reuse them in the introspection datasource?

devsergiy (changes requested):

v2/pkg/engine/resolve/errors.go
StarpTech: I tried but `ExternalError` is not the representation we need to serialize the error. `Path` is from type `ast.Path` but I need `[]string`

v2/pkg/engine/resolve/errors.go
StarpTech: For my interface, I'm only interested in the code. Once we have the requirement to serialize more, we could change it.

StarpTech: But I can change it, it makes the code a bit more verbose.

v2/pkg/engine/resolve/loader.go
StarpTech: My plan was to put this in the same package where we defined the datasources but I run into circular dependencies. I can export them as const in this pkg.

StarpTech: Updated, please check.

v2/pkg/engine/resolve/errors.go
devsergiy: hmm, basically we could adjust [RequestError](https://github.com/wundergraph/graphql-go-tools/blob/c281b9c1142317ba50053e8e87287377298f3283/v2/pkg/graphqlerrors/errors.go#L112)
Currently for the path it is using custom struct, but it is not mandatory

If you want I could adjust it for you purpose

v2/pkg/engine/resolve/loader.go
devsergiy: Yeah, I know about circulars, that is why I proposed to put it here

The missing part is to reuse these consts in datasource as IDs

v2/pkg/engine/resolve/errors.go
StarpTech: Please go ahead. Thank you.

v2/pkg/engine/resolve/loader.go
StarpTech: Done

jensneuse (approved):

devsergiy (approved):

PR #772, opened 2024-04-16
fix: merging response nodes

devsergiy:
- fix merging response nodes issue leading to "non-nullable field can't be null" errors
- fix duplicated fetches on complex abstract types queries
- add fetch input to trace for skipped fetches


StarpTech (approved):
LGTM

PR #776, opened 2024-04-18
fix: validate if a proper GraphQL response has been received

StarpTech:


v2/pkg/engine/resolve/resolve_test.go
```diff
@@ -2213,9 +2213,9 @@ func TestResolver_ResolveGraphQLResponse(t *testing.T) {
 					},
 				},
 			},
-		}, Context{ctx: context.Background()}, `{"errors":[{"message":"Array cannot represent non-array value.","path":[]}],"data":null}`
```
StarpTech, line 2216: These are internal errors that were exposed because we didn't validate the response.

v2/pkg/engine/resolve/authorization_test.go
```diff
@@ -205,7 +205,7 @@ func TestAuthorization(t *testing.T) {
 		resolveCtx := &Context{ctx: context.Background(), Variables: nil, authorizer: authorizer}
 
 		return res, resolveCtx,
-			`{"errors":[{"message":"Unauthorized request to Subgraph 'users' at Path 'query', Reason: Not allowed to fetch from users Subgraph."}],"data":null}`,
+			`{"errors":[{"message":"Unauthorized request to Subgraph 'users' at Path 'query', Reason: Not allowed to fetch from users Subgraph."},{"message":"Failed to fetch from Subgraph 'reviews' at Path 'query.me'.","extensions":{"errors":[{"message":"could not render fetch input","path":["me"]}]}},{"message":"Failed to fetch from Subgraph 'products' at Path 'query.me.reviews.@.product'.","extensions":{"errors":[{"message":"could not render fetch input","path":["me","reviews","@","product"]}]}}],"data":null}`,
```
jensneuse, line 208: Should we expose errors like this? Rendering the fetch input is an internal that I think should not be exposed.

v2/pkg/engine/resolve/loader.go
```diff
@@ -412,31 +412,45 @@ func (l *Loader) mergeResult(res *result, items []int) error {
...
 			if err != nil {
 				return errors.WithStack(err)
 			}
 		}
 	}
+
+	// We also check if any data is there to processed
 	if res.postProcessing.SelectResponseDataPath != nil {
 		node = l.data.Get(node, res.postProcessing.SelectResponseDataPath)
+		// Check if the data is an array and not null
```
jensneuse: This comment is wrong I think.
We should check if data is defined and if it's an object.

v2/pkg/engine/resolve/loader.go
```diff
@@ -412,31 +412,45 @@ func (l *Loader) mergeResult(res *result, items []int) error {
...
 		if !l.data.NodeIsDefined(node) {
```
jensneuse, line 442: Did you miss to check if data is an object?

v2/pkg/engine/resolve/authorization_test.go
jensneuse: I think a better solution could be to determine what is the cause of error during the error, e.g. wrong input.
In general, such an error should almost never happen because we validate the input from the user.
So the only possibility is that we've got data back from a Subgraph that's invalid, so I think such a scenario is more like an internal error. We should log it, but not expose too many details to the client.

jensneuse (changes requested):

v2/pkg/engine/resolve/authorization_test.go
StarpTech: Those errors have been propagated because I configured the `SelectResponseErrorsPath` in the fetch config. My PR did not touch this.

v2/pkg/engine/resolve/loader.go
StarpTech: I updated it. The response is validated before https://github.com/wundergraph/graphql-go-tools/pull/776/files#diff-4a656e1687222772aeae74410688ea6636143a823951dff49c01736b72294bd2R419

v2/pkg/engine/resolve/loader.go
StarpTech: No, the response from the subngraph is validated before 

```go
// We expect a JSON object according to GraphQL over HTTP spec
node, err := l.data.AppendObject(res.out.Bytes())
```

StarpTech: Updated. It seems in some scenarios we also work with arrays. I moved the validation to the `SelectResponseDataPath` path.

v2/pkg/engine/resolve/authorization_test.go
StarpTech: Also don't forget those tests are running with `PropagateSubgraphErrors` enabled.

jensneuse (approved):

PR #782, opened 2024-04-23
feat: support multiple edfs argument templates

Aenimus:


v2/pkg/engine/datasource/pubsub_datasource/pubsub_datasource_test.go
```diff
@@ -38,45 +39,89 @@ func TestPubSub(t *testing.T) {
...
-		helloMutation(id: String!, input: String!): String! @eventsPublish(subject: "helloMutation.{{ args.id }}")
+		helloMutation(userKey: UserKey!): User! @edfs__natsPublish(subject: "tenants.{{ args.userKey.tenantId }}.users.{{ args.userKey.id }}")
 	}
 
 	type Subscription {
-		helloSubscription(id: String!): String! @eventsSubscribe(subjects: ["helloSubscription.{{ args.id }}"])
-		subscriptionWithMultipleSubjects(firstId: String!, secondId: String!): String! @eventsSubscribe(subjects: ["firstSubscription.{{ args.firstId }}", "secondSubscription.{{ args.secondId }}"])
-	}`
+		helloSubscription(userKey: UserKey!): User! @edfs__natsSubscribe(subjects: ["tenants.{{ args.userKey.tenantId }}.users.{{ args.userKey.id }}"])
+		subscriptionWithMultipleSubjects(userKeyOne: UserKey!, userKeyTwo: UserKey!): User! @edfs__natsSubscribe(subjects: ["tenantsOne.{{ args.userKeyOne.tenantId }}.users.{{ args.userKeyOne.id }}", "tenantsTwo.{{ args.userKeyTwo.tenantId }}.users.{{ args.userKeyTwo.id }}"])
```
jensneuse, line 51: I'd suggest the convention `@edfs__nats_subscribe`

v2/pkg/engine/datasource/pubsub_datasource/pubsub_datasource_test.go
```diff
@@ -164,26 +241,44 @@ func TestPubSub(t *testing.T) {
...
+									Renderer: resolve.NewPlainVariableRendererWithValidation(`{"type":["integer"]}`),
+								},
+								&resolve.ContextVariable{
+									Path:     []string{"a", "id"},
+									Renderer: resolve.NewPlainVariableRendererWithValidation(`{"type":["integer"]}`),
+								},
 								&resolve.ContextVariable{
 									Path:     []string{"a"},
-									Renderer: resolve.NewPlainVariableRendererWithValidation(`{"type":["string"]}`),
+									Renderer: resolve.NewPlainVariableRendererWithValidation(`{"type":["object"],"properties":{"id":{"type":["integer"]},"tenantId":{"type":["integer"]}},"required":["id","tenantId"],"additionalProperties":false}`),
```
jensneuse, line 278: I think this should be a JSON Renderer but not 100% sure.
I'm not sure if Plain will render the object correctly.

v2/pkg/engine/datasource/pubsub_datasource/pubsub_datasource.go
```diff
@@ -321,38 +326,113 @@ func (s *RequestDataSource) Load(ctx context.Context, input []byte, w io.Writer)
...
+				break
+			}
+		}
+		if inputObjectDefinitionRef != -1 {
+			return "", fmt.Errorf(`the event subject "%s" defines the final nested input field "%s" in argument template #%d, but it is not a leaf type`, subject, argumentPath[len(argumentPath)-1], templateNumber+1)
+		}
+		if lastInputValueTypeRef == -1 {
+			return "", fmt.Errorf(`the event subject "%s" defines the final nested input field "%s" in argument template #%d, but it does not exist`, subject, argumentPath[len(argumentPath)-1], templateNumber+1)
+		}
+		argumentName := argumentPath[0]
```
jensneuse, line 399: Can this panic?

jensneuse (changes requested):

v2/pkg/engine/datasource/pubsub_datasource/pubsub_datasource_test.go
```diff
@@ -104,6 +149,28 @@ func TestPubSub(t *testing.T) {
...
+					TypeName:   "Subscription",
+					FieldNames: []string{"subscriptionWithArgTemplateAndStaticValue"},
+				},
+				{
+					TypeName:   "User",
+					FieldNames: []string{"id", "tenant"},
+				},
+			},
+			ChildNodes: []plan.TypeField{
+				{
```
devsergiy, line 163: User should not be duplicated in child nodes, in case it is already present in the root nodes

v2/pkg/engine/datasource/pubsub_datasource/pubsub_datasource_test.go
Aenimus: My only comment is that it does not conform to the naming convention of the other directives, which are `@{namespace}__{camelCaseName}`, e.g., `@interfaceObject`, `@requiresScopes`, etc.

If you still think `@{namespace}__{snake_case_name}`, I'll make the change.

v2/pkg/engine/datasource/pubsub_datasource/pubsub_datasource.go
Aenimus:
```
		if len(matches) != 2 {
			return "", fmt.Errorf("expected a single matching group for argument template")
		}
		// the path begins with ".", so ignore the first empty string element
		argumentPath := strings.Split(matches[1][1:], ".")
```

We only have a match if the regex matches, which must be `[a-zA-Z0-9_]` after a period, which means we should be guaranteed to have at least one item in the argument path. If we don't, the regex didn't match, so we'd never get there.

Either way, this tells me that some regex tests would be useful. I'll add them.

v2/pkg/engine/datasource/pubsub_datasource/pubsub_datasource_test.go
devsergiy: look like it should https://github.com/wundergraph/graphql-go-tools/blob/f87c0216b7b1289e6c289dfdf1cba30bd9620b7c/v2/pkg/engine/resolve/inputtemplate_test.go#L53

v2/pkg/engine/datasource/pubsub_datasource/pubsub_datasource.go
```diff
@@ -321,38 +326,113 @@ func (s *RequestDataSource) Load(ctx context.Context, input []byte, w io.Writer)
 	return s.pubSub.Request(ctx, subject, nil, w)
 }
 
-func (p *Planner[T]) extractEventSubject(ref int, subject string) (string, error) {
-	matches := eventSubjectRegex.FindAllStringSubmatch(subject, -1)
-	if len(matches) != 1 || len(matches[0]) != 2 {
-		return "", fmt.Errorf("expected subject to match regex")
+func (p *Planner[T]) inputObjectDefinitionRefByFieldDefinitionRef(fieldDefinitionRef int, argumentNameBytes []byte) (inputObjectDefinitionRef int, inputValueTypeRef int) {
+	inputValueTypeRef = -1
```
devsergiy, line 330: Could you use here and below `ast.InvalidRef` instead of `-1` literal

devsergiy (changes requested):

v2/pkg/engine/datasource/pubsub_datasource/pubsub_datasource_test.go
devsergiy: look like it should https://github.com/wundergraph/graphql-go-tools/blob/f87c0216b7b1289e6c289dfdf1cba30bd9620b7c/v2/pkg/engine/resolve/inputtemplate_test.go#L53

v2/pkg/engine/datasource/pubsub_datasource/pubsub_datasource_test.go
Aenimus: Yeah, I know. Not sure how I managed that. My mistake.

v2/pkg/engine/datasource/pubsub_datasource/pubsub_datasource_test.go
jensneuse: IMHO "edfs_nats_" is the namespace. Other opinions might be valid.
@StarpTech @devsergiy what's your take on naming here?

v2/pkg/engine/datasource/pubsub_datasource/pubsub_datasource.go
Aenimus: Yep. I'll change it.

v2/pkg/engine/datasource/pubsub_datasource/pubsub_datasource_test.go
StarpTech: I like to be consistent. All directives seems to follow camelCase. Another question is: Do we really need to namespace it? The directives can only be applied to subscriptions and there are already unique. What about?

```graphql
@natsSubscribe
@natsPublish
```

but I'm also good with

```graphql
@edfs__natsSubscribe
```

seeing `edfs__` as the namespace but still going camelCase.

v2/pkg/engine/datasource/pubsub_datasource/pubsub_datasource.go
```diff
@@ -321,38 +326,113 @@ func (s *RequestDataSource) Load(ctx context.Context, input []byte, w io.Writer)
...
+		}
+		// We need to replace the template literal with the variable placeholder (and reuse if it already exists)
+		variablePlaceHolder, _ := p.variables.AddVariable(contextVariable) // $$0$$
+		substitutedSubject = strings.ReplaceAll(substitutedSubject, matches[0], variablePlaceHolder)
 	}
-	contextVariable := &resolve.ContextVariable{
-		Path:     []string{string(variableName)},
-		Renderer: renderer,
+	// substitute the variable templates for dummy values to check that the string is a valid NATS subject
+	if natsserver.IsValidSubject(variableTemplateRegex.ReplaceAllLiteralString(substitutedSubject, "a")) {
```
StarpTech, line 432: Can we abstract it in a way that we don't get provider specific in the pubsub implementation? For Kafka, topics should be immutable.

PR #785, opened 2024-04-26
doc: fix mistakes in v2 tutorial

antoineco:
While going over the tutorial code, I spotted a few mistakes in the sample outputs and method calls.

I'm leaving a few comments on selected changes for consideration.


v2/README.md
```diff
@@ -101,7 +102,7 @@ func ExampleParsePrintDocument() {
 	if err != nil {
 		panic(err)
 	}
-	fmt.Println(out.String()) // Output: query { hello }
```
antoineco, line 104: The operation type (`query`) is no longer part of the output. It wasn't clear to me whether this is intentional or not.

v2/README.md
```diff
@@ -221,9 +222,9 @@ func ExampleParseOperationNameAndType() {
 			continue
 		}
 		operationCount++
-		name := document.RootOperationTypeDefinitionNameString(node.Ref)
+		name := document.OperationDefinitionNameString(node.Ref)
 		operationNames = append(operationNames, name)
-		operationType := document.RootOperationTypeDefinitions[node.Ref].OperationType
```
antoineco, line 226: `RootOperationTypeDefinitions` is an empty slice here. Likewise, it wasn't clear to me whether this is intentional or not.

v2/README.md
```diff
@@ -495,51 +489,62 @@ so that the planner knows how to create an execution plan for the DataSource and
...
+					TypeName:   "Query",
+					FieldNames: []string{"hello", "foo"},
+				},
+			},
+			ChildNodes: plan.TypeFields{
+				{
+					TypeName:   "Foo",
+					FieldNames: []string{"bar"},
+				},
+			},
```
antoineco, line 508: Without defining child nodes, the plan action fails with
```
panic: internal: nodesResolvableVisitor: could not select the datasource to resolve Query.foo on a path query.foo
```

v2/README.md
```diff
@@ -560,7 +565,7 @@ func ExampleExecuteOperation() {
 		if err != nil {
 			panic(err)
 		}
-		fmt.Println(out.String()) // Output: {"data":{"hello":"world"}}
+		fmt.Println(out) // Output: {"data":{"hello":"world"}}
```
antoineco, line 569: I am still getting an error instead of the suggested output here:
```json
{"errors":[{"message":"String cannot represent non-string value: \"\"","path":[]}],"data":null}
```
I'll need more time to trace why this is happening, so in case the cause is obvious to someone, feel free to point me in the right direction.

StarpTech:
Thank you for your contribution but we can't merge as it is. Closing for now.

PR #787, opened 2024-04-30
fix normalization of non compatible nested fragment selections

devsergiy:


v2/pkg/astnormalization/inline_selections_from_inline_fragments.go
```diff
@@ -25,25 +25,58 @@ func (m *inlineSelectionsFromInlineFragmentsVisitor) EnterDocument(operation, de
...
+	}
+
+	fragmentSelectionRefs := m.operation.SelectionSetInlineFragmentSelections(selectionSetRef)
+	if len(fragmentSelectionRefs) == 0 {
+		return true
+	}
+
+	// we could inline the current fragment only if all nested fragment types are of the same type as enclosing type
+	// or enclosing type implements nested fragment type
+	fragmentsValid := true
```
Aenimus: isFragmentValid reads better IMO

v2/pkg/astnormalization/inline_selections_from_inline_fragments.go
```diff
@@ -25,25 +25,58 @@ func (m *inlineSelectionsFromInlineFragmentsVisitor) EnterDocument(operation, de
...
+	for _, fragmentSelectionRef := range fragmentSelectionRefs {
+		nestedFragmentRef := m.operation.Selections[fragmentSelectionRef].Ref
+		nestedInlineFragmentTypeName := m.operation.InlineFragmentTypeConditionName(nestedFragmentRef)
+
+		validFragment := bytes.Equal(nestedInlineFragmentTypeName, enclosingTypeName) ||
+			m.definition.TypeDefinitionContainsImplementsInterface(enclosingTypeName, nestedInlineFragmentTypeName)
+
+		if !validFragment {
+			fragmentsValid = false
```
Aenimus: could just return validFragment immediately?

v2/pkg/astnormalization/inline_selections_from_inline_fragments.go
Aenimus: or areFragmentsValid if it's more than one

v2/pkg/astnormalization/inline_selections_from_inline_fragments.go
```diff
@@ -25,25 +25,58 @@ func (m *inlineSelectionsFromInlineFragmentsVisitor) EnterDocument(operation, de
...
+			break
+		}
+	}
 
-	return m.definition.TypeDefinitionContainsImplementsInterface(enclosingTypeName, inlineFragmentTypeName)
+	return fragmentsValid
```
Aenimus: Could you not return true here, and false on line 70? Then you don't need a second variable at all.

PR #790, opened 2024-05-08
feat: implement subscriptions filter

jensneuse:


v2/pkg/engine/plan/configuration_visitor.go
```diff
@@ -801,6 +804,101 @@ func (c *configurationVisitor) addNewPlanner(ref int, typeName, fieldName, curre
...
+			filter.Values[i].Segments = []resolve.TemplateSegment{
+				{
+					SegmentType: resolve.StaticSegmentType,
+					Data:        []byte(value),
+				},
+			}
+			continue
+		}
+		if len(matches) == 1 && len(matches[0]) == 4 {
+			prefix := value[:matches[0][0]]
```
devsergiy, line 870: could you add some explanations about indexes? [0][2] not so clear why it is like this

jensneuse: It's what the std Library gives back.
One match with 2 sub matches is a slice of one with a slice of 4, two indices marking start and end of the match.

devsergiy: please add input examples which regexp should match

jensneuse: done

PR #791, opened 2024-05-10
Sergiy/eng 5139 planning issues with required field

devsergiy:


Aenimus (approved):
AFAICTILGTM
(As Far As I Can Tell It Looks Good To Me :D)

PR #794, opened 2024-05-11
feat: allow to shutdown pubsub implementation

StarpTech:


Aenimus (approved):

jensneuse (changes requested):
You're not calling this interface anywhere. If you manage all implementations in another repository (Router), this interface should live in the Router repository I think. It doesn't seem like the lifecycle is managed by the engine. Alternatively, you could manage the lifecycle of pubsubs by the engine, in which case it would make sense. In idiomatic go, you usually put the interface declaration where you use it. So I guess you could just close this PR and move the interface to the Router where you also call Shutdown.  You could call it PubSubLifecycle or something similar and have all pubsubs implement it.

StarpTech:
> Alternatively, you could manage the lifecycle of pubsubs by the engine, in which case it would make sense

Currently, the engine does not have a "shutdown" routine. I think it is also better to decouple it from the engine because pubsub components are created outside the engine.

> n idiomatic go, you usually put the interface declaration where you use it. So I guess you could just close this PR and move the interface to the Router where you also call Shutdown. You could call it PubSubLifecycle or something similar and have all pubsubs implement it.

The reason why I put this in the engine is that both interfaces are very related to each other and I didn't want to spread it across two codebases. I agree that when the interface isn't used here that it makes less sense to host it here except for the reason I mentioned before. If this is the way, there many other types that can be moved. I find my way just pragmatic but I can make the change and move it to the router.

PR #795, opened 2024-05-12
fix: improve ws subprotocol selection

JivusAyrus:


v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -30,12 +30,25 @@ type subscriptionClient struct {
...
 	handlers                   map[uint64]ConnectionHandler
 	handlersMu                 sync.Mutex
-	wsSubProtocol              string
 	onWsConnectionInitCallback *OnWsConnectionInitCallback
 
 	readTimeout time.Duration
 }
 
+type InvalidWsSubprotocolError struct {
+	Message string
```
jensneuse: Instead of putting a generic string here, create public fields for expected and actual value, then this can be parsed in the Router and we don't have to deal with a generic string.

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -333,21 +338,25 @@ func (c *subscriptionClient) newWSConnectionHandler(reqCtx context.Context, opti
...
 
-	switch c.wsSubProtocol {
+	switch wsSubProtocol {
 	case ProtocolGraphQLWS:
 		return newGQLWSConnectionHandler(c.engineCtx, conn, c.readTimeout, c.log), nil
 	case ProtocolGraphQLTWS:
 		return newGQLTWSConnectionHandler(c.engineCtx, conn, c.readTimeout, c.log), nil
 	default:
-		return nil, fmt.Errorf("unknown protocol %s", conn.Subprotocol())
+		return nil, NewInvalidWsSubprotocolError(fmt.Sprintf("provided websocket subprotocol %s is not supported. The supported subprotocols are graphql-ws and graphql-transport-ws. Please configure your subsciptions with the mentioned subprotocols", wsSubProtocol))
```
jensneuse: You can move this formatting to the Error Object's Error() string func and just pass the subprotocol here.

jensneuse (changes requested):

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
devsergiy: Note: we will not parse it in the router

we just will log it

When websocket connection is established it lives in a separate goroutine, and we won't be able to communicate this error to the router itself

JivusAyrus: No, this error can be returned. This is while creating the wsConnectionHandler. The thing happening inside a seperate go routine is StartBlocking function stuff right?

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client_test.go
```diff
@@ -296,3 +293,206 @@ func TestWebsocketSubscriptionClientWithServerDisconnect(t *testing.T) {
...
+	serverCancel()
+	assert.Eventuallyf(t, func() bool {
+		client.handlersMu.Lock()
+		defer client.handlersMu.Unlock()
+		return len(client.handlers) == 0
+	}, time.Second, time.Millisecond, "client handlers not 0")
+}
+
+// In this case the subgraph doesnt return the subprotocol, so falls back to graphql-ws
+func TestSubprotocolNegotiationWithNoSubprotocol(t *testing.T) {
```
jensneuse, line 437: There's a case missing where the protocol is hardcoded through options to "graphql-transport-ws" or did I miss this?

jensneuse (changes requested):

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client_test.go
JivusAyrus: No, you are right. Done

jensneuse (approved):

PR #799, opened 2024-05-13
fix(edfs): stable sort subject

StarpTech:
Avoid creating multiple triggers when the topics are defined in different orders in the schema.


StarpTech:
I think it is smarter to do this at composition.

jensneuse (approved):

PR #801, opened 2024-05-14
fix: allow inlining intersecting fragments on interfaces

jensneuse:


v2/pkg/astnormalization/directive_include_skip_test.go
```diff
@@ -54,4 +54,178 @@ func TestDirectiveIncludeVisitor(t *testing.T) {
...
+							includeName: name
+						}
+					}
+					withAlias: dog {
+						name @include(if: $yes)
+					}
+				}`, `
+				query($yes: Boolean!) {
+					dog {
+						... {
```
jensneuse, line 71: @devsergiy do we have another rule that inlines fragments/inline fragments with no conditions?

devsergiy: yeah, it is fragment spread inlining modified in this PR

in order to apply it here

pass inline fragment as the main one and skip include as additional

jensneuse: That's fine, no need to test this here as well.

PR #804, opened 2024-05-17
fix: list argument explosion

Aenimus:


v2/pkg/engine/plan/configuration_visitor.go
```diff
@@ -890,7 +894,24 @@ func (c *configurationVisitor) buildSubscriptionFieldFilter(condition *Subscript
...
 			hasPrefix := len(prefix) > 0
-			variableName := value[matches[0][2]:matches[0][3]]
+			// the path begins with ".", so ignore the first empty string element with trailing [1:]
+			argumentPath := value[matches[0][2]:matches[0][3]][1:]
+			argumentPathSlice := strings.Split(argumentPath, ".")
+			argumentName := argumentPathSlice[0]
+			argumentRef, ok := c.operation.FieldArgument(c.fieldRef, []byte(argumentName))
+			if !ok {
+				// error
+				//fmt.Errorf(`field argument "%s" is not defined`, argumentName)
```
StarpTech: Please remove unused prints.

v2/pkg/engine/plan/configuration_visitor.go
```diff
@@ -797,6 +798,9 @@ func (c *configurationVisitor) addNewPlanner(ref int, typeName, fieldName, curre
 	// fetch id is an index of the current planner
 	fetchID := len(c.planners)
 
+	// the filter needs access fieldRef
+	c.fieldRef = ref
```
StarpTech, line 802: Does it needs to be reset?

StarpTech (changes requested):

v2/pkg/engine/plan/configuration_visitor.go
Aenimus: No, I forgot to propagate the errors.

v2/pkg/engine/plan/configuration_visitor.go
Aenimus: I don't think so, because it's per field. But @jensneuse or @devsergiy can weigh in.

devsergiy: it will be fine like this, but maybe we should add more precise comment where this param is used

alternative is to pass field ref into resolveSubscriptionFilterCondition

devsergiy (approved):

StarpTech (approved):
LGTM

StarpTech (approved):
LGTM

jensneuse (approved):

PR #805, opened 2024-05-18
fix(edfs): filter only when type matches

StarpTech:
This PR ensures that we respect the value type when comparing the argument value with the event payload. Before this PR, we treated "1" as 1 and vice versa, as well as "true" = true.

Changes in a nutshell: When we have one segment we can extract the type. If we have more than one it will always result in a byte value comparison. In case of array values, we can always get the type and compare it with the argument type.

### DIFF
--- a/v2/pkg/engine/resolve/inputtemplate.go
+++ b/v2/pkg/engine/resolve/inputtemplate.go
@@ -26,6 +26,7 @@ type TemplateSegment struct {
 	VariableSourcePath []string
 	Renderer           VariableRenderer
 	Segments           []TemplateSegment
+	VariableValueType  jsonparser.ValueType
 }
 
 type InputTemplate struct {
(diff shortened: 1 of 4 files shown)


v2/pkg/engine/resolve/inputtemplate.go
```diff
@@ -113,6 +114,9 @@ func (i *InputTemplate) renderObjectVariable(ctx context.Context, variables []by
 		_, _ = preparedInput.Write(literal.NULL)
 		return nil
 	}
+
+	segment.VariableValueType = valueType
```
jensneuse: The template is used for concurrent requests. This is a race.
We should never write to a template during resolving.

jensneuse (changes requested):

v2/pkg/engine/resolve/inputtemplate.go
StarpTech: Good catch. I moved the logic to filter.

StarpTech: How can we prevent such situations? I suggest to document those constraints and also add tests in the future that tests the code in real life conditions.

jensneuse: You could add a test that will trigger the race.

jensneuse (approved):

PR #810, opened 2024-05-22
fix: level of null data propagation

Aenimus:


v2/pkg/engine/resolve/resolve_test.go
```diff
@@ -2291,15 +2326,15 @@ func TestResolver_ResolveGraphQLResponse(t *testing.T) {
 					},
 				},
 			},
-		}, Context{ctx: context.Background()}, `{"errors":[{"message":"Failed to fetch from Subgraph at Path 'query', Reason: no data or errors in response."}],"data":null}`
+		}, Context{ctx: context.Background()}, `{"errors":[{"message":"Failed to fetch from Subgraph at Path 'query', Reason: no data or errors in response."},{"message":"Cannot return null for non-nullable field 'Query.nonNullArray'.","path":["nonNullArray"]}],"data":null}`
```
StarpTech, line 2473: How does it behave when a nested object is nullable + when it comes from a different subgraph?

PR #821, opened 2024-06-11
fix leak in planner

phob0s-pl:
We found excessive memory usage when using the library on scale with huge requests. The following fix cleans the planner visitor fields so that they do not grow excessively..


devsergiy (approved):

devsergiy:
added tag [v1.67.3](https://github.com/wundergraph/graphql-go-tools/releases/tag/v1.67.3)

Issue #830, opened 2024-06-23
Could not plan the operation missing path


argoyle:
I have a federated graph using a version of the federation example as my gateway.

Most of my backends extend the Company-type and add fields to it.

Some queries now fails with the below error but not all and I see no pattern to which fail and which succeed. In some cases it's even different for queries which target the same backend that behaves differently.

`internal: bad datasource configuration - could not plan the operation. missing path: [query.company.__typename]`

What might be the problem? What can I do to provide more information to identify the problem?

argoyle:
Reverting back to github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.46 seems to make it behave correctly again.

jensneuse:
Can you publish a test or example that triggers the issue?

argoyle:
Would love to. Not sure really were to begin since I'm not sure what is causing the problem though. Currently looking through the diff between rc.46 and rc.49 to get some kind if hint. :-)

devsergiy:
Hi @argoyle
will it be possible for you to share subgraphs involved in your query and query itself?

argoyle:
Sure thing.

Failing query:
```
curl 'http://localhost:4444/query' \
  -H 'Accept-Language: en,sv;q=0.9,en-US;q=0.8' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'DNT: 1' \
  -H 'Origin: http://localhost:3300' \
  -H 'Pragma: no-cache' \
  -H 'Referer: http://localhost:3300/' \
  -H 'Sec-Fetch-Dest: empty' \
  -H 'Sec-Fetch-Mode: cors' \
  -H 'Sec-Fetch-Site: same-site' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/126.0.0.0 Safari/537.36' \
  -H 'accept: */*' \
  -H 'authorization: Bearer eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI...<redacted>' \
  -H 'content-type: application/json' \
  -H 'sec-ch-ua: "Not/A)Brand";v="8", "Chromium";v="126"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
[...]

PR #837, opened 2024-06-28
chore: add __typename field to trigger bad datasource config

argoyle:
Reproduction-test for #830


argoyle:
Closing since it didn't trigger the correct problem

PR #838, opened 2024-06-28
fix: improve trigger metrics and do not cancel trigger on subscription completion

StarpTech:


v2/pkg/engine/resolve/resolve.go
```diff
@@ -333,27 +334,47 @@ func (r *Resolver) handleEvent(event subscriptionEvent) {
...
+
 	go func() {
 		if wg != nil {
 			wg.Wait()
 		}
 		for _, s := range trig.subscriptions {
 			s.writer.Complete()
 		}
-		if r.reporter != nil {
+		if r.reporter != nil && isInitialized {
```
jensneuse: I don't think you can access a shared value "isInitialized" from a goroutine

jensneuse (changes requested):

v2/pkg/engine/resolve/resolve.go
StarpTech: It is not shared. It is copied before to `isInitialized` the same we do for the subscription count.

jensneuse (approved):

jensneuse (approved):

PR #841, opened 2024-07-01
fix: properly merge representations on the same type but with different nested selections

devsergiy:
…nt nested selections

### DIFF
--- a/v2/pkg/engine/datasource/graphql_datasource/representation_variable.go
+++ b/v2/pkg/engine/datasource/graphql_datasource/representation_variable.go
@@ -58,6 +58,62 @@ func buildRepresentationVariableNode(definition *ast.Document, cfg plan.Federati
 	return visitor.rootObject, nil
 }
 
+func mergeFields(left, right *resolve.Field) *resolve.Field {
+	switch left.Value.NodeKind() {
+	case resolve.NodeKindObject:
+		left.Value = mergeObjects(left.Value, right.Value)
+	case resolve.NodeKindArray:
+		left.Value = mergeArrays(left.Value, right.Value)
+	default:
+	}
+
+	return left
+}
+
+func mergeArrays(left, right resolve.Node) resolve.Node {
+	leftArray, _ := left.(*resolve.Array)
+	rightArray, _ := right.(*resolve.Array)
+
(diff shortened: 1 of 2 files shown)


v2/pkg/engine/datasource/graphql_datasource/representation_variable.go
```diff
@@ -58,38 +58,75 @@ func buildRepresentationVariableNode(definition *ast.Document, cfg plan.Federati
...
+	}
+
+	return leftObject
+}
+
+func isOnTypeEqual(a, b [][]byte) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	for i := range a {
```
jensneuse, line 100: will onType order be stable?

v2/pkg/engine/datasource/graphql_datasource/representation_variable_test.go
```diff
@@ -193,50 +193,255 @@ func TestBuildRepresentationVariableNode(t *testing.T) {
...
+								},
+							},
+							{
+								Name: []byte("addresses"),
+								Value: &resolve.Array{
+									Path: []string{"addresses"},
+									Item: &resolve.Object{
+										Fields: []*resolve.Field{
+											{
+												Name: []byte("zip"),
```
jensneuse, line 409: Can you add a test where zip and street are objects that contain fields to merge? (double nesting)

jensneuse (changes requested):

v2/pkg/engine/datasource/graphql_datasource/representation_variable_test.go
devsergiy: added

v2/pkg/engine/datasource/graphql_datasource/representation_variable.go
devsergiy: yes, we create it [here](https://github.com/wundergraph/graphql-go-tools/blob/633c34481facfa0f2ae432ba07dcbdbd3e45dbc9/v2/pkg/engine/datasource/graphql_datasource/representation_variable.go#L217)

it could be of length 1 or 2
in the case of length 2 - entity interfaces case, the order is always the same - concrete, abstract

PR #843, opened 2024-07-02
fix: JSON block string encoding

### LINKED ISSUE #839: Error in datasource when querying using triple quotes syntax containing quotes or newlines
Given a schema such as:

```graphql
type Query {
    foo(bar: String):String
}
```

I can accept queries like this:

```graphql
query {
  foo(bar: "abc")
}
```

or:

```graphql
query {
  foo(bar: "abc\ndef")
}
```

or with variables (containing any newlines or not):

```graphql
query MyQuery($bar: String){
  foo(bar: $bar)
}
```

But I believe I should be able to also do this, and it fails:

```graphql
query {
  foo(bar: 
"""
abc
def
""")
}
```

I'm using a custom data source with a `JsonVariableRenderer`.

With some trace debugging, I believe the problem is here:

https://github.com/wundergraph/graphql-go-tools/blob/3c4cb175dafb214644c3eee89960808e03924d54/v2/pkg/ast/ast_value.go#L146-L147

Since the string value returned by `d.StringValueContentBytes(value.Ref)` contains a raw newline character, and it's only wrapping it in quotes, the resulting value doesn't have the newline escaped as `"\n"` to be valid JSON.

The failure occurs in my custom datasource at the start of the `Load` function when I try to unmarshal the `input` string.  It should be valid JSON, but in this case it's malformed.

A minimal repro will take some time if you need one, but I thought I'd report it in the meantime.  Thanks.

mattjohnsonpint:
Fixes #839 

Note, I considered using `gjson.AppendJSONString` for better performance, but it currently doesn't have an option to disable escaping of HTML characters - which I think would be undesirable here.  See https://github.com/tidwall/gjson/issues/362

### DIFF
--- a/pkg/ast/ast_val_string_value.go
+++ b/pkg/ast/ast_val_string_value.go
@@ -37,6 +37,101 @@ func (d *Document) StringValueIsBlockString(ref int) bool {
 	return d.StringValues[ref].BlockString
 }
 
+func (d *Document) BlockStringValueContentRawBytes(ref int) []byte {
+
+	// Gets the full block string content, just inside the """ quotes.
+	// This is needed because the lexer ignores whitespace and we need to preserve it
+	// to account for the indentation of the block string.
+
+	blockStart := 0
+	for i := int(d.StringValues[ref].Content.Start) - 1; i >= 0; i-- {
+		if d.Input.RawBytes[i] == '"' {
+			blockStart = i + 1
+			break
+		}
+	}
+
+	blockEnd := d.Input.Length
+	for i := int(d.StringValues[ref].Content.End); i < d.Input.Length; i++ {
+		if d.Input.RawBytes[i] == '"' {
+			blockEnd = i
+			break
+		}
+	}
+
+	return d.Input.RawBytes[blockStart:blockEnd]
+}
+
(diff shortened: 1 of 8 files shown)


mattjohnsonpint:
... actually hold on this for a minute.  I just realized that while it fixes the bug mentioned, it also now creates a new bug since pre-escaped characters will get double-escaped.

I'll add a test for that and update this PR shortly.  Thanks.

mattjohnsonpint:
Ok, it should be good now.  Thanks.

mattjohnsonpint:
LMK if you want it fixed in V1 also.  Thanks.

mattjohnsonpint:
Block indentation has been fixed, with passing tests.

BTW, I tried figuring out how to fix it in the lexer, but ran into problems because the resulting tokens are always just start/end pointers back to the original raw text.  For the indentation parsing algorithm to work correctly, it has to be able to exclude data from the _middle_ of a character sequence.  So instead, I just wrote functions to find the original raw block string, and then parsed it according to the spec.  Hope that works for y'all.

devsergiy (approved):

StarpTech:
@mattjohnsonpint thank you!

PR #847, opened 2024-07-09
fix: ignore __typename when setting select reason

### LINKED ISSUE #830: Could not plan the operation missing path

argoyle:
Fixes #830

### DIFF
--- a/v2/pkg/engine/plan/datasource_filter_visitor.go
+++ b/v2/pkg/engine/plan/datasource_filter_visitor.go
@@ -147,7 +147,9 @@ func (f *DataSourceFilter) applySuggestionHints(hints []NodeSuggestionHint) {
 					f.nodes.items[itemIdx].SelectionReasons = nil
 				}
 			} else {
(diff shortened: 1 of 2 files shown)


argoyle:
Of course. Just wanted to get some hint on if this was reasonable or if it will break something else.

devsergiy:
Hi @argoyle 

Thanks a lot for your contribution
I don't know currently which side effects it could have

Currently, we are doing a rewrite of this functionality, which may fix your issue

I will let you know when we have some results

argoyle:
@devsergiy Should I hold off on creating a test for this then?

argoyle:
Added a couple of tests anyway

PR #874, opened 2024-08-14
fix: correctly render trace and query plan together

jensneuse:


execution/engine/testdata/complex_nesting_query_with_art.json
```diff
@@ -430,7 +430,6 @@
                                 "__typename": "Review",
                                 "attachments": [
                                   {
-                                    "__typename": "Question",
```
devsergiy, line 433: why typenames are removed?

devsergiy: ah, disregard

I see now

jensneuse: it was duplicate coming from subgraph, now deduplicated (to generate valid JSON) by trace deduplication

devsergiy (approved):

PR #876, opened 2024-08-21
feat: expose compose method of engine federation config factory

ikawaha:
First of all, we want to express our respect for the wundergraph team's work on graphql-go-tools. We appreciate the effort that has gone into developing this tool, and we hope that this PR can contribute to improving the performance and memory efficiency of subgraph composition in GraphQL federation.

When building the EngineConfiguration, the BuildEngineConfiguration method performs subgraph composition. This composition is done using the Cosmo library, which generates a serializable data structure known as the router config in the Cosmo library.

As subgraphs grow larger, the time complexity and memory usage of the composition process can become significantly higher. However, if we know when the structure of a subgraph changes, we can reuse the router config to build the EngineConfiguration without recomposing the subgraph.

In this PR, we enable the serialization of the router config (referred to as the intermediate config in graphql-go-tools) used in Cosmo, allowing it to be reused when constructing the EngineConfiguration.

By serializing the router config, we can execute the EngineConfiguration at a lower cost when there are no changes to the subgraph. This is particularly useful during server startup or when refreshing a GraphQL gateway.


execution/engine/config_factory_federation.go
```diff
@@ -150,6 +157,44 @@ func (f *FederationEngineConfigFactory) BuildEngineConfiguration() (conf Configu
...
+	res, err := f.compose()
+	if err != nil {
+		return nil, err
+	}
+	return &RouterConfig{routerConfig: res}, nil
+}
+
+// RouterConfig is an intermediate data that holds the result of the composition
+// and can be saved to a file to configure the router data sources when creating an engine config.
+type RouterConfig struct {
```
devsergiy: Hi @ikawaha 

We probably could use proto types directly
Is there any special need for the wrapping structure?

ikawaha: Thank you for your feedback. I initially wrapped nodev1.RouterConfig because it was hidden, and my goal was to make it serializable. However, I agree that using the proto types directly would be clearer and more straightforward. I've made the necessary adjustments to avoid the wrapping structure and now utilize nodev1.RouterConfig directly.

I've pushed the changes—does this look better to you? [24b4a97](https://github.com/wundergraph/graphql-go-tools/pull/876/commits/24b4a97049e34885563ccfa9977fa727cffefa64)

PR #883, opened 2024-09-10
feat: subgraph error propagation improvements

StarpTech:
This PR adds the following improvements:

- Ensure that no extension fields are propagated by default.
- Use the dataSource name (Subgraph name) for error messages.
- Allow defining granularly which extension fields should be propagated.
- Allow defining a fallback value for the extension `code` field.
- Allow attaching the dataSource name as the `serviceName` field to the extension object.
- `data: null` is never returned for errors that are happening before query execution. This compliant with the [spec](https://spec.graphql.org/draft/#sec-Data).


df-wg (approved):
lgtm

v2/pkg/engine/resolve/errors.go
```diff
@@ -45,10 +45,18 @@ func (e *SubgraphError) Error() string {
 
 	var bf bytes.Buffer
 
-	if e.SubgraphName == "" {
+	var subgraphName string
+
+	if e.DataSourceInfo.Name != "" {
+		subgraphName = e.DataSourceInfo.Name
+	} else {
+		subgraphName = e.DataSourceInfo.ID
```
jensneuse: This is the only case where we access the ID, which is a fallback if name is not available. Is it possible that we really only need either ID or Name? So it could jus tbe ID or Name, like before, and we populate it either with ID or Name if available?

I've checked and we never really use ID except here as a fallback.

v2/pkg/engine/resolve/loader.go
```diff
@@ -649,34 +660,128 @@ func (l *Loader) mergeErrors(res *result, fetchItem *FetchItem, value *astjson.V
...
+
+			hasCode := false
+			extensions.GetObject().Visit(func(k []byte, v *astjson.Value) {
+				if unsafebytes.BytesToString(k) == "code" {
+					hasCode = true
+				}
+			})
+
+			if !hasCode {
+				extensions.Set("code", astjson.MustParse(`"`+defaultErrorExtensionCode+`"`))
```
jensneuse: I was thinking what is the appropriate and efficient way of parsing a string.
As we own astjson as our own package, how about adding a func to it?
```go
extensions.Set("code", astjson.MustParseString(defaultErrorExtensionCode))
```

safe alternative:

```go
v,err := astjson.ParseString(defaultErrorExtensionCode)
if err != nil {
	return err
}
extensions.Set("code", v)
```

It seems unnecessary to manually concat 3 strings only to parse the string as a string between 2 quotes.

v2/pkg/engine/resolve/loader.go
```diff
@@ -649,34 +660,128 @@ func (l *Loader) mergeErrors(res *result, fetchItem *FetchItem, value *astjson.V
...
+	}
+	for _, value := range values {
+		if value.Exists("extensions") {
+			extensions := value.Get("extensions")
+			if extensions.Type() != astjson.TypeObject {
+				continue
+			}
+			extensions.Set("serviceName", astjson.MustParse(`"`+serviceName+`"`))
+		} else {
+			value.Set("extensions", astjson.MustParse(`{"serviceName":"`+serviceName+`"}`))
```
jensneuse: You can use this approach to create an object with serviceName:
https://github.com/wundergraph/astjson/blob/bb15f94bd362179ec50bad54dd16b59807f99fbb/util_test.go#L98

v2/pkg/engine/resolve/loader.go
```diff
@@ -738,18 +843,26 @@ func (l *Loader) optionallyRewriteErrorPaths(fetchItem *FetchItem, values []*ast
...
+
+	for _, value := range values {
+		if value.Exists("extensions") {
+			extensions := value.Get("extensions")
+			if extensions.Type() != astjson.TypeObject {
+				continue
+			}
+			extensions.Set("statusCode", astjson.MustParse(strconv.Itoa(statusCode)))
+		} else {
+			value.Set("extensions", astjson.MustParse(`{"statusCode":`+strconv.Itoa(statusCode)+`}`))
```
jensneuse, line 868: same as above

v2/pkg/engine/resolve/loader.go
```diff
@@ -649,34 +660,128 @@ func (l *Loader) mergeErrors(res *result, fetchItem *FetchItem, value *astjson.V
...
+
+	// Allow to delete extensions entirely
+	l.optionallyOmitErrorExtensions(v)
+
 	astjson.AppendToArray(l.resolvable.errors, errorObject)
+
 	return nil
 }
 
+func (l *Loader) optionallyAllowCustomExtensionProperties(values []*astjson.Value, allowedProperties map[string]struct{}) {
```
jensneuse: please add a description on the purpose and how it's being used e2e

v2/pkg/engine/resolve/loader.go
```diff
@@ -649,34 +660,128 @@ func (l *Loader) mergeErrors(res *result, fetchItem *FetchItem, value *astjson.V
...
+			// If there are no more properties, we remove the extensions object
+			if len(allowedProperties) == 0 || extObj.Len() == 0 {
+				value.Del("extensions")
+				continue
+			}
+		}
+	}
+}
+
+func (l *Loader) optionallyEnsureExtensionErrorCode(values []*astjson.Value, defaultErrorExtensionCode string) {
```
jensneuse: please add a description on the purpose and how it's being used e2e

v2/pkg/engine/resolve/loader.go
```diff
@@ -649,34 +660,128 @@ func (l *Loader) mergeErrors(res *result, fetchItem *FetchItem, value *astjson.V
...
+			if !hasCode {
+				extensions.Set("code", astjson.MustParse(`"`+defaultErrorExtensionCode+`"`))
+			}
+		} else {
+			value.Set("extensions", astjson.MustParse(`{"code":"`+defaultErrorExtensionCode+`"}`))
+		}
+	}
+}
+
+func (l *Loader) optionallyAttachServiceNameToErrorExtension(values []*astjson.Value, serviceName string) {
```
jensneuse, line 766: please add a description on the purpose and how it's being used e2e

v2/pkg/engine/resolve/loader.go
```diff
@@ -649,34 +660,128 @@ func (l *Loader) mergeErrors(res *result, fetchItem *FetchItem, value *astjson.V
...
 		}
 
 		l.ctx.appendSubgraphError(goerrors.Join(res.err, subgraphError))
+
 	}
 
-	l.optionallyOmitErrorExtensions(values)
 	l.optionallyOmitErrorLocations(values)
 	l.optionallyRewriteErrorPaths(fetchItem, values)
+	l.optionallyAllowCustomExtensionProperties(values, l.allowedErrorExtensionFields)
```
jensneuse: is it necessary to pass these arguments?
They are accessible to the loader funcs anyways

v2/pkg/engine/resolve/loader.go
```diff
@@ -649,34 +660,128 @@ func (l *Loader) mergeErrors(res *result, fetchItem *FetchItem, value *astjson.V
...
+
+	for _, value := range values {
+		if value.Exists("extensions") {
+			extensions := value.Get("extensions")
+			if extensions.Type() != astjson.TypeObject {
+				continue
+			}
+
+			hasCode := false
+			extensions.GetObject().Visit(func(k []byte, v *astjson.Value) {
```
jensneuse: why not `extension.Get("code") != nil`?

v2/pkg/engine/resolve/loader.go
```diff
@@ -649,34 +660,128 @@ func (l *Loader) mergeErrors(res *result, fetchItem *FetchItem, value *astjson.V
...
+				if unsafebytes.BytesToString(k) == "code" {
+					hasCode = true
+				}
+			})
+
+			if !hasCode {
+				extensions.Set("code", astjson.MustParse(`"`+defaultErrorExtensionCode+`"`))
+			}
+		} else {
+			value.Set("extensions", astjson.MustParse(`{"code":"`+defaultErrorExtensionCode+`"}`))
```
jensneuse: MustParseString?

v2/pkg/engine/resolve/loader.go
```diff
@@ -649,34 +660,128 @@ func (l *Loader) mergeErrors(res *result, fetchItem *FetchItem, value *astjson.V
...
+	if !l.attachServiceNameToErrorExtension {
+		return
+	}
+	for _, value := range values {
+		if value.Exists("extensions") {
+			extensions := value.Get("extensions")
+			if extensions.Type() != astjson.TypeObject {
+				continue
+			}
+			extensions.Set("serviceName", astjson.MustParse(`"`+serviceName+`"`))
```
jensneuse: MustParseString?

v2/pkg/engine/resolve/loader.go
```diff
@@ -738,18 +843,26 @@ func (l *Loader) optionallyRewriteErrorPaths(fetchItem *FetchItem, values []*ast
...
-	if err != nil {
-		return
+
+	for _, value := range values {
+		if value.Exists("extensions") {
+			extensions := value.Get("extensions")
+			if extensions.Type() != astjson.TypeObject {
+				continue
+			}
+			extensions.Set("statusCode", astjson.MustParse(strconv.Itoa(statusCode)))
```
jensneuse, line 866: We could add `astjson.MustParseInt`

jensneuse (changes requested):
Overall it looks very good and is a significant improvement! I'd like to see if we can move some "workarounds" to the astjson package.

v2/pkg/engine/resolve/errors.go
StarpTech: Yes, we can assume name is always present.

StarpTech:
@jensneuse astjson has already a concept of Arenas which allows for easy value creation including memory reuse. Please check again.

jensneuse:
> @jensneuse astjson has already a concept of Arenas which allows for easy value creation including memory reuse. Please check again.

The idea is good. Great find!

I'd slightly change the implementation though.
Instead of having an arena pool, you can add one arena to each Resolvable.
The Resolvable has the lifecycle of the Request, which means that it can be invalidated once the Request handling is done.
This is already implemented and we're using it to free resources live throughout the lifecycle of a Request.
So I think the simple optimal solution would be to add one arena to each resolvable and create JSON AST nodes from it.

WDYT?

jensneuse (approved):

PR #891, opened 2024-09-19
feat: add __typename validation in resolvable

jensneuse:


v2/pkg/engine/resolve/resolve_test.go
```diff
@@ -1876,6 +1876,64 @@ func TestResolver_ResolveGraphQLResponse(t *testing.T) {
...
+										Nullable:   false,
+										IsTypeName: true,
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		}, Context{ctx: context.Background()}, `{"errors":[{"message":"Subgraph 'Users' returned invalid value 'NotUser' for __typename field.","path":["__typename"]}],"data":null}`
```
StarpTech: How does it behave with different error formats? Does it respect the propagation settings? By default it should be wrapped but this looks like passthrough. It looks to me that we handle errors differently here in the engine.

StarpTech (changes requested):

v2/pkg/engine/resolve/resolve_test.go
```diff
@@ -1876,6 +1876,64 @@ func TestResolver_ResolveGraphQLResponse(t *testing.T) {
 			},
 		}, Context{ctx: context.Background()}, `{"data":{"user":{"id":1,"name":"Jannik","__typename":"User","aliased":"User","rewritten":"User"}}}`
 	}))
+	t.Run("__typename checks", testFn(func(t *testing.T, ctrl *gomock.Controller) (node *GraphQLResponse, ctx Context, expectedOutput string) {
```
df-wg, line 1879: This seems like a pretty significant change, maybe some more tests would help validate it (e.g. one that a `User` type works successfully, another test which isn't being rewritten, one where you enable multiple typenames, etc)

df-wg (changes requested):

v2/pkg/engine/resolve/resolve_test.go
jensneuse: All other tests validate the "correct" behaviour already.

v2/pkg/engine/resolve/resolve_test.go
jensneuse: This is not an error originating from a Subgraph. A subgraph is misbehaving and returning invalid data for a field. The engine (Router) catches it and ensures that the client doesn't rely on incorrect information by bubbling up an error, setting the field to nil, and propagating an extensions code.

StarpTech: My question: What is responded to the client? If we propagate this error in all subgraph propagation modes, we have inconsistencies and this will break clients.

StarpTech (approved):
LGTM

df-wg (approved):
lgtm

PR #895, opened 2024-09-23
fix readme documentation for Validator

nogates:
is the [other way around](https://github.com/wundergraph/graphql-go-tools/blob/master/v2/pkg/astvalidation/operation_validation.go#L63), right?


devsergiy:
@nogates You are right
Thanks for the contribution!

PR #901, opened 2024-09-27
fix: propogate typenames when inteface field is used

JivusAyrus:
When the field of an interface is used, currently it propagates only the name of the interface in ParentTypeNames.
This PR adds all the types which implement the interface also.


jensneuse (changes requested):
Please add a test.

v2/pkg/engine/plan/visitor.go
```diff
@@ -383,6 +383,16 @@ func (v *Visitor) resolveFieldInfo(ref, typeRef int, onTypeNames [][]byte) *reso
 		parentTypeNames = append(parentTypeNames, onTypeName)
 	}
 
+	if v.Walker.EnclosingTypeDefinition.Kind == ast.NodeKindInterfaceTypeDefinition {
```
jensneuse, line 375: What if the enclosing type is a union?

JivusAyrus: Havent tested, will test

JivusAyrus: DIscussed with david, will create a new ticket and handle it

jensneuse (approved):

PR #914, opened 2024-10-07
feat: allow list for subgraph error fields

thisisnithin:


v2/pkg/engine/resolve/loader.go
```diff
@@ -816,18 +816,26 @@ func (l *Loader) optionallyOmitErrorExtensions(values []*astjson.Value) {
...
 }
 
-// optionallyOmitCustomErrorFields removes all custom fields from the subgraph error
-func (l *Loader) optionallyOmitCustomErrorFields(values []*astjson.Value) {
-	if !l.omitCustomSubgraphErrorFields {
-		return
+// optionallyOmitErrorFields removes all fields from the subgraph error which are not whitelisted. We do not remove message.
+func (l *Loader) optionallyOmitErrorFields(values []*astjson.Value) {
+	// always allow message
+	l.allowedSubgraphErrorFields["message"] = struct{}{}
```
StarpTech: It is better to construct this data once in resolve.go. The same way we do it with allowedExtensionFields

StarpTech (approved):
LGTM

PR #916, opened 2024-10-08
feat: Add Heartbeat to updater

df-wg:
This will be used in https://github.com/wundergraph/cosmo/pull/1227, specifically so that all of the subscriptions can create heartbeats to keep connections alive. This is a necessary step for supporting https://www.apollographql.com/docs/graphos/routing/operations/subscriptions/multipart-protocol#executing-a-subscription

### DIFF
--- a/v2/pkg/engine/datasource/graphql_datasource/graphql_sse_handler.go
+++ b/v2/pkg/engine/datasource/graphql_datasource/graphql_sse_handler.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"math"
 	"net/http"
+	"time"
 
 	"github.com/buger/jsonparser"
 	log "github.com/jensneuse/abstractlogger"
@@ -52,11 +53,24 @@ func (h *gqlSSEConnectionHandler) StartBlocking(sub Subscription) {
 
 	go h.subscribe(reqCtx, sub, dataCh, errCh)
 
+	var ticker *time.Ticker
+	if sub.options.SendHeartbeat {
+		ticker = time.NewTicker(resolve.HearbeatInterval)
+	}
+
 	for {
 		select {
+		case <-tickerC(ticker):
+			sub.updater.Heartbeat()
 		case data := <-dataCh:
+			if ticker != nil {
+				ticker.Reset(resolve.HearbeatInterval)
+			}
 			sub.updater.Update(data)
 		case data := <-errCh:
+			if ticker != nil {
+				ticker.Reset(resolve.HearbeatInterval)
+			}
 			sub.updater.Update(data)
 			return
 		case <-reqCtx.Done():
--- a/v2/pkg/engine/datasource/graphql_datasource/graphql_sse_handler.go
+++ b/v2/pkg/engine/datasource/graphql_datasource/graphql_sse_handler.go
@@ -56,6 +56,7 @@ func (h *gqlSSEConnectionHandler) StartBlocking(sub Subscription) {
 	var ticker *time.Ticker
 	if sub.options.SendHeartbeat {
 		ticker = time.NewTicker(resolve.HearbeatInterval)
+		defer ticker.Stop()
 	}
 
 	for {
--- a/v2/pkg/engine/datasource/graphql_datasource/graphql_sse_handler.go
+++ b/v2/pkg/engine/datasource/graphql_datasource/graphql_sse_handler.go
@@ -53,25 +53,18 @@ func (h *gqlSSEConnectionHandler) StartBlocking(sub Subscription) {
 
 	go h.subscribe(reqCtx, sub, dataCh, errCh)
 
-	var ticker *time.Ticker
-	if sub.options.SendHeartbeat {
-		ticker = time.NewTicker(resolve.HearbeatInterval)
-		defer ticker.Stop()
-	}
+	ticker := time.NewTicker(resolve.HearbeatInterval)
+	defer ticker.Stop()
 
 	for {
 		select {
-		case <-tickerC(ticker):
+		case <-ticker.C:
 			sub.updater.Heartbeat()
 		case data := <-dataCh:
-			if ticker != nil {
-				ticker.Reset(resolve.HearbeatInterval)
-			}
+			ticker.Reset(resolve.HearbeatInterval)
 			sub.updater.Update(data)
 		case data := <-errCh:
-			if ticker != nil {
-				ticker.Reset(resolve.HearbeatInterval)
-			}
+			ticker.Reset(resolve.HearbeatInterval)
 			sub.updater.Update(data)
 			return
 		case <-reqCtx.Done():
--- a/v2/pkg/engine/datasource/graphql_datasource/graphql_sse_handler_test.go
+++ b/v2/pkg/engine/datasource/graphql_datasource/graphql_sse_handler_test.go
@@ -81,6 +81,73 @@ func TestGraphQLSubscriptionClientSubscribe_SSE(t *testing.T) {
 	serverCancel()
 }
 
+func TestGraphQLSubscriptionClientSubscribe_Heartbeat(t *testing.T) {
+	if flags.IsWindows {
+		t.Skip("skipping test on windows")
+	}
+
+	serverDone := make(chan struct{})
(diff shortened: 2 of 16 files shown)


v2/pkg/engine/resolve/resolve.go
```diff
@@ -412,11 +420,58 @@ func (r *Resolver) handleEvent(event subscriptionEvent) {
...
+	trig, ok := r.triggers[id]
+	if !ok {
+		return
+	}
+	if r.options.Debug {
+		fmt.Printf("resolver:heartbeat:%d\n", id)
+	}
+	wg := &sync.WaitGroup{}
+	trig.inFlight = wg
+	for c, s := range trig.subscriptions {
```
df-wg, line 439: Ideally, I'd love to only send the heartbeat for one particular subscription on a trigger, but I couldn't find an example where we send a subscription ID along.

v2/pkg/engine/resolve/resolve.go
```diff
@@ -412,11 +420,58 @@ func (r *Resolver) handleEvent(event subscriptionEvent) {
...
+			if err != nil {
+				// client disconnected
+				_ = r.AsyncUnsubscribeSubscription(s.id)
+				return
+			}
+			if r.options.Debug {
+				fmt.Printf("resolver:heartbeat:subscription:flushed:%d\n", s.id.SubscriptionID)
+			}
+			if r.reporter != nil {
+				r.reporter.SubscriptionUpdateSent()
```
df-wg, line 470: Should I define `SubscriptionHeartbeatSent`? Would that be a valuable metric?

v2/pkg/engine/datasource/graphql_datasource/graphql_sse_handler.go
```diff
@@ -52,11 +53,24 @@ func (h *gqlSSEConnectionHandler) StartBlocking(sub Subscription) {
...
+	if sub.options.SendHeartbeat {
+		ticker = time.NewTicker(resolve.HearbeatInterval)
+	}
+
 	for {
 		select {
+		case <-tickerC(ticker):
+			sub.updater.Heartbeat()
 		case data := <-dataCh:
+			if ticker != nil {
```
StarpTech: It would be great to document the current design. This allows to avoid sending heartbeats when the subgraph has already send data.

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -112,6 +112,9 @@ func NewGraphQLSubscriptionClient(httpClient, streamingClient *http.Client, engi
 // If connection protocol is SSE, a new connection is always created
 // If no connection exists, the client initiates a new one
 func (c *subscriptionClient) Subscribe(reqCtx *resolve.Context, options GraphQLSubscriptionOptions, updater resolve.SubscriptionUpdater) error {
+	if reqCtx.ExecutionOptions.SendHeartbeat {
```
StarpTech: I'm not sure if this is the right approach. If the first subscription is not multipart while the second one is deduplicated, it could result in no heartbeat being sent. You would need to run the heartbeat continuously or enable the heartbeat before the duplication check.

v2/pkg/engine/datasource/graphql_datasource/graphql_ws_handler.go
```diff
@@ -56,6 +57,11 @@ func (h *gqlWSConnectionHandler) StartBlocking(sub Subscription) {
 
 	go h.readBlocking(readCtx, dataCh, errCh)
 
+	ticker := &time.Ticker{}
+	if sub.options.SendHeartbeat {
+		ticker = time.NewTicker(resolve.HearbeatInterval)
```
StarpTech: You should stop the ticker after the function returned.

v2/pkg/engine/resolve/resolve.go
```diff
@@ -412,11 +420,58 @@ func (r *Resolver) handleEvent(event subscriptionEvent) {
...
+func (r *Resolver) handleHeartbeat(id uint64, data []byte) {
+	trig, ok := r.triggers[id]
+	if !ok {
+		return
+	}
+	if r.options.Debug {
+		fmt.Printf("resolver:heartbeat:%d\n", id)
+	}
+	wg := &sync.WaitGroup{}
+	trig.inFlight = wg
```
StarpTech: This is not needed and would collide with the logic in `handleTriggerUpdate`. The WaitGroup is only used to graceful shutdown a trigger. Heartbeats can be ignored.

v2/pkg/engine/resolve/resolve.go
StarpTech: How would this work when we trigger a heartbeat from the handler that distributes to all subscriptions? If this becomes ever a bottleneck, we could create indexes.

v2/pkg/engine/resolve/resolve.go
StarpTech: I'd say no, at least I couldn't make use of it right now.

StarpTech (changes requested):

v2/pkg/engine/datasource/graphql_datasource/graphql_sse_handler.go
df-wg: I added an additional comment in the Heartbeat godoc. Is that what you had intended?

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
df-wg: What else do you think makes sense to do? From what I could see in the `handleAddSubscription`, after ensuring that the triggerID is present, it will add a subscription to the map and then return, without calling `add.resolve.Trigger.Source.Start`. Should it instead always send the heartbeat, and then in `handleHeartbeat`, when looping over all of the `trig.subscriptions`, it would only send it to the ones who have `sub.sendHeartbeat` as `true`? 

My only concern is performance. That change would mean that every subscription will be sending a `heartbeat`, and then on the resolver side it's only sent if the subscription has opted in. That could conceivably mean significantly more events for the event handler to handle that users hadn't opted into. Do you think it's worth the potential performance issue, am I just exaggerating it?

StarpTech: > What else do you think makes sense to do? From what I could see in the handleAddSubscription, after ensuring that the triggerID is present, it will add a subscription to the map and then return, without calling add.resolve.Trigger.Source.Start. Should it instead always send the heartbeat, and then in handleHeartbeat, when looping over all of the trig.subscriptions, it would only send it to the ones who have sub.sendHeartbeat as true?

For me, it sounds fine, especially when considering that heartbeats are only sent when no data has been recently sent on the trigger subscriptions.

But, I'd make the heartbeat configurable. This is also speed up tests.

v2/pkg/engine/resolve/resolve.go
```diff
@@ -19,6 +19,14 @@ import (
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/pool"
 )
 
+const (
+	HearbeatInterval = 5 * time.Second
+)
+
+var (
+	heartbeat = []byte("{}")
```
StarpTech: Currently, we have only one but this is protocol specific. I'd give it a proper name prefix.

df-wg: Makese sense. Do you think `Heartbeat()` should take a data parameter, like `Update`?

StarpTech: The callee of `hearbeat` don't know about the type of the subscription. I'd ensure everything is well commented, add a prefix and call it done 😃

v2/pkg/engine/datasource/graphql_datasource/graphql_sse_handler_test.go
```diff
@@ -81,6 +81,73 @@ func TestGraphQLSubscriptionClientSubscribe_SSE(t *testing.T) {
 	serverCancel()
 }
 
+func TestGraphQLSubscriptionClientSubscribe_Heartbeat(t *testing.T) {
+	if flags.IsWindows {
```
StarpTech, line 85: Why skipped on windows?

StarpTech (approved):
LGTM

PR #917, opened 2024-10-09
fix: Refresh the planner before reusing it

ikawaha:
Hi team,

I have concerns that reusing the Planner in the execution engine of wundergraph/graphql-go-tools may be causing issues. Specifically, the execution engine retains the Planner and uses it to compute plans when executing GraphQL queries. However, during plan computation, the Planner holds onto values in its internal state. As a result, when computing the next plan, the internal state of the Planner may be in an unexpected state, potentially leading to incorrect plan values.

To resolve this issue, I propose a fix that clears the internal state of the Planner before reusing it, effectively resetting it to its initial state. This ensures that each plan computation starts from a clean state and is not affected by residual data from previous computations.

We have been experiencing a bug where, occasionally, executing GraphQL queries results in incorrect responses. It appears that this is caused by incorrect plan calculations due to the Planner's residual internal state. With this fix, the issue we've been facing seems to be improved.

In cosmo, for example, the planner seems to be generated each time.

Best regards,

### DIFF
--- a/execution/engine/execution_engine.go
+++ b/execution/engine/execution_engine.go
@@ -247,6 +247,8 @@ func (e *ExecutionEngine) getCachedPlan(ctx *internalExecutionContext, operation
 
 	e.plannerMu.Lock()
 	defer e.plannerMu.Unlock()
+
+	e.planner.Refresh()
 	planResult := e.planner.Plan(operation, definition, operationName, report)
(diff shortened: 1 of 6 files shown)


ikawaha:
I am closing this PR and will handle the issue in https://github.com/wundergraph/graphql-go-tools/pull/922 .

PR #921, opened 2024-10-14
feat: add epoll support to ws origin requests

jensneuse:


v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -245,29 +338,89 @@ func (c *subscriptionClient) requestHash(ctx *resolve.Context, options GraphQLSu
...
+			return nil, &UpgradeRequestError{
+				URL:        subgraphHttpURL,
+				StatusCode: upgradeResponse.StatusCode,
+			}
+		}
 		return nil, err
 	}
-	// Disable the maximum message size limit. Don't use MaxInt64 since
-	// the nhooyr.io/websocket doesn't handle it correctly on 32 bit systems.
 	conn.SetReadLimit(math.MaxInt32)
```
StarpTech: Is this still valid?

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -92,7 +148,9 @@ func NewGraphQLSubscriptionClient(httpClient, streamingClient *http.Client, engi
 	for _, option := range options {
 		option(op)
 	}
-	return &subscriptionClient{
+	// ignore error is ok, it means that epoll is not supported, which is handled gracefully by the client
```
StarpTech: How is it graceful handled? Will it work on Windows?

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -286,8 +439,10 @@ func (c *subscriptionClient) newWSConnectionHandler(reqCtx context.Context, opti
 		}
 	}
 
+	netConn := conn.NetConn()
```
StarpTech: Just to double check. Did you know that the old library also has a https://pkg.go.dev/github.com/coder/websocket#NetConn utility? Wasn't this the reason why you choose gorilla?

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -387,3 +537,88 @@ func waitForAck(ctx context.Context, conn *websocket.Conn) error {
...
+			}
+			c.activeConnectionsMu.Unlock()
+			if active {
+				continue
+			}
+			go c.handleConnection(id, handler, conn)
+		}
+		c.connectionsMu.Unlock()
+
+		if len(connections) == 50 {
```
StarpTech: Why is it necessary to check for the connection count? Is this an performance improvement to do this on the same iteration?

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -387,3 +537,88 @@ func waitForAck(ctx context.Context, conn *websocket.Conn) error {
...
+		select {
+		case <-done:
+			return
+		case <-tick.C:
+			continue
+		}
+	}
+}
+
+func (c *subscriptionClient) handleConnection(id int, handler ConnectionHandler, conn net.Conn) {
```
StarpTech: We should verify if sockets are properly cleaned up when subgraphs disconnects or when connections are dead.

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler_test.go
```diff
@@ -240,96 +238,6 @@ func TestWebsocketSubscriptionClientError_GQLTWS(t *testing.T) {
 	}, time.Second, time.Millisecond*10, "server did not close")
 }
 
-func TestWebSocketSubscriptionClientInitIncludePing_GQLTWS(t *testing.T) {
-	if flags.IsWindows {
```
StarpTech, line 244: Our previous implementation was Windows compatible. That should be our goal with this implementation too.

v2/pkg/internal/epoller/README.md
```diff
@@ -0,0 +1,8 @@
+# Epoller
```
StarpTech: If we copy paste the epoll implementation to the engine, we should import it from the router to avoid maintaining the same implementation twice. I commented it on the cosmo PR.

StarpTech (changes requested):

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -387,3 +575,99 @@ func waitForAck(ctx context.Context, conn *websocket.Conn) error {
 		}
 	}
 }
+
+func (c *subscriptionClient) runEpoll(ctx context.Context) {
+	var (
+		ticker <-chan time.Time
```
StarpTech: remove ticker

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go
```diff
@@ -1,88 +1,166 @@
...
+	updater                       resolve.SubscriptionUpdater
+}
+
+func (h *gqlTWSConnectionHandler) ServerClose() {
+	h.updater.Done()
+	_ = h.conn.Close()
+}
+
+func (h *gqlTWSConnectionHandler) ClientClose() {
+	h.updater.Done()
```
StarpTech, line 37: I think it would be better if we would use the websocket library to handle such cases. We basically, replicate the library. You can call `websocketConn.Close()`.

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go
```diff
@@ -1,88 +1,166 @@
...
+		engineContext:  engineContext,
+		log:            l,
+		updater:        updater,
+		options:        options,
 	}
 }
 
-func (h *gqlTWSConnectionHandler) StartBlocking(sub Subscription) {
-	readCtx, cancel := context.WithCancel(h.ctx)
+func (h *gqlTWSConnectionHandler) StartBlocking() error {
```
StarpTech, line 102: Why do we still need this logic? Is this for Windows support?

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go
```diff
@@ -93,18 +171,21 @@ func (h *gqlTWSConnectionHandler) StartBlocking(sub Subscription) {
 			switch messageType {
 			case messageTypePing:
```
StarpTech, line 136: This logic is duplicated in read message. Can we consolidate it?

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go
```diff
@@ -1,88 +1,166 @@
...
+		h.log.Error("failed to write complete message", log.Error(err))
+	}
+	_ = h.conn.Close()
+}
+
+func (h *gqlTWSConnectionHandler) Subscribe() error {
+	return h.subscribe()
+}
+
+func (h *gqlTWSConnectionHandler) ReadMessage() (done, timeout bool) {
```
StarpTech: Heartbeat handling for subscriptions over multipart is currently not handled when using epoll. We also need to ensure that this logic is part of epoll to avoid excessive use of goroutines.

See https://github.com/wundergraph/graphql-go-tools/blob/7b40eb8fc88127ce992b7fb6b71a48b963546f65/v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go#L149

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
StarpTech: Remove as discussed

StarpTech (changes requested):

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -387,3 +530,81 @@ func waitForAck(ctx context.Context, conn *websocket.Conn) error {
...
+			c.activeConnectionsMu.Lock()
+			_, active := c.activeConnections[id]
+			if !active {
+				c.activeConnections[id] = struct{}{}
+			}
+			c.activeConnectionsMu.Unlock()
+			if active {
+				continue
+			}
+			go c.handleConnection(id, handler, conn)
```
StarpTech: As discussed. We should try to limit the concurrency here without service interruption.

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
jensneuse: yes

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
jensneuse: removed

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler_test.go
jensneuse: It's compatible now

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
jensneuse: done

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go
jensneuse: correct, Windows doesn't use epoll, so it will use the blocking approach with a goroutine.

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go
jensneuse: no :D

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go
jensneuse: This properly cleans up the connection when a client to the router closes the connection or ends the subscription, so it's necessary like this.

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go
jensneuse: Heartbeat doesn't even belong into the origin client implementation. It's a concern of the resolver, which is router client facing. I'll give a longer explanation on slack, but the short one is that the heartbeat implementation doesn't belong into the origin client, because all it does it call updater.Heartbeat, which the resolver can do by itself, which means that we don't even need to expose heartbeat to the updater interface.

StarpTech (approved):
LGTM

PR #925, opened 2024-10-17
fix: execution validation order, do not reuse planner

devsergiy:


devsergiy:
@mattjohnsonpint FYI - variables validation should go after variables extraction - as variables coercion happening along with the extraction

PR #929, opened 2024-10-20
feat: include subgraph name in ART

StarpTech:


execution/engine/testdata/complex_nesting_query_with_art.json
```diff
@@ -91,6 +91,7 @@
               "kind": "Single",
               "path": "",
               "source_id": "0",
+              "source_name": "0",
```
StarpTech, line 94: @Aenimus The composition version that is currently used by the engine does not use the subgraph name for the subgraph name in the generated router config. I created a ticket for this.

StarpTech: We have two different implementations of how the plan is generated. In the test we use `NewDataSourceConfiguration` in the router `NewDataSourceConfigurationWithName`. Therefore the ID is used for name and id.

thisisnithin (approved):

PR #930, opened 2024-10-20
chore: update composition-go version

Aenimus:


Aenimus:
`type Query` changed places because extensions are dealt with differently in composition. It used to be that they were dealt with at the end (so occurred at the end of the schema).
CC: @StarpTech @devsergiy

jensneuse (approved):

PR #934, opened 2024-10-24
feat: add an option to omit variables content in the variables validator error messages

Sam-tesouro:
per discussion at https://github.com/wundergraph/cosmo/issues/1182


v2/pkg/variablesvalidation/variablesvalidation.go
```diff
@@ -87,6 +89,11 @@ const (
 	pathItemKindArray
 )
 
+func devModeCheck() bool {
+	d := strings.ToLower(os.Getenv("DEV_MODE"))
```
jensneuse: Can we turn this into a property on the variablesVisitor that we can pass on creation?
This way, it won't depend on env vars, etc. which makes it more universally usable.

jensneuse (changes requested):

v2/pkg/variablesvalidation/variablesvalidation.go
Tesouro-Chris: Hey @jensneuse , thanks for the suggested change. @Sam-tesouro and I were just chatting, and I'm afraid my go foo is a bit rusty so wondering how we'd pass that into the variables visitor. If you have a quick example, suggestion, or link, that'd help. We're working on it either way though. Thanks for all you do!

Sam-tesouro: Implemented a thing that might be what you want, possibly.

Sam-tesouro: I see the issue with being so coupled to the DEV_MODE env var, working on decoupling still. Biggest hangup currently is just figuring out how I am going to achieve test coverage.

jensneuse: I can help you, need to see when I find the time.

Sam-tesouro: Thank you! I believe I achieved decoupling. Glad to make further modifications if you have preferences!

devsergiy (approved):

Sam-tesouro:
@devsergiy @StarpTech @jensneuse many thanks for taking this over the finish line, I was out on paternity and y'all are amazing.

PR #936, opened 2024-10-25
feat: validate returned enum values

Aenimus:


v2/pkg/ast/ast_enum_type_definition.go
```diff
@@ -56,6 +56,9 @@ func (d *Document) EnumTypeDefinitionHasEnumValueDefinition(ref int) bool {
 func (d *Document) EnumTypeDefinitionContainsEnumValue(enumTypeDef int, valueName ByteSlice) bool {
 	for _, i := range d.EnumTypeDefinitions[enumTypeDef].EnumValuesDefinition.Refs {
 		if bytes.Equal(valueName, d.EnumValueDefinitionNameBytes(i)) {
+			if _, isInaccessible := d.EnumValueDefinitionDirectiveByName(i, []byte("inaccessible")); isInaccessible {
```
jensneuse: Would it make sense to reflect "inaccessible" in the AST directly?
You could then have 2 checks:
1. does the enum definition contain the value?
2. is the value accessible?
Currently, you return false for 1 when it's inaccessible when in reality it exists, but is not accessible, which, depending on the use case might not what the function caller expects.

v2/pkg/engine/plan/visitor.go
```diff
@@ -707,10 +707,22 @@ func (v *Visitor) resolveFieldValue(fieldRef, typeRef int, nullable bool, path [
...
-				Path:                 path,
-				Nullable:             nullable,
-				UnescapeResponseJson: unescapeResponseJson,
+			values := make([]string, 0, len(v.Definition.EnumTypeDefinitions[typeDefinitionNode.Ref].EnumValuesDefinition.Refs))
+			inaccessibleValues := make([]string, 0)
+			for _, valueRef := range v.Definition.EnumTypeDefinitions[typeDefinitionNode.Ref].EnumValuesDefinition.Refs {
+				valueName := v.Definition.EnumValueDefinitionNameString(valueRef)
+				values = append(values, valueName)
+
+				if _, isInaccessible := v.Definition.EnumValueDefinitionDirectiveByName(valueRef, []byte("inaccessible")); isInaccessible {
```
jensneuse, line 716: I think it would be better to add inaccessible as a feature to our AST and not have it as implicit property of a graph

jensneuse (approved):

v2/pkg/engine/plan/visitor.go
devsergiy: It is not ast feature, it is schema behavior 

didn't like the idea to have it directly in the ast

v2/pkg/ast/ast_enum_type_definition.go
devsergiy: +1 this method should return every value which exists

devsergiy (changes requested):
please undo ast modifications

v2/pkg/engine/resolve/node_enum.go
```diff
@@ -0,0 +1,67 @@
...
+
+	if !slices.Equal(e.Path, other.Path) {
+		return false
+	}
+
+	return true
+}
+
+func (e *Enum) isValidValue(returnedValue string) bool {
+	for _, value := range e.Values {
```
devsergiy: maybe we could use `slices.Contains`?

v2/pkg/engine/resolve/node_enum.go
```diff
@@ -0,0 +1,67 @@
...
+		if value == returnedValue {
+			return true
+		}
+	}
+	return false
+}
+
+func (e *Enum) isAccessibleValue(returnedValue string) bool {
+	for _, value := range e.InaccessibleValues {
```
devsergiy: maybe we could use slices.Contains?

v2/pkg/ast/ast_enum_value_definition.go
```diff
@@ -21,6 +21,7 @@ type EnumValueDefinition struct {
 	EnumValue     ByteSliceReference // e.g. NORTH (Name but not true, false or null
 	HasDirectives bool
 	Directives    DirectiveList // optional, e.g. @foo
+
```
devsergiy: leftover?

devsergiy (approved):

devsergiy (approved):

PR #938, opened 2024-10-25
chore: improve epoll impl

jensneuse:
I was able to remove the requirement to read twice into a timeout. This allows us to poll faster, which resulted in unnecessary CPU pressure, which I fixed with a ticker. I've made the ticker dynamic so that it can "speed" up when the events queue is full. I've tested this with our benchmarks. Finally, I've tweaked the defaults to balance CPU pressure, latency and throughput for 10k subscriptions and higher. In addition, I've moved some code out of the protocol handlers to simplify the code a bit.


StarpTech (approved):
LGTM

PR #939, opened 2024-10-25
feat: add further apollo-compatible error support

Aenimus:


Aenimus:
Required for https://github.com/wundergraph/cosmo/pull/1311

v2/pkg/graphqlerrors/errors.go
```diff
@@ -49,6 +49,18 @@ func RequestErrorsFromError(err error) RequestErrors {
 	}
 }
 
+// Used by Cosmo
+func RequestErrorsFromApolloCompatibilityError(err *operationreport.ApolloCompatibilityError) RequestErrors {
```
StarpTech: Looks to me that when used only in Cosmo that we can implement it in the router. All interfaces are public.

v2/pkg/engine/resolve/resolve.go
```diff
@@ -74,6 +74,10 @@ type Resolver struct {
 	subscriptionTools *sync.Pool
 }
 
+func (r *Resolver) Options() ResolvableOptions {
+	return r.options.ResolvableOptions
```
StarpTech: Sure, that we need to create a public getter? `resolve.ResolverOptions` is created in the Router.

StarpTech: I just saw your cosmo PR. If you need access to the Apollo flags you can simply pass `apolloCompatibilityFlags` to the preHandler.

Aenimus: There is no reason for it to be public besides needing it in the prehandler. But it's just Apollo Compatibility flags, so I am not sure it matters that it is exposed. How would you pass the options to the prehandler? It would have to be done upon creation of the options.

v2/pkg/graphqlerrors/errors.go
Aenimus: I just put it in the same place as the original. I can move it to the router.

v2/pkg/operationreport/externalerror.go
```diff
@@ -26,12 +25,18 @@ const (
 	UnknownFieldOfInputObjectErrMsg         = `Field "%s" is not defined by type "%s".`
 	DuplicatedFieldInputObjectErrMsg        = `There can be only one input field named "%s".`
 	ValueIsNotAnInputObjectTypeErrMsg       = `Expected value of type "%s", found %s.`
+	GraphQLValidationFailed                 = "GRAPHQL_VALIDATION_FAILED"
+	httpStatusBadRequest                    = 400
```
devsergiy: Maybe use existing constant from std package? `http.StatusBadRequest`

v2/pkg/operationreport/externalerror.go
```diff
@@ -26,12 +25,18 @@ const (
 	UnknownFieldOfInputObjectErrMsg         = `Field "%s" is not defined by type "%s".`
 	DuplicatedFieldInputObjectErrMsg        = `There can be only one input field named "%s".`
 	ValueIsNotAnInputObjectTypeErrMsg       = `Expected value of type "%s", found %s.`
+	GraphQLValidationFailed                 = "GRAPHQL_VALIDATION_FAILED"
```
devsergiy: GraphQLValidationFailedErrorCode?

v2/pkg/variablesvalidation/variablesvalidation.go
```diff
@@ -11,8 +11,13 @@ import (
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/operationreport"
 )
 
+const (
+	BadUserInput = "BAD_USER_INPUT"
```
devsergiy: proposal

should we create separate package `operationreport/errorcodes` with const.go? to use it like this

```
errorcodes.GraphQLValidationFailed
errorcodes.BadUserInput
```

devsergiy (changes requested):

v2/pkg/variablesvalidation/variablesvalidation.go
devsergiy: with this approach we don't need ErrorCode prefix to const name

devsergiy: we could also move there `InvalidGraphqlErrorCode` from resolve package

devsergiy (approved):

PR #941, opened 2024-10-26
fix(epoll): limit concurrency

StarpTech:
This PR ensures that we always process epoll events with limited concurrency. Additionally, to avoid spinning the CPU in the event loop, we ensure that minimum `delay` is waited. This takes the processing time into consideration. Those changes produces very stable CPU times.

We should potentially expose EpollConfiguration to allow users to tweak the performance in more demanding environments.

@jensneuse I was not able to reproduce an epoll issue while handling unsubscribe events concurrently. Let me know if you have an reproduction case.


df-wg (approved):
Did you see any performance change as a result of this?

StarpTech:
@df-wg much less goroutine management 22 system threads vs 300. CPU utilization seems to be more stable.

PR #942, opened 2024-10-27
feat: auto create version bumps on every release to master

df-wg:
This PR adds 3 elements:
* a workflow to ensure that the PR titles fit conventional commit, copied from cosmo
* a workflow to auto-tag on release to master
* a workflow that will send a repository dispatch (https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows), using https://github.com/peter-evans/repository-dispatch. This will require that a service admin generates a PAT token (it can be under the wundergraph org instead of tied to a user) with rights to read `cosmo` and saves it as a secret in this repo 

There will be a corresponding PR in cosmo with the rest of the cycle, accepting that dispatch and operating on it


.github/workflows/autotag.yaml
```diff
@@ -0,0 +1,15 @@
+name: Bump version
```
StarpTech: How does it behave with release candidates? Our current version is [v2.0.0-rc.112](https://github.com/wundergraph/graphql-go-tools/releases/tag/v2.0.0-rc.112) In my opinion this isn't a blocker yet, we can also iterate with patch, minor and majors. @devsergiy

.github/workflows/autotag.yaml
```diff
@@ -0,0 +1,15 @@
...
+on:
+  push:
+    branches:
+      - master
+jobs:
```
StarpTech: This is a great idea but I'm not sure if we should automate it with the Cosmo repository. Often you develop side-by-side in the Cosmo and engine branch. After merging and releasing the engine changes, you want to bump the version in the same cosmo branch. Additionally, what will happen when you do two merges in a row? Will it update an existing release PR in cosmo or create two?

Therefore, do we really want to release any commit to master? I think we should try https://github.com/wundergraph/graphql-go-tools/pull/942#discussion_r1818168489 which allows to release when the release PR is merged.

.github/workflows/autotag.yaml
```diff
@@ -0,0 +1,15 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Bump version and push tag
+        id: tag_version
+        uses: mathieudutour/github-tag-action@v6.2
```
StarpTech: I'd give https://github.com/googleapis/release-please a try which we use for Helm Chart release in Cosmo. It maintains release PR's for changes and creates a release summary and github release based on conventional commits. It also takes care of tagging. 

This flow has the side effect of making people more aware of what changes will be released.

StarpTech (changes requested):

.github/workflows/release.yaml
```diff
@@ -0,0 +1,18 @@
...
+  contents: write
+  pull-requests: write
+
+name: release-please
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: googleapis/release-please-action@v4
```
StarpTech: I think you have to specify the mod file. In this repo we have multiple modules. We only want to release v2.

StarpTech (changes requested):

PR #943, opened 2024-10-27
feat: make epoll impl lock free and non CPU blocking when idle

jensneuse:
The previous implementation relied on mutexes and locking to achieve concurrency.
This resulted in a lot of complex code.
In addition, I believe that it was possible that events are being handled while we continue with the next loop as we didn't wait for their completion.
Furthermore, one shortcoming of the previous version was that we would always have some CPU spinning when there are 0 connections as we had no means to stop the loop.

This refactor replaces the shared memory + mutex approach with an event loop.
As a result, we have 0% CPU in idle.
Another benefit is that we're able to handle incoming connection events parallel to managing connections.
This is possible as we've moved all connection management to the main loop while handling connection events in the workers. Previously, workers were also managing connections which made a mutex and locking necessary.
As a consequence, the code in most areas is now single threaded and much simpler.

### DIFF
--- a/v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
+++ b/v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
@@ -47,12 +47,21 @@ type subscriptionClient struct {
 	epoll       epoller.Poller
 	epollConfig EpollConfiguration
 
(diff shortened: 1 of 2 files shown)


StarpTech (approved):
LGTM

PR #949, opened 2024-11-03
fix: use parser pool, avoid panic during parse

StarpTech:
This PR ensures that we use our parser pool for all parse operations. Therefore we wanted to panic in several places. I removed those places and favor explicit error handling or partial responses instead of service interruption.

The changes reduces object allocation by an big magnitude and comes with a nice memory improvements ~6% in my simplified benchmark. I do expect much bigger saving in real-life scenarios. This change can be significant for use cases with large variables input.

Inuse_objects

CleanShot 2024-11-03 at 14 26 27@2x

Alloc_space

CleanShot 2024-11-03 at 14 34 24@2x


df-wg (approved):
lgtm!

jensneuse (approved):

devsergiy (approved):

PR #982, opened 2024-11-16
fix(ws): deadlock on unsubscribe when epoll disabled

StarpTech:
We have used the unsubscription channel of the epoll implementation for both paths, although epoll can be disabled. This led to a deadlock.

Related Cosmo PR: https://github.com/wundergraph/cosmo/pull/1380

### DIFF
--- a/v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
+++ b/v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
@@ -29,6 +29,26 @@ import (
 
 const ackWaitTimeout = 30 * time.Second
 
+type epollState struct {
+	// connections is a map of fd -> connection to keep track of all active connections
+	connections    map[int]*connection
+	hasConnections atomic.Bool
+	// triggers is a map of subscription id -> fd to easily look up the connection for a subscription id
+	triggers map[uint64]int
+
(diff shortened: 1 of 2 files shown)


v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -29,6 +29,26 @@ import (
 
 const ackWaitTimeout = 30 * time.Second
 
+type epollState struct {
+	// connections is a map of fd -> connection to keep track of all active connections
```
Aenimus, line 33: fd?

StarpTech: file descriptor

Aenimus: Am I the exception that it's not clear? Is it an acronym most know?

jensneuse (approved):

Aenimus (approved):
LGTM

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
StarpTech: yes, in the domain

PR #993, opened 2024-11-21
feat: propagate operation name

Noroth:


v2/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -398,14 +400,40 @@ func (p *Planner[T]) ConfigureSubscription() plan.SubscriptionConfiguration {
...
 
+func (p *Planner[T]) buildUpstreamOperationName(ref int) string {
+	operationName := p.visitor.Operation.OperationDefinitionNameBytes(ref)
+	if len(operationName) == 0 {
+		return ""
+	}
+
+	fetchID := strconv.Itoa(p.dataSourcePlannerConfig.FetchID)
+
+	builder := strings.Builder{}
```
StarpTech, line 411: Is the usage of the builder really worth it instead of using sprintf?

Noroth: I would say it was worth it with the previous implementation where we also had the subpath. The builder is still twice as fast as using `fmt.Sprintf` but we are talking about `35` and `78` ns.

StarpTech (approved):
LGTM

PR #997, opened 2024-11-25
fix: run actions in release PRs

Noroth:
PRs created via Github actions bot cannot run Github actions. https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

We need a token with more permissions to trigger those actions for release PRs


StarpTech (approved):
LGTM

PR #998, opened 2024-11-26
[DO NOT MERGE] feat: reduce allocs & improve throughput

jensneuse:
...by introducing a limit buffer pool and fetch buffer heuristics

I figures out that we can reduce allocs by almost 50% and increase throughput by 30% by doing 2 changes:

1. add a limit buffer pool to resolvable instead of using sync.Pool or no pool at all
2. add heuristics to loader to estimate the ideal size of fetch buffers

Both changes reduce bytes.growSlice operations to almost zero in hot paths, see below.

before:

image

after:

image

### DIFF
--- a/v2/pkg/engine/datasource/httpclient/nethttpclient.go
+++ b/v2/pkg/engine/datasource/httpclient/nethttpclient.go
@@ -199,11 +199,6 @@ func makeHTTPRequest(client *http.Client, ctx context.Context, url, method, head
 	}
 
 	if !enableTrace {
-		if response.ContentLength > 0 {
-			out.Grow(int(response.ContentLength))
-		} else {
-			out.Grow(1024 * 4)
-		}
 		_, err = out.ReadFrom(respReader)
 		return
 	}
--- a/v2/pkg/engine/resolve/fetch.go
+++ b/v2/pkg/engine/resolve/fetch.go
@@ -1,8 +1,10 @@
 package resolve
 
 import (
+	"bytes"
 	"encoding/json"
 	"slices"
+	"sync"
 
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/ast"
 )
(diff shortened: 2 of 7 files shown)


v2/pkg/pool/limitbufferpool.go
```diff
@@ -0,0 +1,97 @@
...
+
+func NewLimitBufferPool(ctx context.Context, options LimitBufferPoolOptions) *LimitBufferPool {
+	if options.MaxBufferSize == 0 {
+		options.MaxBufferSize = 1024 * 1024 * 10 // 10MB
+	}
+	if options.DefaultBufferSize < 1024*8 {
+		options.DefaultBufferSize = 1024 * 8 // 8KB
+	}
+	if options.MaxBuffers == 0 {
+		options.MaxBuffers = runtime.NumCPU()
```
StarpTech: This won't be picked up in container environments. It will always use the cpu count of the hosts.

StarpTech (changes requested):

v2/pkg/pool/limitbufferpool.go
```diff
@@ -0,0 +1,97 @@
...
+	"bytes"
+	"context"
+	"runtime"
+	"time"
+)
+
+// LimitBufferPool is a pool of buffers that is limited in size and is limiting the max size of buffers that should be recycled
+// In addition, it runs a GC runtime that randomly resets a buffer every second to keep the memory usage low when usage is low
+// This is an alternative to sync.Pool, which can grow unbounded rather quickly
+type LimitBufferPool struct {
```
StarpTech, line 12: Please add a few tests that ensure the correctness and efficiency of the library. Testing this in an integration test will be hard without automation.

StarpTech (changes requested):

v2/pkg/pool/limitbufferpool.go
jensneuse: added

v2/pkg/pool/limitbufferpool.go
```diff
@@ -0,0 +1,98 @@
...
+
+func (p *LimitBufferPool) runGC(ctx context.Context) {
+	ticker := time.NewTicker(p.options.GCTime)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			b := p.Get()
```
StarpTech, line 75: This means, it will reset a buffer every second. You can't ensure that you get an existing buffer. In the worst you allocate and reset it in one step. I have never seen this in real world. How can we be sure that this works out in advanced use case?

v2/pkg/pool/limitbufferpool.go
StarpTech: Did you forget to push?

v2/pkg/engine/datasource/httpclient/nethttpclient.go
```diff
@@ -199,11 +199,6 @@ func makeHTTPRequest(client *http.Client, ctx context.Context, url, method, head
 	}
 
 	if !enableTrace {
-		if response.ContentLength > 0 {
```
StarpTech, line 202: Why was it removed?

v2/pkg/engine/resolve/fetch.go
```diff
@@ -20,6 +22,42 @@ type Fetch interface {
 	FetchKind() FetchKind
 	Dependencies() FetchDependencies
 	DataSourceInfo() DataSourceInfo
+
+	GetBuffer() *bytes.Buffer
+	ReportResponseSize(out *bytes.Buffer)
+}
+
+// FetchBufferSizeCalculator calculates the right size for a buffer based on the previous 64 fetches
```
StarpTech, line 30: How can we be right by checking the last 64 fetches that this number represents the average / quantile of the overall traffic? I think this heuristic is too simple and doesn't reflect real world scenarios (mixed usage). I think working with a pre sized / capped buffer brings similar improvements.

v2/pkg/engine/datasource/httpclient/nethttpclient.go
jensneuse: moved into a different place

v2/pkg/pool/limitbufferpool.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (p *LimitBufferPool) runGC(ctx context.Context) {
+	ticker := time.NewTicker(p.options.GCTime)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
```
StarpTech, line 74: Why do we think that resetting the buffer is right after X amount? We did not take into consideration if the buffer is actually right sized and used. In that case, why do we want to reclaim it?

jensneuse:
We want to merge this at a later stage once we have better testing architecture in place.

PR #1001, opened 2024-11-27
feat: add http datasource onfinished hook 

df-wg:
As part of our efforts in https://github.com/wundergraph/cosmo/pull/1401, we need to have access to the request and response from the calls to subgraphs. As a result, we're exposing a new hook which is specialized for a HTTP datasource
```
OnHttpFinished(ctx context.Context, ds DataSourceInfo, err error, request *http.Request, response *http.Response)
```

which users can use to gain access to the sanitized request/response and properly handle that information


v2/pkg/engine/resolve/loader.go
```diff
@@ -178,6 +193,9 @@ func (l *Loader) resolveSingle(item *FetchItem) error {
 		if l.ctx.LoaderHooks != nil && res.loaderHookContext != nil {
 			l.ctx.LoaderHooks.OnFinished(res.loaderHookContext, res.statusCode, res.ds, goerrors.Join(res.err, l.ctx.subgraphErrors))
 		}
+		if l.ctx.HttpLoaderHooks != nil && res.httpResponseContext != nil {
```
StarpTech: We can solve all use cases with a single hook implementation. We can get rid of `LoaderHooks`. You should be able to call `OnLoad` within the `Load` function of the DataSource implementation and `OnHttpFinished` in the loader. This would also allow us to unify telemetry instrumentation in the router because currently we use a custom transport + LoaderHooks to collect metrics.

v2/pkg/engine/resolve/loader.go
```diff
@@ -40,6 +41,10 @@ type LoaderHooks interface {
 	OnFinished(ctx context.Context, statusCode int, ds DataSourceInfo, err error)
 }
 
+type HttpLoaderHooks interface {
+	OnHttpFinished(ctx context.Context, ds DataSourceInfo, err error, request *http.Request, response *http.Response)
```
StarpTech: Let's avoid too many arguments and introduce a struct. Additionally, it would not be safe for a user of the hooks to consume the request / response body because it needs to be re-created before passing it to the hook. This will have a performance impact and should not run by default.

StarpTech (changes requested):

v2/pkg/engine/resolve/loader.go
jensneuse: Typically, above 3 args it's best practice to use a struct or a combination of ctx + struct.

v2/pkg/engine/resolve/loader.go
```diff
@@ -37,7 +38,66 @@ type LoaderHooks interface {
...
+type DataSourceInfo struct {
+	ID   string
+	Name string
+}
+
+type ResponseInfo struct {
+	StatusCode int
+	Err        error
+	// When accessing the request, consumers should avoid accessing the Body field, because it is already read and closed.
+	Request         *http.Request
```
StarpTech: Go comments always start with the Field / method to describe.

v2/pkg/engine/resolve/loader.go
```diff
@@ -37,7 +38,66 @@ type LoaderHooks interface {
...
+	ResponseHeaders http.Header
```
StarpTech, line 56: This would also not be safe to use. You didn't clone them.

v2/pkg/engine/datasource/httpclient/nethttpclient.go
```diff
@@ -190,8 +194,10 @@ func makeHTTPRequest(client *http.Client, ctx context.Context, url, method, head
 		return err
 	}
 	defer response.Body.Close()
+	request.Header = redactHeaders(request.Header)
```
StarpTech: This looks wrong to me because we don't want to modify the original headers. We only want to redact them for analytics, tracing.

StarpTech (changes requested):

v2/pkg/engine/datasource/httpclient/nethttpclient.go
```diff
@@ -216,12 +220,12 @@ func makeHTTPRequest(client *http.Client, ctx context.Context, url, method, head
 		Request: TraceHTTPRequest{
 			Method:  request.Method,
 			URL:     request.URL.String(),
-			Headers: redactHeaders(request.Header),
+			Headers: redactHeaders(request.Header.Clone()),
```
df-wg: For safety purposes, I thought we should clone the headers, since we're also exposing them through the hooks

v2/pkg/engine/datasource/httpclient/nethttpclient.go
df-wg: I could argue that we should modify the headers, because clients shouldn't be accessing sensitive information like `authorization`/`cookie` headers. I can also expose the `redactHeaders` function so that we can utilize it in our hook implementation

v2/pkg/engine/datasource/httpclient/nethttpclient.go
StarpTech: They are already copied by `json.Marshal`

StarpTech (approved):
LGTM

PR #1016, opened 2024-12-15
fix: set request in hook context before send request, in case of error

df-wg:
This will be added to https://github.com/wundergraph/cosmo/pull/1445, to ensure that it's more resilient

Previously, we weren't setting the request (even when we really should have had access), because in error cases we were exiting before setting the request/response context. This ensures that we can pass the correct information to the access logs


StarpTech (approved):
LGTM

PR #1030, opened 2025-01-10
fix: heartbeat go routine gone rogue

alepane21:
When running integration tests with high concurrency, race warnings frequently appeared, often triggered by the heartbeat. It seems that after some refactoring, a goroutine was left behind without any synchronization mechanism, leading to the issue.

The fix should be straightforward: simply remove the goroutine.


df-wg (commented):
Have you checked the router with this adjustment? From what I recall, this caused race issues in the router as a result

alepane21:
> Have you checked the router with this adjustment? From what I recall, we had initially added in the go routine because without it, there were race issues in the router as a result

I made this change to solve a race issues with the heart beat on the router that I caught sometime when running integration tests, e.g.: https://github.com/wundergraph/cosmo/actions/runs/12675890778/job/35374593911#step:5:776 . With this change applied I have not seen data races anymore in the integration tests. Maybe the data race that you are talking about has been fixed with https://github.com/wundergraph/cosmo/pull/1410 ?

df-wg (approved):

PR #1033, opened 2025-01-17
fix: delete leftover heartbeat connections

df-wg:
Users reported seeing the below sigfault when a connection to a subgraph is interrupted over multipart: 
```
cosmo-router     | 21:07:16 PM ERROR core/graphql_handler.go:380 Unable to write error response {"hostname": "ad1991cdcfd2", "pid": 1, "component": "@wundergraph/router", "service_version": "0.158.0", "request_id": "ad1991cdcfd2/RTnCLg1J8M-000022", "trace_id": "77476c6fe295cd7dafeb71e00183879b", "error": "context canceled"}
cosmo-router     | github.com/wundergraph/cosmo/router/core.(*GraphQLHandler).WriteError
cosmo-router     |      github.com/wundergraph/cosmo/router/core/graphql_handler.go:380
cosmo-router     | github.com/wundergraph/graphql-go-tools/v2/pkg/engine/resolve.(*Resolver).handleHeartbeat.func1
cosmo-router     |      github.com/wundergraph/graphql-go-tools/v2@v2.0.0-rc.136/pkg/engine/resolve/resolve.go:431
```

After investigating, the cause seemed to be a number of times where we deleted the subscription trigger but didn't clean up the heartbeat (which is running in a separate thread), causing it to write on a non-existent context. This PR cleans that up


StarpTech (commented):
Great catch, looks reasonable to me. Can we add a test for it here or in the router?

v2/pkg/engine/resolve/resolve.go
```diff
@@ -51,6 +51,7 @@ type Resolver struct {
 	maxConcurrency chan struct{}
 
 	triggers               map[uint64]*trigger
+	heartbeatSubLock       *sync.RWMutex
```
Noroth: Why is this a `RWMutex`? I can't see that you are using `RLock` anywhere

v2/pkg/engine/resolve/resolve.go
```diff
@@ -482,8 +493,14 @@ func (r *Resolver) handleTriggerDone(triggerID uint64) {
...
 		}
-		for _, s := range trig.subscriptions {
+		for c, s := range trig.subscriptions {
 			s.writer.Complete()
+			s.mux.Lock()
+			r.heartbeatSubLock.Lock()
+			delete(r.heartbeatSubscriptions, c)
+			r.heartbeatSubLock.Unlock()
+			delete(trig.subscriptions, c)
+			s.mux.Unlock()
```
Noroth: This looks a bit complicated to understand, would be good to have a function with a descriptive name or some kind of commenting to understand the different muxes?

v2/pkg/engine/resolve/resolve.go
```diff
@@ -642,7 +661,9 @@ func (r *Resolver) handleRemoveSubscription(id SubscriptionIdentifier) {
 				if ctx.Context().Err() == nil {
 					s.writer.Complete()
 				}
+				r.heartbeatSubLock.Lock()
 				delete(r.heartbeatSubscriptions, ctx)
+				r.heartbeatSubLock.Unlock()
```
Noroth: I personally would wrap these kind of blocks into a functions like `safeDeleteXYZ` 

e.g.
```
func safe(){
  r.lock()
  defer r.unlock() // make sure to always unlock
  ... function logic
}
```

v2/pkg/engine/resolve/resolve.go
df-wg: Good point, I refactored this into a cleaner interface

Noroth (commented):
Looks good. Do you think we should have one more test with 50 subscriptions at once and make sure that all of them are deleted properly?

e.g.:
`require.Eventually`

v2/pkg/engine/resolve/resolve_test.go
```diff
@@ -5186,6 +5186,48 @@ func TestResolver_ResolveGraphQLSubscription(t *testing.T) {
...
+
+		ctx := &Context{
+			ctx: context.Background(),
+			ExecutionOptions: ExecutionOptions{
+				SendHeartbeat: true,
+			},
+		}
+
+		for i := 1; i <= 10; i++ {
+			id.ConnectionID = int64(i)
```
Noroth, line 5209: I believe you meant to run this loop in parallel right?
I would guess AwaitComplete is a blocking operation

df-wg: I think that it wouldn't work if we run it in parallel, and it would be more likely to introduce a flakiness into the test. When I tried making it parallel locally, it failed

Noroth (approved):

PR #1034, opened 2025-01-18
feat: implement variables mapper

devsergiy:


v2/pkg/astnormalization/variables_mapping.go
```diff
@@ -0,0 +1,122 @@
...
+		// set new variable name for all variable values
+		for _, variableValueRef := range variableItem.valueRefs {
+			v.operation.VariableValues[variableValueRef].Name = newVariableName
+		}
+
+		// set new variable name for variable definition
+		v.operation.VariableValues[v.operation.VariableDefinitions[variableItem.variableDefinitionRef].VariableValue.Ref].Name = newVariableName
+	}
+
+	slices.SortFunc(v.operation.OperationDefinitions[v.operationRef].VariableDefinitions.Refs, func(i, j int) int {
```
jensneuse, line 104: please add a comment why this is important

v2/pkg/astnormalization/variables_mapping.go
```diff
@@ -0,0 +1,122 @@
...
+		return
+	}
+	if len(v.Ancestors) == 0 || v.Ancestors[0].Kind != ast.NodeKindOperationDefinition {
+		return
+	}
+
+	varValueRef := v.operation.Arguments[ref].Value.Ref
+	varNameBytes := v.operation.VariableValueNameBytes(varValueRef)
+	// explicitly convert to string to convert unsafe
+	varName := string(varNameBytes)
```
jensneuse: you could do this in line 83

v2/pkg/astnormalization/variables_mapping.go
```diff
@@ -0,0 +1,122 @@
...
+	v.variables = make([]*variableItem, 0, len(operation.VariableDefinitions))
+}
+
+func (v *variablesMappingVisitor) EnterOperationDefinition(ref int) {
+	v.operationRef = ref
+}
+
+const alphabet = `abcdefghijklmnopqrstuvwxyz`
+
+func (v *variablesMappingVisitor) generateUnusedVariableMappingName() []byte {
```
jensneuse, line 166: add a comment to get an idea of what it's going to generate

jensneuse (approved):

v2/pkg/astnormalization/variables_mapper_test.go
```diff
@@ -0,0 +1,432 @@
...
+					}
+				}`,
+			variablesMapping: map[string]string{
+				"a": "a",
+				"b": "d",
+				"c": "c",
+			},
+		},
+		{
+			name: "11 Reused external variable",
```
Aenimus, line 391: Where is 10? :D

v2/pkg/astnormalization/variables_mapper_test.go
```diff
@@ -0,0 +1,432 @@
...
+					object(id: $a) {
+						echo(value: $b)
+						echoTwo(value: $b)
+					}
+				}`,
+			variablesMapping: map[string]string{
+				"a": "a",
+				"b": "b",
+			},
+		},
```
df-wg, line 431: Can you add one error case, where `report.HasErrors()` would be `true`?

v2/pkg/astnormalization/variables_mapper.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/operationreport"
+)
+
+type VariablesMapper struct {
+	walker                  *astvisitor.Walker
+	variablesMappingVisitor *variablesMappingVisitor
+}
+
+func NewVariablesMapper() *VariablesMapper {
+	walker := astvisitor.NewWalker(8)
```
df-wg: Can you either make this `8` a const with a clear name, or add a comment explaining why it's used?

v2/pkg/astnormalization/variables_mapping.go
```diff
@@ -0,0 +1,122 @@
...
+	var i, k int64
+
+	for i = 1; i < math.MaxInt16; i++ {
```
df-wg, line 169: Can you add a test case for what happens if there are more than 26 variables, just so we can ensure that it's handled correctly for large complex cases?

df-wg (commented):
⭐ 🤩 

Super impressive work!

v2/pkg/astnormalization/variables_mapper_test.go
devsergiy: actually it will be hard to do that, we do not produce any error in the visitor

so the query should be very corrupted to get an error from walker itself

v2/pkg/astnormalization/variables_mapper.go
devsergiy: it is ancestors size, we use 8 in every normalization rule in the normalization package

v2/pkg/astnormalization/variables_mapping.go
devsergiy: next names will be `aa` `aaa` and so on

I could add a separate test for names generator

v2/pkg/astnormalization/variables_mapper_test.go
devsergiy: 10 was fully equal to nine, so this one could be 10

v2/pkg/astnormalization/variables_mapper.go
devsergiy: I could add astvisitor.NewDefaultWalker with built in ancestors size

df-wg: No need for now, but it would be great for the future (as someone who knows very little about this package)

v2/pkg/astnormalization/variables_mapping.go
df-wg: If possible that would be great, but if it'll hold things up then it can be deferred

v2/pkg/engine/resolve/resolve_test.go
```diff
@@ -4692,6 +4692,79 @@ func TestResolver_WithHeader(t *testing.T) {
...
+				RemapVariables: tc.remap,
+			}
+
+			ctrl := gomock.NewController(t)
+			fakeService := NewMockDataSource(ctrl)
+			fakeService.EXPECT().
+				Load(gomock.Any(), gomock.Any(), gomock.AssignableToTypeOf(&bytes.Buffer{})).
+				Do(func(ctx context.Context, input []byte, w io.Writer) (err error) {
+					actual := string(input)
+					assert.Equal(t, "bazz", actual)
```
jensneuse: I would prefer if each tc could have a different value instead of using "bazz" for all cases.

PR #1043, opened 2025-01-23
fix: remove semaphore from ResolveGraphQLSubscription

alepane21:
The semaphore in the sync method was blocking requests after the 1024th from proceeding.

### DIFF
--- a/v2/pkg/engine/resolve/resolve.go
+++ b/v2/pkg/engine/resolve/resolve.go
@@ -50,13 +50,14 @@ type Resolver struct {
 	options        ResolverOptions
 	maxConcurrency chan struct{}
 
-	triggers               map[uint64]*trigger
-	heartbeatSubLock       *sync.Mutex
-	heartbeatSubscriptions map[*Context]*sub
-	events                 chan subscriptionEvent
-	triggerEventsSem       *semaphore.Weighted
-	triggerUpdatesSem      *semaphore.Weighted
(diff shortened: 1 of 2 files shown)


v2/pkg/engine/resolve/resolve.go
```diff
@@ -805,11 +805,6 @@ func (r *Resolver) AsyncUnsubscribeClient(connectionID int64) error {
 }
 
 func (r *Resolver) ResolveGraphQLSubscription(ctx *Context, subscription *GraphQLSubscription, writer SubscriptionResponseWriter) error {
-	if err := r.triggerEventsSem.Acquire(r.ctx, 1); err != nil {
```
df-wg, line 808: Can you clarify why this helps a bit more? Are we removing the semaphore entirely? Or why only remove it from the sync and not the async method?

jensneuse: Async only subscribes, so we limit the concurrency in creating new subscriptions by that. Synchronous Subscriptions on the other hand never return if the subscription is not completed. So we cannot use the same sema for this. We "could" now add another sema to limit max open synchronous subscriptions. However, as a synchronous subscription is still an HTTP request, it's also ok to just use e.g. an HTTP loadbalancer to limit max concurrent HTTP requests. As such, I think removing the sema here is fine overall.

jensneuse (approved):

PR #1045, opened 2025-01-25
feat: add normalizedQuery to query plan and request info to trace

alepane21:


v2/pkg/engine/resolve/trace.go
```diff
@@ -59,16 +61,30 @@ func (r *TraceOptions) DisableAll() {
 	r.IncludeTraceOutputInResponseExtensions = false
 }
 
+type RequestData struct {
+	Method  string          `json:"method"`
+	URL     string          `json:"url"`
+	Headers http.Header     `json:"headers"`
+	Body    json.RawMessage `json:"body,omitempty"`
```
devsergiy: could we have body as a struct too?

```
type RequestBody struct {
  Query string
  OperationName string
  Variables json.RawMessage
```

devsergiy (approved):

PR #1048, opened 2025-01-29
fix: wait for updates in flight to be delivered before shutting down the trigger

alepane21:
After the last refactor we weren't waiting anymore for the in flight requests to be delivered, so I restored the previous behavior.

### DIFF
--- a/v2/pkg/engine/resolve/resolve_test.go
+++ b/v2/pkg/engine/resolve/resolve_test.go
@@ -4777,6 +4777,7 @@ type SubscriptionRecorder struct {
 	messages []string
 	complete atomic.Bool
 	mux      sync.Mutex
+	onFlush  func(p []byte)
 }
 
 func (s *SubscriptionRecorder) AwaitMessages(t *testing.T, count int, timeout time.Duration) {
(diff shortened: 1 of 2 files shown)


df-wg (approved):

StarpTech (changes requested):
Can we write a test for it or is it covered by the router?

alepane21:
> Can we write a test for it or is it covered by the router?

Mmm not sure if this is covered by the router test suite. I can make a test to check on this behavior, ti shouldn't be too complex, thanks for the suggestion!

StarpTech (approved):
LGTM

v2/pkg/engine/resolve/resolve_test.go
```diff
@@ -4777,6 +4777,7 @@ type SubscriptionRecorder struct {
 	messages []string
 	complete atomic.Bool
 	mux      sync.Mutex
+	onFlush  func(p []byte)
```
StarpTech, line 4780: This is just for testing right?

alepane21: Yes

Issue #405, opened 2022-08-18
astprinter: PrintIndent applies indentation twice


polarina:
Check out the following snippet: https://go.dev/play/p/FkEAZlnbd-E

The indentation is set to a single tab (`"\t"`), however, the printed output uses two tabs for indentation. Only one is expected.

---

```go
package main

import (
	"fmt"
	"os"

	"github.com/wundergraph/graphql-go-tools/pkg/ast"
	"github.com/wundergraph/graphql-go-tools/pkg/astparser"
	"github.com/wundergraph/graphql-go-tools/pkg/astprinter"
	"github.com/wundergraph/graphql-go-tools/pkg/operationreport"
)

const graph = `
	{ foo bar(id: "123") }
`

func main() {
	parser := astparser.NewParser()
	doc := ast.NewDocument()
	doc.Input.ResetInputBytes([]byte(graph))

	report := operationreport.Report{}
	parser.Parse(doc, &report)

	if report.HasErrors() {
		fmt.Println(report.Error())
		return
	}

	if err := astprinter.PrintIndent(doc, doc, []byte("\t"), os.Stdout); err != nil {
		fmt.Println(err)
	}

	fmt.Println()
}
```

Output:
```graphql
{
		foo
		bar(id: "123")
}
```

PR #1049, opened 2025-01-30
Fix: Honor the indentation

SoulPancake:
Re-opened this with tests to validate

For #405


SoulPancake:
Please review @jensneuse @devsergiy

pkg/astprinter/astprinter.go
```diff
@@ -1043,3 +1043,13 @@ func (p *printVisitor) writeFieldType(ref int) {
 	p.write(literal.SPACE)
 	p.must(p.document.PrintType(p.document.FieldDefinitionType(ref), p.out))
 }
+
+func countIndentDepth(indent []byte) int {
```
devsergiy, line 1047: Not fully understand what you are doing here?

In case the indent doesn't have a tab character returned value will be 0

pkg/astprinter/astprinter.go
```diff
@@ -93,12 +93,12 @@ func (p *printVisitor) indentationDepth() (depth int) {
 	case ast.NodeKindOperationDefinition,
 		ast.NodeKindFragmentDefinition:
 	default:
-		return 2
+		return countIndentDepth(p.indent)
```
devsergiy, line 96: Indentation depth is used to determine at what level of nesting we are

You just need to have 1 as a base value instead of 2 as it was before

Old behavior value 2 and indent `.` results in `..` value
With new behavior same could be achieved with value 1 and indent `..`

devsergiy (changes requested):
With this change a lot of old tests will be failing due to wrong indent, so such places should be fixed to have `....` 4 spaces if it was `..` and `..` 2 spaces if it was `.`

PR #1073, opened 2025-02-15
fix: deadlock when waiting on inflight events of a trigger

StarpTech:
This PR fixes an issue where `WaitGroup.Wait` caused a deadlock with the progressing update events because waiting in the event loop starves it forever.

Additionally, I removed the field `done`, which caused a race with a concurrent event from the WS providers. It also wasn't useful at all.

I battle-tested the change with a load test scenario where connections are reestablished multiple times a second.

### DIFF
--- a/v2/Makefile
+++ b/v2/Makefile
@@ -2,6 +2,11 @@
 test:
 	go test ./...
 
+test-fresh: clean-testcache test
+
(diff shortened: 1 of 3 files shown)


Aenimus (approved):
LGTM

PR #1076, opened 2025-02-17
feat(engine): mutex free subscription handling

StarpTech:
This PR removes all mutexes needed and uses channels to synchronize access in subscription handling. Every subscription spawns a single goroutine where the fetch and resolve process is run (connection write). Passing a task to a subscription is done on a buffered channel and on the event loop (single threaded). In that way, we can guarantee that we send messages in the order they were received, no matter how high the concurrency is. Before, it could happen that events arrived out of order. By using a buffered channel, we can increase the throughput of how many updates can be offloaded to a subscription worker while other subscriptions can continue. This implementation make use of the natural back-pressure behaviour of go channels. A `MaxSubscriptionFetchTimeout` is added to ensure that a single subscription fetch can't block the event loop forever. This timeout is configurable and set to `30s` by default.

Semaphores have been deleted as well because we now naturally limit the throughput by writing to subscriptions sequentially via channels. Additionally, heartbeat handling is done in the worker goroutine of every subscription. This simplified the implementation and removed another synchronization need.

We also massively improved the way synchronous subscriptions are managed. There is no difference in handling anymore, except that for SSE/Multipart, we wait until the client request is done or the subscription is completed and drained.

I did a few benchmarks with 200 concurrent subscriptions and around 10 triggers. Writing to all subscriptions was still pretty fast (<1ms, ~3ms Max). The next iteration of the implementation should focus on leveraging concurrency better. We should handle triggers individually so that the blast radius of a fast producer and slow writer and can't impact others. Today, a fast producer will result in increased latency but more stable CPU utilization due to backpressure.

Fixes ENG-6513


v2/pkg/engine/resolve/resolve.go
```diff
@@ -285,31 +270,65 @@ type trigger struct {
...
 	executor chan func()
+	// workChan is used to send work to the writer goroutine. All work is processed sequentially.
+	workChan chan func()
 }
 
-func (r *Resolver) executeSubscriptionUpdate(ctx *Context, sub *sub, sharedInput []byte) {
+// startWorker runs in its own goroutine to process fetches and write data to the client synchronously
+// it also takes care of sending heartbeats to the client but only if the subscription supports it
+func (s *sub) startWorker() {
+	heartbeatTicker := time.NewTicker(s.resolver.heartbeatInterval)
```
jensneuse, line 295: I'd remove the heartbeat if the subscription didn't need one.
This creates additional overhead like a goroutine internally.

v2/pkg/engine/resolve/resolve.go
```diff
@@ -285,31 +270,65 @@ type trigger struct {
...
+	defer heartbeatTicker.Stop()
+
+	for {
+		select {
+		case <-heartbeatTicker.C:
+			if s.heartbeat {
```
jensneuse: check not needed if we check it at the beginning

v2/pkg/engine/resolve/resolve.go
```diff
@@ -348,39 +359,34 @@ func (r *Resolver) executeSubscriptionUpdate(ctx *Context, sub *sub, sharedInput
...
-func (r *Resolver) handleEvents() {
-	done := r.ctx.Done()
-	heartbeat := time.NewTicker(r.multipartSubHeartbeatInterval)
-	defer heartbeat.Stop()
+// processEvents maintains the single threaded event loop that processes all events
+func (r *Resolver) processEvents() {
 	for {
 		select {
-		case <-done:
+		case <-r.ctx.Done():
```
jensneuse: why call Done() in a loop when we can call it once above?

v2/pkg/engine/resolve/resolve.go
```diff
@@ -410,51 +416,44 @@ func (r *Resolver) handleEvent(event subscriptionEvent) {
...
-		skipHeartbeat := now.Sub(s.lastWrite) < r.multipartSubHeartbeatInterval
-		s.mux.Unlock()
-		if skipHeartbeat || (c.Context().Err() != nil && errors.Is(c.Context().Err(), context.Canceled)) {
-			continue
-		}
 
-		if r.options.Debug {
-			fmt.Printf("resolver:heartbeat:subscription:%d\n", s.id.SubscriptionID)
-		}
+	if r.ctx.Err() != nil {
```
jensneuse, line 436: Is this necessary?

jensneuse (changes requested):

jensneuse:
I answered my own change requests with a draft for improvements, except one where I'm not sure:
https://github.com/wundergraph/graphql-go-tools/pull/1080

jensneuse (approved):

jensneuse (approved):

v2/pkg/engine/resolve/resolve.go
```diff
@@ -285,31 +270,75 @@ type trigger struct {
...
 type sub struct {
-	mux       sync.Mutex
 	resolve   *GraphQLSubscription
+	resolver  *Resolver
+	ctx       *Context
 	writer    SubscriptionResponseWriter
 	id        SubscriptionIdentifier
-	completed chan struct{}
+	heartbeat bool
 	lastWrite time.Time
```
alepane21: I think lastWrite is not used anymore

StarpTech: Good catch!

jensneuse (approved):

PR #1079, opened 2025-02-18
test: correctly enable subrequest http errors for the subrequest http error + ok error code test

endigma:
`ApolloRouterCompatibilitySubrequestHTTPError` to `true` when testing ok-ish error codes, still passes.


Aenimus (approved):
LGTM

PR #1080, opened 2025-02-18
feat(engine): mutex free subscription handling

jensneuse:


v2/pkg/engine/resolve/resolve.go
```diff
@@ -680,24 +700,32 @@ func (r *Resolver) handleTriggerUpdate(id uint64, data []byte) {
...
+		updates = append(updates, func() {
+			select {
+			case <-r.ctx.Done():
+			case <-c.ctx.Done():
+			case s.workChan <- fn: // Channel is buffered, but it can still block in case of a slow writer
+			}
+		})
 	}
+
+	go func() {
```
StarpTech, line 723: This won't guarantee order. There is still a race between the fastest batch of updates. It could happen that the first update of the first batch is slow so that the second batch overtake.

StarpTech:
Changes applied in original draft, thank you!

PR #1090, opened 2025-02-22
Fix: README.md code snippet

jupblb:
I've noticed some examples from the README do not work properly. Updated the code so that it compiles and outputs expected values.

Only https://github.com/wundergraph/graphql-go-tools/commit/81fc13c27ffb10f13660e44c8fd25e1d57d5645a should be properly reviewed. The second commit makes the indentation uniform for entire code snippet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,22 +95,22 @@ here's a speed run tutorial for you, based on how we use this library in Cosmo R
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


devsergiy:
Hi @jupblb 

thanks for your contribution!

PR #1091, opened 2025-02-24
Update gqlgen (and remove packr generate)

StevenACoffman:
Signed-off-by: Steve Coffman <steve@khanacademy.org>


StevenACoffman:
Ok, it looks like this is non-trivial, so I'm going to split this up into some smaller PRs and do this more incrementally, but I'll leave this PR around for a while to remind me of some of the files that are going to need be affected.

devsergiy:
Hi @StevenACoffman 

thanks for the try
Latest gqlgen requires some changes in the resolver, as the generated code changed a bit

also, you could remove this line here https://github.com/wundergraph/graphql-go-tools/blob/16602c669d4fcd4c66cf52586d5cbfd6324c2bc9/pkg/astvalidation/reference/main.go#L14

to not produce unnecessary changes

StevenACoffman:
Sure. Thanks for merging that other PR, btw.

PR #1092, opened 2025-02-24
feat(ast): add normalizer option to sort operation fields

alenalato:
### What
I've added support for sorting of selection sets fields.
### Why
Normalize them in a deterministic order.
### How
I created a new option for the AST Normalizer that allows passing a compare function and a built-in compare function that sorts  selections lexicographically

- fields by alias or name
- inline fragments by type condition
- fragment spreads by name

Sort is stable.
This normalizer acts after all other normalizers so it can behave correctly in case of merging and inlining of fragments.

### Testing
Updated the general normalization unit test and wrote dedicated unit tests for the new normalizer


alenalato:
> inline fragments by type condition

Is not flawless

PR #1094, opened 2025-02-25
fix(subscriptions): skip event after worker shutdown

StarpTech:
This will fix another deadlock where the write to the events channel was blocked because the receiver was shut down too early. We fixed this by listening for the closed channel signal in the write select construct to prevent blocking the event loop.


jensneuse (approved):

v2/pkg/engine/resolve/resolve.go
```diff
@@ -689,8 +694,18 @@ func (r *Resolver) handleTriggerUpdate(id uint64, data []byte) {
...
 		case <-c.ctx.Done():
+			// Skip sending the event if the client disconnected
 		case s.workChan <- fn:
+			// Send the work to the subscription worker
+		case <-s.completed:
+			// Stop sending if the subscription is completed. Otherwise, this could block the event loop forever
+			// when the subscription worker was shutdown after channel close but the event was still scheduled.
+			if s.resolver.options.Debug {
+				fmt.Printf("resolver:trigger:subscription:completed:%d:%d\n", s.id.ConnectionID, s.id.SubscriptionID)
+			}
```
endigma, line 708: Is this for debugging? Shouldn't it be a structured log?

StarpTech: That's how we did and doing it for now. I agree, it's not the best approach. There is a ticket in backlog to use Zap for everything in the engine to align it with the router.

endigma: if it's from scratch we should probably use https://pkg.go.dev/log/slog if it's an option, zap and zerolog etc are cool but there's a standard library structured logger now

StarpTech: Let's have this discussion outside of this PR.

alepane21 (approved):

PR #1095, opened 2025-02-25
feat: support files in nested variable input

devsergiy:


v2/pkg/astnormalization/variables_mapping.go
```diff
@@ -125,8 +125,7 @@ func (v *variablesMappingVisitor) EnterArgument(ref int) {
 		return
 	}
 
-	if v.operation.OperationDefinitions[v.operationRef].OperationType == ast.OperationTypeMutation &&
-		v.operation.ResolveTypeNameString(v.operation.VariableDefinitions[variableDefinitionRef].Type) == "Upload" {
+	if v.operation.ResolveTypeNameString(v.operation.VariableDefinitions[variableDefinitionRef].Type) == "Upload" {
```
Aenimus, line 128: Shouldn't you also take that it's a scalar?

v2/pkg/astnormalization/uploads/upload_finder.go
```diff
@@ -0,0 +1,270 @@
...
+
+func (v *UploadFinder) FindUploads(operation, definition *ast.Document, variables []byte, argRef int, argInputValueDefinitionRef int) (uploadPathMapping []UploadPathMapping, err error) {
+	v.definition = definition
+	v.operation = operation
+
+	if variables == nil || bytes.Equal(variables, []byte("null")) || bytes.Equal(variables, []byte("")) {
+		variables = []byte("{}")
+	}
+
+	v.variables, err = astjson.ParseBytesWithoutCache(variables)
```
jensneuse, line 77: If definition doesn't contain the Upload Scalar, can we immediately return here?

v2/pkg/astnormalization/uploads/upload_finder.go
```diff
@@ -0,0 +1,270 @@
...
+	if len(v.uploadPathMapping) > 0 {
+		return v.uploadPathMapping, nil
+	}
+
+	return nil, nil
+}
+
+func (v *UploadFinder) traverseValue(value ast.Value, valueTypeRef int) {
+	switch value.Kind {
+	case ast.ValueKindVariable:
```
jensneuse: Can you add more comments for each case to explain why and how you handle them? e.g Variables

v2/pkg/astnormalization/uploads/upload_finder.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+func (v *UploadFinder) traverseVariable(variableValueRef int, definitionValueTypeRef int) {
+	operationVariableNameBytes := v.operation.VariableValueNameBytes(variableValueRef)
+	operationVariableName := unsafebytes.BytesToString(operationVariableNameBytes)
+
+	v.currentVariableName = operationVariableNameBytes
+	v.currentVariableValue = v.variables.Get(operationVariableName)
+
+	// we rely on the fact that there is only one operation in the document
```
jensneuse, line 149: Is this the right approach?
From my understanding, e.g. persisted operations can have multiple ops per doc

jensneuse (changes requested):

v2/pkg/astnormalization/uploads/upload_finder.go
devsergiy: after normalization we should always have a single operation

jensneuse (approved):

PR #1099, opened 2025-02-27
fix: fix validation of variables used in nested fields of type list of an input object

devsergiy:
fix validation if lists with the variable value nested in the input object; merge valid arguments rule with the values validation rule

### DIFF
--- a/v2/pkg/astvalidation/operation_validation_test.go
+++ b/v2/pkg/astvalidation/operation_validation_test.go
@@ -2118,7 +2118,7 @@ func TestExecutionValidation(t *testing.T) {
 									doesKnowCommand(dogCommand: $catCommand)
 								}
 							}`,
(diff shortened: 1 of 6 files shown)


v2/pkg/astvalidation/operation_validation_test.go
```diff
@@ -4870,7 +4957,7 @@ type __Field {
     deprecationReason: String
 }
 
-"""ValidArguments provided to FieldSelections or Directives and the input fields of an
+"""Arguments provided to FieldSelections or Directives and the input fields of an
```
jensneuse, line 4960: intentional?

jensneuse (approved):

v2/pkg/astvalidation/operation_validation_test.go
devsergiy: yeah in a base schema it is just arguments
some old autoreplace

PR #1107, opened 2025-03-11
fix: allow custom scalar values of any kind

### LINKED ISSUE #1103: Custom numeric scalar value doesn't work as literal
Consider a custom scalar that is represented numerically in JSON rather than a string - such as the `Long` scalar in this test:

https://github.com/wundergraph/graphql-go-tools/blob/978f2420662736877a7d3ea711cb05b34f3d8e93/execution/engine/execution_engine_test.go#L762-L823

The test appears to be designed only to check that the scalar can be used as part of the response output.  There doesn't appear to be any existing test (that I could find) for custom scalars used as input.

When such a scalar is used for output, it is correctly represented numerically, as in `{"id": 1234}`, so I would expect that I would also be able to pass the values as numeric literals in input to a query or mutation.

```graphql
query Item {
    item(id: 1234)
}
```

... but in my testing, that fails with an error such as `Long cannot represent value: 1234`

It does work if it's passed as a string literal:

```graphql
query Item {
    item(id: "1234")
}
```

Or if passed as variable in either number or string format

```graphql
query Item($id: Long!} {
    item(id: $id)
}
```

```json
{"id": 1234}
```

```json
{"id":  "1234"}
```

(either work)

I notice that any custom resolver set appears to only be invoked when resolving an output field.  It doesn't appear to have any affect on parsing input.  Instead, that appears to be handled here:

https://github.com/wundergraph/graphql-go-tools/blob/978f2420662736877a7d3ea711cb05b34f3d8e93/v2/pkg/astvalidation/operation_rule_values.go#L282-L303

... which explains why variables work, and why the only allowed input for a custom scalar literal is a string.

mattjohnsonpint:
Custom scalar values should be able to be passed as any kind, not just strings.  For example, a custom 64-bit integer value would be passed numerically.

Fixes #1103 

For cleanup, also removes the now extraneous `builtInStringScalar` parameter to `valueSatisfiesScalarString`, as it is only called on the built-in string scalar type.


mattjohnsonpint:
@devsergiy - added tests.  Thanks.

mattjohnsonpint:
@devsergiy - ping

devsergiy (approved):

devsergiy:
@mattjohnsonpint 

Sorry for the delay
thanks again for the contribution!

PR #1113, opened 2025-03-14
fix: set proper write / read timeouts

StarpTech:


v2/pkg/engine/resolve/resolve.go
```diff
@@ -344,11 +354,7 @@ func (s *sub) complete() {
 	// to a subscription that is already done.
 	defer close(s.completed)
 
-	// We put the complete handshake to the work channel of the subscription
-	// to ensure that it is the last message that is sent to the client.
-	if s.ctx.Context().Err() == nil {
-		s.writer.Complete()
-	}
+	s.writer.Complete()
```
StarpTech, line 357: Not conditional because we want to sent complete to the client also when the subgraph has shutdown.

v2/pkg/engine/resolve/resolve.go
```diff
@@ -306,7 +306,12 @@ func (s *sub) startWorkerWithHeartbeat() {
 
 	for {
 		select {
-		case <-s.resolver.ctx.Done(): // Skip sending events if the resolver is shutting down
+		case <-s.ctx.ctx.Done():
+			// Complete when the client request context is done for synchronous subscriptions
+			s.complete()
```
StarpTech, line 311: Chan is closed in .complete and used to coordinate the shutdown of synchronous subscription.

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -28,7 +28,11 @@ import (
 	"go.uber.org/atomic"
 )
 
-const ackWaitTimeout = 30 * time.Second
+const (
```
StarpTech, line 31: We can make this configurable in the future but passing them down from Cosmo wasn't that straightforward. I'm gonna do this in a separate PR.

devsergiy (approved):

jensneuse (approved):

alepane21 (approved):

PR #1114, opened 2025-03-14
Return init error on init-failed ws connection

### LINKED ISSUE #1050: Terminating subscriptions not working on initial step
I was trying to use subscriptions (graphql-ws) and found a little problem on initial step: when `p.handleInit()` returns error then `engine.TerminateAllSubscriptions()` does nothing, because at this moment `engine.subCancellations` is empty, so it returns nil error to `UniversalProtocolHandler.Handle()`, and subscription keeps alive. 

(element is added to `engine.subCancellations` only on  start step, not initial).

```go
func (p *ProtocolGraphQLWSHandler) Handle(ctx context.Context, engine subscription.Engine, data []byte) error {
	...
	switch message.Type {
	case GraphQLWSMessageTypeConnectionInit:
		ctx, err = p.handleInit(ctx, message.Payload)
		if err != nil {
                        ...
			return engine.TerminateAllSubscriptions(&p.writeEventHandler)
		}

		...
	case GraphQLWSMessageTypeStart:
		return engine.StartOperation(ctx, message.Id, message.Payload, &p.writeEventHandler)
	...
	}

	return nil
}
```

I suppose we should return error instead of terminating all subscriptioins

kubaracek:
Fixes #1050

### DIFF
--- a/execution/subscription/websocket/protocol_graphql_ws.go
+++ b/execution/subscription/websocket/protocol_graphql_ws.go
@@ -307,7 +307,10 @@ func (p *ProtocolGraphQLWSHandler) Handle(ctx context.Context, engine subscripti
 		ctx, err = p.handleInit(ctx, message.Payload)
 		if err != nil {
 			p.writeEventHandler.HandleWriteEvent(GraphQLWSMessageTypeConnectionError, "", nil, errors.New("failed to accept the websocket connection"))
-			return engine.TerminateAllSubscriptions(&p.writeEventHandler)
+			if termErr := engine.TerminateAllSubscriptions(&p.writeEventHandler); termErr != nil {
+				return fmt.Errorf("failed to accept websocket connection: %v; additionally, failed to terminate subscriptions: %v", err, termErr)
+			}
+			return err
 		}
 
 		go p.handleKeepAlive(ctx)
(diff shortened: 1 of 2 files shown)


execution/subscription/websocket/protocol_graphql_ws.go
```diff
@@ -307,7 +307,10 @@ func (p *ProtocolGraphQLWSHandler) Handle(ctx context.Context, engine subscripti
 		ctx, err = p.handleInit(ctx, message.Payload)
 		if err != nil {
 			p.writeEventHandler.HandleWriteEvent(GraphQLWSMessageTypeConnectionError, "", nil, errors.New("failed to accept the websocket connection"))
-			return engine.TerminateAllSubscriptions(&p.writeEventHandler)
+			if termErr := engine.TerminateAllSubscriptions(&p.writeEventHandler); termErr != nil {
+				return fmt.Errorf("failed to accept websocket connection: %v; additionally, failed to terminate subscriptions: %v", err, termErr)
+			}
```
kubaracek, line 312: Same as the author of the original issue, I'm questioning whether it's  necessary here to be terminating the subscriptions instead of a simple:

```go
if err != nil {
  return err
}
```

Is there a good reason for it?

PR #1122, opened 2025-04-04
fix(websocket): handle ping/pong correctly

StarpTech:
This PR adds proper Ping/Pong support for upstream subgraphs. It ensures that the subgraph doesn't mark the connection as dead. For instance, the `graphql-ws` library supports lazy close timeouts. If the router doesn't signal that it is alive, the connection will be aborted, along with all subscriptions.

### DIFF
--- a/v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
+++ b/v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
@@ -626,7 +626,7 @@ type ConnectionHandler interface {
 	// HandleMessage handles the incoming message from the connection
 	HandleMessage(data []byte) (done bool)
 	// Ping used to send a ping message to the upstream server to remain the connection alive
-	// It also keeps track of when the last ping was sent and pong was received to early detect connection drops
+	// It also keeps track of when the last ping was sent and pong was received to initiate a close
 	Ping()
 	// ServerClose closes the connection from the server side
 	ServerClose()
--- a/v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
+++ b/v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
@@ -32,6 +32,7 @@ const (
 	writeTimeout       = 10 * time.Second
 	readMessageTimeout = 1 * time.Second
 	ackWaitTimeout     = 30 * time.Second
+	pongWaitTimeout    = 5 * time.Second
 )
 
 type netPollState struct {
@@ -68,7 +69,8 @@ type subscriptionClient struct {
 	hashPool                   sync.Pool
 	onWsConnectionInitCallback *OnWsConnectionInitCallback
 
-	readTimeout time.Duration
+	readTimeout  time.Duration
+	pingInterval time.Duration
 
 	netPoll       netpoll.Poller
 	netPollConfig NetPollConfiguration
@@ -126,6 +128,12 @@ func WithReadTimeout(timeout time.Duration) Options {
 	}
 }
 
+func WithPingInterval(interval time.Duration) Options {
+	return func(options *opts) {
+		options.pingInterval = interval
+	}
+}
+
 type NetPollConfiguration struct {
 	// Enable can be set to true to enable netPoll
 	Enable bool
@@ -165,6 +173,7 @@ func WithNetPollConfiguration(config NetPollConfiguration) Options {
 
 type opts struct {
 	readTimeout                time.Duration
+	pingInterval               time.Duration
 	log                        abstractlogger.Logger
 	onWsConnectionInitCallback *OnWsConnectionInitCallback
 	netPollConfiguration       NetPollConfiguration
@@ -189,8 +198,9 @@ func IsDefaultGraphQLSubscriptionClient(client GraphQLSubscriptionClient) bool {
 
 func NewGraphQLSubscriptionClient(httpClient, streamingClient *http.Client, engineCtx context.Context, options ...Options) GraphQLSubscriptionClient {
 	op := &opts{
-		readTimeout: time.Millisecond * 100,
-		log:         abstractlogger.NoopLogger,
+		readTimeout:  time.Millisecond * 100,
+		pingInterval: 10 * time.Second,
+		log:          abstractlogger.NoopLogger,
 	}
 
 	op.netPollConfiguration.ApplyDefaults()
@@ -205,6 +215,7 @@ func NewGraphQLSubscriptionClient(httpClient, streamingClient *http.Client, engi
 		engineCtx:       engineCtx,
 		log:             op.log,
 		readTimeout:     op.readTimeout,
+		pingInterval:    op.pingInterval,
 		hashPool: sync.Pool{
 			New: func() interface{} {
 				return xxhash.New()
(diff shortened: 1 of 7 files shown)


v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go
```diff
@@ -259,6 +273,32 @@ func (h *gqlTWSConnectionHandler) handleMessageTypePing() {
...
+func (h *gqlTWSConnectionHandler) handleMessageTypePong() {
+	h.lastPongReceivedAt = time.Now()
+}
+
+func (h *gqlTWSConnectionHandler) Ping() {
+
+	// Check if we have received a pong message in the pongWaitTimeout
+	// If not we assume the connection is dead and close it
+	if !h.lastPingSentAt.IsZero() {
+		if h.lastPongReceivedAt.Sub(h.lastPingSentAt) > pongWaitTimeout {
```
jensneuse: Can you make this a function that is easy to understand by name?
e.g. "hasReceivedPongBeforeTimeout" or similar

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go
```diff
@@ -259,6 +273,32 @@ func (h *gqlTWSConnectionHandler) handleMessageTypePing() {
...
+			h.log.Error("no pong received in time. Closing connection")
+			h.ServerClose()
```
jensneuse, line 307: Will this properly clean up the connection?

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go
```diff
@@ -259,6 +273,32 @@ func (h *gqlTWSConnectionHandler) handleMessageTypePing() {
...
 }
 
+func (h *gqlTWSConnectionHandler) handleMessageTypePong() {
+	h.lastPongReceivedAt = time.Now()
+}
+
+func (h *gqlTWSConnectionHandler) Ping() {
+
+	// Check if we have received a pong message in the pongWaitTimeout
+	// If not we assume the connection is dead and close it
```
jensneuse: Will it be confusing to close connections on ping?
I'd assume that a ping func is side effect free.

jensneuse: That said, I like your approach to store the pong time and then check it in ping.
One note, you want to make sure that there's enough time between ping and pong, otherwise we close connections where we weren't able to handle pongs yet.

jensneuse (changes requested):

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go
StarpTech: Yes

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go
StarpTech: Done

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go
StarpTech: That's why I start measuring the time after sending the pong successful or not. Then it depends on how fast we can consume messages from the subgraph if the PONG event arrives within timeout. I think this is safe enough as long as we don't work with a too small timeout.

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -603,10 +634,19 @@ func (c *subscriptionClient) getConnectionInitMessage(ctx context.Context, url s
 }
 
 type ConnectionHandler interface {
+	// StartBlocking starts the connection handler and blocks until the connection is closed
+	// Only used as fallback when epoll is not available
 	StartBlocking() error
+	// HandleMessage handles the incoming message from the connection
 	HandleMessage(data []byte) (done bool)
+	// Ping used to send a ping message to the upstream server to remain the connection alive.
+	// It also keeps track of when the last ping to initiate a shutdown of the dead connection
```
jensneuse:
```suggestion
	// Ping sends a ping message to the upstream server to keep the connection alive.
	// Implementers must keep track of the last ping time to initiate a connection shutdown if the upstream is not sending a pong.
```

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go
```diff
@@ -259,6 +277,47 @@ func (h *gqlTWSConnectionHandler) handleMessageTypePing() {
...
+			h.ServerClose()
+			return
+		}
+	}
+
+	// Only send a new ping if we haven't sent one yet, or if we received a pong for the previous one
+	// (indicated by lastPingSentUnix being 0)
+	if lastPingTimestamp == 0 {
+		_ = h.conn.SetWriteDeadline(time.Now().Add(writeTimeout))
+		err := wsutil.WriteClientText(h.conn, []byte(pingMessage))
```
jensneuse: I think we should be using a mutex instead of atomic because once we enter this "if" branch, we're doing IO, which might be blocking for a while. Until then, it's possible that lastPingTimestamp stays 0 because we're only storing something below (L317). So I think a mutex with a defer unlock prevents this.

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client_test.go
```diff
@@ -1883,3 +1886,457 @@ func TestAsyncSubscribe(t *testing.T) {
...
+		case <-time.After(5 * time.Second):
+			t.Fatal("Timed out waiting for server to complete")
+		}
+
+		clientCancel()
+		serverCancel()
+	})
+}
+
+func TestClientClosesConnectionOnPingTimeout(t *testing.T) {
```
jensneuse, line 2172: Can you the following tests:
1. server sends pong every 100ms, independent of other workloads
2. server sends pong before responding with ack
3. protocol ws example where server sends "ka" events

Side note. How is the "ping" scenario handled with graphql-ws?
There's no ping, just "ka" from server to client (engine is client).
In this case, how do we know if a server is alive?

jensneuse (changes requested):

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
StarpTech: Updated manually

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client_test.go
StarpTech: As discussed, we need to assume that subgraph frameworks have implemented the protocol correctly. Regarding 3), the protocol does not state how keep-alive messages are handled. It is likely that it should be the responsibility of the consumer to decide if the connection is still alive. Implementing this is out of scope for this PR and should be addressed separately.

v2/pkg/engine/datasource/graphql_datasource/graphql_tws_handler.go
StarpTech: Please have another look.

PR #1124, opened 2025-04-14
feat: operation input to MCP compatible json schema converter

StarpTech:
This pull request introduces a new `jsonschema` package to handle JSON Schema definitions in the `v2/pkg/engine/jsonschema/schema.go` file. The conversion logic was optimized for the use case to produce JSON schemas that are compatible with LLM (Large Language Models). This includes removing support for `$def` and working with a fixed recursion depth.

The key changes include defining a `JsonSchema` struct, custom JSON serialization, and various helper functions to create and manipulate schema objects.

Key changes:

### JSON Schema Definition

* Introduced the `JsonSchema` struct to represent a JSON Schema definition, including fields for core schema properties, array-specific fields, enum values, and additional validation.

### Custom Serialization

* Added a `MarshalJSON` method to the `JsonSchema` struct to customize JSON serialization and omit empty fields.

### Helper Functions

* Created helper functions to generate new schemas for different types (`NewObjectSchema`, `NewArraySchema`, `NewStringSchema`, `NewIntegerSchema`, `NewNumberSchema`, `NewBooleanSchema`, `NewEnumSchema`).
* Added methods to the `JsonSchema` struct to add descriptions, default values, formats, and nullable flags (`WithDescription`, `WithDefault`, `WithFormat`, `WithNullable`).

### Cloning Functionality

* Implemented a `CloneSchema` function to create deep copies of existing schema objects.


v2/pkg/engine/jsonschema/schema.go
```diff
@@ -0,0 +1,250 @@
...
+	Required             []string               `json:"required,omitempty"`
+	AdditionalProperties *bool                  `json:"additionalProperties,omitempty"`
+	Description          string                 `json:"description,omitempty"`
+	Nullable             bool                   `json:"nullable,omitempty"`
+
+	// Array-specific fields
+	Items *JsonSchema `json:"items,omitempty"`
+
+	// Enum values
+	Enum []interface{} `json:"enum,omitempty"`
```
jensneuse: can this be array of string?

v2/pkg/engine/jsonschema/schema.go
```diff
@@ -0,0 +1,250 @@
...
+	// Number-specific fields
+	Minimum *float64 `json:"minimum,omitempty"`
+	Maximum *float64 `json:"maximum,omitempty"`
+
+	// Additional validation
+	Pattern string `json:"pattern,omitempty"`
+}
+
+// MarshalJSON customizes JSON serialization to omit empty fields
+func (s *JsonSchema) MarshalJSON() ([]byte, error) {
```
jensneuse, line 51: You could also achieve this by making things a pointer and then have "omitempty"

v2/pkg/engine/jsonschema/schema.go
```diff
@@ -0,0 +1,250 @@
...
+func NewEnumSchema(values []interface{}) *JsonSchema {
+	return &JsonSchema{
+		Type:     TypeString,
+		Enum:     values,
+		Nullable: true, // Default to nullable
+	}
+}
+
+// CloneSchema creates a deep copy of a schema
+func CloneSchema(schema *JsonSchema) *JsonSchema {
```
jensneuse: Is this used somewhere or just generated because we can? :D

v2/pkg/engine/jsonschema/variables_schema.go
```diff
@@ -0,0 +1,461 @@
...
+}
+
+// Build traverses the operation and builds a unified JSON schema for its variables
+func (v *VariablesSchemaBuilder) Build() (*JsonSchema, error) {
+	v.schema = NewObjectSchema()
+	v.recursionTracker = make(map[string]int) // Reset recursion tracker for each build
+
+	// Extract descriptions from root fields
+	var descriptions []string
+	operationDefinition := v.operationDocument.OperationDefinitions[0]
```
jensneuse: What if there is no operation definition?

I recommend to look through RootNodes of operationDocument to check if there is a document. How will you deal with documents containing multiple operations? fail?

v2/pkg/engine/jsonschema/variables_schema.go
```diff
@@ -0,0 +1,461 @@
...
+			selection := v.operationDocument.Selections[selectionRef]
+			if selection.Kind == ast.SelectionKindField {
+				fieldName := v.operationDocument.FieldNameString(selection.Ref)
+
+				// Look up field in schema definition to get description
+				operationType := operationDefinition.OperationType
+				var rootTypeName string
+
+				// Determine root type based on operation type
+				switch operationType {
```
jensneuse, line 72: This is unnecessary to do it while iterating selection refs

v2/pkg/engine/jsonschema/variables_schema.go
```diff
@@ -0,0 +1,461 @@
...
+
+	switch node.Kind {
+	case ast.NodeKindEnumTypeDefinition:
+		return v.processEnumType(node)
+
+	case ast.NodeKindInputObjectTypeDefinition:
+		return v.processInputObjectType(node)
+
+	case ast.NodeKindScalarTypeDefinition:
+		return NewStringSchema()
```
jensneuse: Custom Scalar should be any, not string. I think any in JSON Schema can be represented as `{}` but I'm not 100% sure.

v2/pkg/engine/jsonschema/variables_schema.go
```diff
@@ -0,0 +1,461 @@
...
+
+	default:
+		// If we can't determine the type, default to object
+		return NewObjectSchema()
```
jensneuse: will this lead to a good outcome?

v2/pkg/engine/jsonschema/variables_schema_test.go
```diff
@@ -0,0 +1,1266 @@
...
+		data, err := json.MarshalIndent(schema, "", "  ")
+		require.NoError(t, err)
+
+		// Verify schema structure
+		var parsed map[string]interface{}
+		err = json.Unmarshal(data, &parsed)
+		require.NoError(t, err)
+
+		// Verify top-level structure
+		assert.Equal(t, "object", parsed["type"])
```
jensneuse: Can you rewrite these tests to assert against a fully generated JSON Schema that's pretty printed?
I know that LLMs prefer the testing style you're using here, but it's hard to read and follow the structure.

jensneuse (changes requested):

v2/pkg/engine/jsonschema/schema.go
StarpTech: Good catch

v2/pkg/engine/jsonschema/schema.go
StarpTech: leftover

v2/pkg/engine/jsonschema/schema.go
StarpTech: This leads to a bad api. I'm going to keep it

v2/pkg/engine/jsonschema/variables_schema.go
StarpTech: Adressed

v2/pkg/engine/jsonschema/variables_schema.go
StarpTech: Good one. Empty object can be used for it.

v2/pkg/engine/jsonschema/variables_schema.go
StarpTech: This is equivalent of any. Best we can do.

v2/pkg/engine/jsonschema/variables_schema_test.go
StarpTech: I agree. I'm going to come up with a better approach either inlining or fixture testing.

v2/pkg/engine/jsonschema/variables_schema.go
StarpTech: This is necessary to collect the descriptions of the root fields and add it to the root "description" field of the json schema. The input to json schema conversion is done in `VisitVariableDefinition`

v2/pkg/engine/jsonschema/variables_schema.go
```diff
@@ -0,0 +1,492 @@
...
+}
+
+// EnterDocument implements the astvisitor.EnterDocumentVisitor interface
+func (v *VariablesSchemaBuilder) EnterDocument(operation, definition *ast.Document) {
+	v.schema = NewObjectSchema()
+	v.recursionTracker = make(map[string]int) // Reset recursion tracker for each build
+
+	// Extract descriptions from root fields
+	var descriptions []string
+	if len(operation.OperationDefinitions) > 0 {
```
Noroth: invert if statement to reduce indentation and improve readability

```suggestion
	if len(operation.OperationDefinitions) == 0 {
	    return
	}
```

v2/pkg/engine/jsonschema/variables_schema.go
```diff
@@ -0,0 +1,492 @@
...
+		operationDefinition := operation.OperationDefinitions[0]
+
+		// Process SelectionSet to extract field descriptions
+		if operationDefinition.HasSelections {
```
Noroth: Can also be inverted

v2/pkg/engine/jsonschema/variables_schema.go
```diff
@@ -0,0 +1,492 @@
...
+	}
+
+	return nil
+}
+
+// processTypeByName processes a type by its name, looking it up in the definition document
+func (v *VariablesSchemaBuilder) processTypeByName(typeName string) *JsonSchema {
+	// Handle built-in scalars
+	switch typeName {
+	case "String":
```
Noroth:
```suggestion
	case "ID", "String":
```

Noroth (commented):
small nits

v2/pkg/engine/jsonschema/variables_schema.go
```diff
@@ -0,0 +1,492 @@
...
+import (
+	"fmt"
+
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/ast"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/astvisitor"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/operationreport"
+)
+
+// VariablesSchemaBuilder creates a unified JSON schema for the variables of a GraphQL operation
+type VariablesSchemaBuilder struct {
```
jensneuse, line 12: It would be much simpler if you used the astvisitor.Walker for visiting the ast. It's ok to reinvent the wheel here but could lead to errors. In addition, we're now creating multiple patterns to do the same thing.
It's ok to keep it but I would then refactor it in a second step.

v2/pkg/engine/jsonschema/variables_schema.go
```diff
@@ -0,0 +1,492 @@
...
+			if currentDepth > v.maxRecursionDepth {
+				return nil
+			}
+		} else {
+			// First time seeing this type
+			v.recursionTracker[typeName] = 1
+		}
+
+		// Defer the cleanup of the recursion tracker
+		defer func() {
```
jensneuse: I feel like it's a code smell that you have to use defer here.

v2/pkg/engine/jsonschema/variables_schema_test.go
```diff
@@ -0,0 +1,1570 @@
...
+        },
+        "employmentStatus": {
+          "type": "string",
+          "enum": [
+            "FULL_TIME",
+            "PART_TIME",
+            "CONTRACTOR",
+            "INTERN"
+          ],
+          "nullable": true
```
jensneuse, line 104: are you sure this exists in the spec?

jensneuse (changes requested):

v2/pkg/engine/jsonschema/variables_schema_test.go
StarpTech: It is not supported by everyone but e.g. by gemini, claude

v2/pkg/engine/jsonschema/variables_schema.go
StarpTech: I'd keep it

v2/pkg/engine/jsonschema/variables_schema.go
StarpTech: Done

v2/pkg/engine/jsonschema/variables_schema.go
StarpTech: Done

v2/pkg/engine/jsonschema/variables_schema.go
StarpTech: This is what you are looking for right?

https://github.com/wundergraph/graphql-go-tools/pull/1124/files#diff-6b02bebe51b76c44fcecd0646c473fc2ce20b7601fc5ec3f52cdc6980d83bab6R167-R177

Noroth (approved):

jensneuse (approved):

PR #1126, opened 2025-04-21
feat: skip printing extensions with override list

SkArchon:
# Summary

This PR is used to allow skipping printing certain extensions in the extensions json node in the response. We add any missing flags that hide extensions which includes

* authorization


v2/pkg/engine/resolve/resolvable.go
```diff
@@ -388,18 +388,19 @@ func (r *Resolvable) printValueCompletionExtension() error {
 }
 
 func (r *Resolvable) hasExtensions() bool {
-	if r.ctx.authorizer != nil && r.ctx.authorizer.HasResponseExtensionData(r.ctx) {
+	if !r.ctx.AuthorizerOptions.SkipPrintExtension && r.ctx.authorizer != nil && r.ctx.authorizer.HasResponseExtensionData(r.ctx) {
 		return true
 	}
-	if r.ctx.RateLimitOptions.Enable && r.ctx.RateLimitOptions.IncludeStatsInResponseExtension && r.ctx.rateLimiter != nil {
+	if !r.ctx.RateLimitOptions.SkipPrintExtension && r.ctx.RateLimitOptions.Enable && r.ctx.RateLimitOptions.IncludeStatsInResponseExtension && r.ctx.rateLimiter != nil {
```
SkArchon: @StarpTech fyi Ive gone with the approach where there are two flags for this and RateLimiting, by doing this there will be a central place in the config just for extensions.

I chose not to merge the flags internally to make it obvious that there are two flags for this. Lmk wdyt.

v2/pkg/engine/resolve/context.go
```diff
@@ -14,18 +14,19 @@ import (
...
-	InitialPayload   []byte
-	Extensions       []byte
-	LoaderHooks      LoaderHooks
+	ctx               context.Context
+	Variables         *astjson.Value
+	Files             []*httpclient.FileUpload
+	Request           Request
+	RenameTypeNames   []RenameTypeName
+	RemapVariables    map[string]string
+	AuthorizerOptions AuthorizerOptions
```
SkArchon, line 23: This line was only added

PR #1132, opened 2025-04-24
feat: improve error handling

StarpTech:
This PR cleaned up error handling. Before, we joined errors twice, which resulted in duplicated error messages for subgraph access logs. I renamed `subgraphErrors` to `error` because it never only represented subgraph logs. I also wrapped the http client error with stack information for more verbose logs.

### DIFF
--- a/v2/pkg/engine/resolve/resolvable.go
+++ b/v2/pkg/engine/resolve/resolvable.go
@@ -703,7 +703,7 @@ func (r *Resolvable) addRejectFieldError(reason string, ds DataSourceInfo, field
 	} else {
 		errorMessage = fmt.Sprintf("Unauthorized to load field '%s', Reason: %s.", fieldPath, reason)
 	}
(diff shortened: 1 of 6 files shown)


v2/pkg/engine/resolve/resolvable.go
```diff
@@ -703,7 +703,7 @@ func (r *Resolvable) addRejectFieldError(reason string, ds DataSourceInfo, field
 	} else {
 		errorMessage = fmt.Sprintf("Unauthorized to load field '%s', Reason: %s.", fieldPath, reason)
 	}
-	r.ctx.appendSubgraphError(goerrors.Join(errors.New(errorMessage),
+	r.ctx.appendError(goerrors.Join(errors.New(errorMessage),
```
endigma, line 706: `appendError` calls `errors.Join` inside, which takes a variadic second parameter, perhaps we could just pass this through to avoid nesting `errors.Join` calls and this odd `goerrors` alias?

PR #1141, opened 2025-05-05
fix: update pubsub datasource to use JSON variable renderer

alepane21:
This PR solves an issue with mutations that had a variable with a value not valid in plain JSON in a document. So something like this worked:
```graphql
updateEmployeeMyNats(id: 123, update:{
    id:"realComplexId"
  }) {
    success
  }
```
because the message published was like this:
```json
{"id": 123, "update": {"id": "test"}}
```

But something like this would have failed:
```graphql
updateEmployeeMyNats(id: "realComplexId", update:{
    id:"realComplexId"
  }) {
    success
  }
```
because the message produced was like this:
```json
{"id": realComplexId, "update": {"id": "test"}}
```
and this is not a valid JSON.

With this PR, the first published message will still be like before, and the second one will be properly escaped:
```json
{"id": "realComplexId", "update": {"id": "test"}}
```

This change should be backward compatible for most of the users. The only exception is if someone relied on the old formatting behaviour (like if the id was including the escaping quotes).


jensneuse (changes requested):
Can you add tests that verify the change? I assume that there's no test coverage for this change as you didn't have to update any test fixtures.

alepane21:
> Can you add tests that verify the change? I assume that there's no test coverage for this change as you didn't have to update any test fixtures.

I had to change the Response plan to match the change of renderer, but there is no test that verify the output on the engine. There are some on the router, but I can think of good tests to add here too, so I'll work on that.

Thanks for your feedback!

devsergiy (approved):

PR #1147, opened 2025-05-08
fix: print indent once per level by default

### LINKED ISSUE #405: astprinter: PrintIndent applies indentation twice

ysmolski:
For v1 and v2 this fixes the default behaviour of the PrintIndent and
PrintStringIndent functions of astprinter. Instead of printing indent
twice for every depth level, it prints it once per level.

The fix does not alter output for other parts, golden tests
are intact. Tests were modified by passing correctly sized indent.

Fixes #405

### DIFF
--- a/pkg/ast/ast_description.go
+++ b/pkg/ast/ast_description.go
@@ -45,7 +45,6 @@ func (d *Document) PrintDescription(description Description, indent []byte, dept
 				skippedWhitespace += 1
 				continue
 			case runes.SPACE:
-				skippedWhitespace += 0.5
 				continue
 			}
 		}
--- a/pkg/astprinter/astprinter.go
+++ b/pkg/astprinter/astprinter.go
@@ -18,6 +18,7 @@ func Print(document, definition *ast.Document, out io.Writer) error {
 }
 
 // PrintIndent is the same as Print but accepts an additional indent parameter to set indentation.
+// Indent is written once for every depth level.
 func PrintIndent(document, definition *ast.Document, indent []byte, out io.Writer) error {
 	printer := Printer{
 		indent: indent,
@@ -93,12 +94,12 @@ func (p *printVisitor) indentationDepth() (depth int) {
 	case ast.NodeKindOperationDefinition,
 		ast.NodeKindFragmentDefinition:
 	default:
-		return 2
+		return 1
 	}
 
 	for i := range p.Ancestors {
 		if p.Ancestors[i].Kind == ast.NodeKindSelectionSet {
-			depth += 2
+			depth += 1
 		}
 	}
 
--- a/execution/graphql/normalization_test.go
+++ b/execution/graphql/normalization_test.go
@@ -48,7 +48,7 @@ func TestRequest_Normalize(t *testing.T) {
         name
     }
 }`
-		op, _ := astprinter.PrintStringIndent(&request.document, "  ")
+		op, _ := astprinter.PrintStringIndent(&request.document, "    ")
 		assert.Equal(t, normalizedOperation, op)
 	})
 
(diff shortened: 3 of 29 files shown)


devsergiy:
Hi @ysmolski 

thanks for the PR

v1 considered as unmaintained
We would appreciate it if you could port these changes to v2

Thanks
Wundergraph Team!

pkg/ast/ast_description.go
```diff
@@ -45,7 +45,6 @@ func (d *Document) PrintDescription(description Description, indent []byte, dept
 				skippedWhitespace += 1
 				continue
 			case runes.SPACE:
-				skippedWhitespace += 0.5
```
devsergiy, line 48: does it have any consequences for the printing of multiline descriptions?

pkg/astprinter/astprinter.go
```diff
@@ -12,12 +12,14 @@ import (
 
 // Print takes a document as well as a definition (optional) and prints it to the io.Writer.
 // The definition is only necessary in case a GraphQL Operation should be printed.
+// Print uses four spaces as indentation.
```
devsergiy: Does the comment reflect the behavior? Print should by default print unintended document

pkg/federation/schema_test.go
```diff
@@ -10,6 +11,7 @@ import (
 func TestSchemaBuilder_BuildFederationSchema(t *testing.T) {
 	actual, err := BuildFederationSchema(baseSchema, serviceSDL)
 	assert.NoError(t, err)
+	fmt.Println(actual)
```
devsergiy: debug code?

devsergiy (changes requested):

pkg/ast/ast_description.go
ysmolski: If I keep this line it breaks the following golden test. It tries to add two unneeded spaces. I might look into this whole function with skipping whitespace, though. I have not spotted any test with tabs too...
```
--- FAIL: TestSchemaBuilder_BuildFederationSchema (0.00s)
    schema_test.go:13: Result did not match the golden fixture. Diff is below:

        --- Expected
        +++ Actual
        @@ -73,4 +73,4 @@
             Explains why this element was deprecated, usually also including a suggestion
        -    for how to access supported similar data. Formatted in
        -    [Markdown](https://daringfireball.net/projects/markdown/).
        +      for how to access supported similar data. Formatted in
        +      [Markdown](https://daringfireball.net/projects/markdown/).
             """
```

pkg/astprinter/astprinter.go
ysmolski: Oh. My bad. Removing this, print does not use indent at all!

pkg/federation/schema_test.go
ysmolski: Yes, removing it now

ysmolski:
> thanks for the PR
> v1 considered as unmaintained We would appreciate it if you could port these changes to v2

All fixes are done. v2 was patched as well.

devsergiy (approved):

PR #1158, opened 2025-05-23
feat: communicate downstream errors more effectively

endigma:
Adds differentiation between Complete/Done and Close/Shutdown events, so that downstream subgraph closure may be communicated more effectively to the downstream client.

### DIFF
--- a/v2/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
+++ b/v2/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
@@ -71,19 +71,19 @@ func TestGraphQLDataSource(t *testing.T) {
 			type Query {
 				a: A
 			}
-	
+
 			type A {
 				a: String
 			}
-	
+
 			type B {
 				b: String
 			}
-	
+
 			union U = A | B
 		`
 
-		t.Run("run", RunTest(definition, ` 
+		t.Run("run", RunTest(definition, `
 			query MyQuery{
 				a {
 					... on U {
@@ -780,20 +780,20 @@ func TestGraphQLDataSource(t *testing.T) {
 			type Query {
 			  thing: Thing
 			}
-			
+
 			type Thing {
 			  id: String!
 			  abstractThing: AbstractThing
 			}
-			
+
 			interface AbstractThing {
 			  name: String
 			}
-			
+
 			type ConcreteOne implements AbstractThing {
 			  name: String
 			}
-			
+
 			type ConcreteTwo implements AbstractThing {
 			  name: String
 			}`
@@ -2140,9 +2140,9 @@ func TestGraphQLDataSource(t *testing.T) {
 			_entities(representations: $representations){
 				... on Product {
 					reviews {
-						body 
+						body
 						author {
-							username 
+							username
 							id
 						}
 					}
@@ -4128,7 +4128,7 @@ func TestGraphQLDataSource(t *testing.T) {
 								author {
 									id
 									username
-								}	
+								}
 								product {
 									name
 									price
@@ -6107,11 +6107,11 @@ func TestGraphQLDataSource(t *testing.T) {
 								me: User
 								self: Identity
 							}
-							
+
 							interface Identity {
 								id: ID!
 							}
-							
+
 							type User implements Identity @key(fields: "id") {
 								id: ID!
 								username: String!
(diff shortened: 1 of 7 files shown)


v2/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
```diff
@@ -780,20 +780,20 @@ func TestGraphQLDataSource(t *testing.T) {
...
 			  id: String!
 			  abstractThing: AbstractThing
 			}
-			
+
 			interface AbstractThing {
 			  name: String
 			}
-			
+
```
devsergiy, line 792: rather aggressive formatter rule - to update line endings in multiline string literals

v2/pkg/engine/resolve/resolve.go
```diff
@@ -358,6 +364,15 @@ func (s *sub) complete() {
 	s.writer.Complete()
 }
 
+func (s *sub) close() {
+	// The channel is used to communicate that the subscription is done
```
devsergiy, line 372: subscription which was closed by server?

v2/pkg/engine/resolve/resolve.go
```diff
@@ -531,7 +546,11 @@ func (r *Resolver) handleTriggerInitialized(triggerID uint64) {
 }
 
 func (r *Resolver) handleTriggerDone(triggerID uint64) {
-	r.shutdownTrigger(triggerID)
+	if r.options.Debug {
+		fmt.Printf("resolver:trigger:done:%d\n", triggerID)
```
devsergiy, line 554: should it be a logger print?

v2/pkg/engine/resolve/resolve.go
endigma: What is the question? It is called on Close with no Complete path, which is almost certainly downstream server crash/close

devsergiy: question is about clarifying comment with info when it happens

endigma: Oh, sure

endigma: Done

v2/pkg/engine/resolve/resolve.go
endigma: Don't think so, all its siblings use this type of print wrapped with `if r.options.Debug`.

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
endigma: AFAIK this was just gofmt, it shouldn't cause any issues, no?

devsergiy: gofmt is not doing that
should not be a problem

devsergiy (approved):

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
endigma: perhaps zed bug, it's trailing whitespacespace on a newline, maybe it's somehow unaware its a string literal

v2/pkg/engine/resolve/resolve.go
```diff
@@ -750,17 +763,38 @@ func (r *Resolver) shutdownTrigger(id uint64) {
...
-		fmt.Printf("resolver:trigger:done:%d\n", trig.id)
+		fmt.Printf("resolver:trigger:complete:%d\n", id)
 	}
 
+	trig, ok := r.triggers[id]
+	if !ok {
+		return
+	}
+
+	removed := r.shutdownTriggerSubscriptions(id, true, nil)
```
StarpTech, line 801: Most of the code here is duplicated from `shutdownTrigger`. Should we unify it?

StarpTech (changes requested):

v2/pkg/engine/resolve/resolve.go
StarpTech: Done

StarpTech (approved):
LGTM

v2/pkg/engine/resolve/resolve.go
```diff
@@ -275,6 +275,14 @@ type trigger struct {
 	initialized bool
 }
 
+// work is used to encapsulate a function that needs to be
+// executed in the worker goroutine. fn will be executed, and if
+// final is true the worker will be stopped after fn is executed.
+type work struct {
```
Noroth: Nit: I would call it workItem

PR #1161, opened 2025-05-28
feat: use new negation regex matching for matching connections

SkArchon:
In the router we use negation in regex expression matching via a negation flag, since we also use this logic when creating connections, and we need to assign connections correctly, we apply this logic in the engine.

**Tests**
Add unit tests for checking the hash. There used to be tests that tests if a connection is reused, but it has been removed with certain changes. I would need more time to go through the engine to figure out how to restructure the tests (and why they were removed in the first place) https://github.com/wundergraph/graphql-go-tools/commit/3ec0ebef5a8ab6133b411ab009fe8da289701c5f#diff-2a8da55a93918e23bbe398ea85eb27a43292dd9cc3090eea26264094dc28931bL355-L365

### DIFF
--- a/v2/pkg/engine/datasource/graphql_datasource/configuration.go
+++ b/v2/pkg/engine/datasource/graphql_datasource/configuration.go
@@ -3,13 +3,14 @@ package graphql_datasource
 import (
 	"errors"
 	"fmt"
+	"net/http"
+
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/ast"
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/astparser"
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/asttransform"
 	grpcdatasource "github.com/wundergraph/graphql-go-tools/v2/pkg/engine/datasource/grpc_datasource"
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/federation"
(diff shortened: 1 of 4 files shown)


v2/pkg/engine/datasource/graphql_datasource/configuration.go
```diff
@@ -3,15 +3,13 @@ package graphql_datasource
...
-	"net/http"
-	"regexp"
-
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/ast"
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/astparser"
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/asttransform"
 	grpcdatasource "github.com/wundergraph/graphql-go-tools/v2/pkg/engine/datasource/grpc_datasource"
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/federation"
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/operationreport"
+	"net/http"
```
devsergiy: please order imports

devsergiy (approved):

v2/pkg/engine/datasource/graphql_datasource/configuration.go
SkArchon: done

PR #1164, opened 2025-06-03
feat: expose datasource metadata on interface

mattjohnsonpint:
`NewDataSourceConfiguration` requires passing a `DataSourceMetadata` object, and it's assigned to the `DataSourceMetadata` field of the `dataSourceConfiguration[T]` struct.  But it's not able to be retrieved after that, because the `DataSourceConfiguration[T]` interface doesn't expose it.

I have a few scenarios where the metadata would be useful, but it's locked away.  This PR adds a `Metadata()` func to the interface to expose it, and uses it in a test.

Thanks.


mattjohnsonpint:
That will work.  Thank you!

PR #1165, opened 2025-06-03
fix: usage of unions fragments in union/object selection set

devsergiy:


v2/pkg/ast/ast_union_type_definition.go
```diff
@@ -121,3 +121,25 @@ func (d *Document) UnionTypeDefinitionMemberTypeNames(ref int) (typeNames []stri
...
 }
+
+func (d *Document) UnionTypeDefinitionMemberTypeNamesAsBytes(ref int) (typeNames [][]byte, ok bool) {
+	if !d.UnionTypeDefinitions[ref].HasUnionMemberTypes {
+		return nil, false
+	}
+
+	typeNames = make([][]byte, 0, len(d.UnionTypeDefinitions[ref].UnionMemberTypes.Refs))
+	for _, typeRef := range d.UnionTypeDefinitions[ref].UnionMemberTypes.Refs {
+		typeNames = append(typeNames, d.TypeNameBytes(typeRef))
```
jensneuse, line 132: can use index

v2/pkg/ast/ast_node.go
```diff
@@ -570,6 +570,23 @@ func (d *Document) UnionNodeIntersectsInterfaceNode(unionNode, interfaceNode Nod
 	return false
 }
 
+func (d *Document) UnionNodeIntersectsUnionNode(parentUnionNode, nestedUnionNode Node) bool {
+	for _, i := range d.UnionTypeDefinitions[parentUnionNode.Ref].UnionMemberTypes.Refs {
```
jensneuse, line 574: could you not just get member names from both and compare if one member is in both?

jensneuse (approved):

v2/pkg/ast/ast_node.go
devsergiy: not sure that it will make a huge difference

v2/pkg/ast/ast_union_type_definition.go
devsergiy: will it be better?

Aenimus: marginally, probably

PR #1166, opened 2025-06-03
feat: add custom field renderer to resolvable

jensneuse:


v2/pkg/engine/resolve/resolvable.go
```diff
@@ -823,7 +861,7 @@ func (r *Resolvable) walkNull() bool {
 func (r *Resolvable) walkStaticString(str *StaticString) bool {
```
devsergiy: static string used internally for rendering static __typename's

don't need to have a render check and could be harmful to have it

v2/pkg/engine/resolve/resolvable.go
```diff
@@ -447,6 +449,41 @@ func (r *Resolvable) printNode(value *astjson.Value) {
...
+	}
+	r.marshalBuf = value.MarshalTo(r.marshalBuf[:0])
+	r.renderFieldBytes(r.marshalBuf, nullable)
+}
+
+func (r *Resolvable) renderFieldBytes(data []byte, nullable bool) {
+	if r.printErr != nil {
+		return
+	}
+	// if we render a variable that's actually a node, we don't have a context
```
devsergiy, line 464: if it applies only to scalar types

should we name it like `renderScalarFieldBytes`?

v2/pkg/engine/resolve/resolvable.go
```diff
@@ -447,6 +449,41 @@ func (r *Resolvable) printNode(value *astjson.Value) {
...
+		value := FieldValue{
+			Name:       r.currentFieldInfo.Name,
+			Type:       r.currentFieldInfo.NamedType,
+			ParentType: r.currentFieldInfo.ExactParentTypeName,
+			IsNullable: nullable,
+			Path:       r.renderFieldPath(),
+			Data:       data,
+		}
+		if len(r.path) > 0 {
+			value.IsList = r.path[len(r.path)-1].Name == ""
```
devsergiy: Did I get it right - it is the ListItem? not the value is a list?

v2/pkg/engine/resolve/resolvable_custom_field_renderer_test.go
```diff
@@ -0,0 +1,936 @@
...
+	resolveFunc func(ctx *Context, value []byte) ([]byte, error)
+}
+
+func (t *testCustomResolve) Resolve(ctx *Context, value []byte) ([]byte, error) {
+	return t.resolveFunc(ctx, value)
+}
+
+func TestResolvable_CustomFieldRenderer(t *testing.T) {
+	t.Parallel()
+	t.Run("string nullable", func(t *testing.T) {
```
devsergiy: should this be a table test? or some test with the runner function? a lot of repetitiveness

v2/pkg/engine/resolve/resolvable.go
```diff
@@ -912,7 +950,7 @@ func (r *Resolvable) walkInteger(i *Integer, value *astjson.Value) bool {
 		return r.err()
 	}
 	if r.print {
-		r.printNode(value)
+		r.renderFieldAstValue(value, i.Nullable)
```
devsergiy: Q: we evaluate nullability and produce errors before we doing a print

if someone returns null from the custom renderer for the not nullable field - it potentially could break some clients?

v2/pkg/engine/resolve/resolvable.go
jensneuse: if last item name is empty string, that means it's a path item with an index for an array, so it's a list

v2/pkg/engine/resolve/resolvable.go
jensneuse: we pass nullable or non nullable info to renderer, it's their responsibility to honour this

v2/pkg/engine/resolve/resolvable.go
devsergiy: i mean that we will pass a single list item to renderer - so should it be named IsListItem instead of isList, cause value itself is not a list

v2/pkg/engine/resolve/resolvable_custom_field_renderer_test.go
```diff
@@ -34,903 +34,530 @@ func (t *testCustomResolve) Resolve(ctx *Context, value []byte) ([]byte, error)
 	return t.resolveFunc(ctx, value)
 }
 
+type fieldValueTestCase struct {
+	name                 string
+	input                string
+	fieldValue           Node
+	fieldInfo            *FieldInfo
+	expectedOutput       string
+	expectedFieldName    string
```
devsergiy: you could use expectedFieldValue and just compare whole struct instead of separate fields

devsergiy (approved):

PR #1189, opened 2025-06-18
fix: don't panic when calling QueryPlan on FetchTreeNode if `includeQueryPlans` is false

endigma:
This currently panics because https://github.com/wundergraph/graphql-go-tools/blob/e75a1dd24d5255b6cc990269c5c7922f851f4fc1/v2/pkg/engine/resolve/fetchtree.go#L212-L213 is a nil pointer dereference when 

This change makes it just fail empty instead of panic, which is fine for what it's needed for in the router.

I originally made this not nil at the source in https://github.com/wundergraph/graphql-go-tools/blob/dd0d9cc2a9192799c4b3c5ecaa89134c03e01801/v2/pkg/engine/plan/visitor.go#L1259-L1261
but it causes like 1000+ inline snapshot tests to fail which are hardcoded and don't have `-update`, so I catch the nilness at usage site instead


StarpTech (approved):
LGTM

devsergiy (approved):

PR #1223, opened 2025-07-10
fix: use NodeFragmentIsAllowedOnNode to check fragmentSpread

ysmolski:
This fixes following false positive error in the "interfaces" benchmark:

    fragment spread: fragment catFrag must be spread on type Cat and not type CatOrDog

Previously rule fragment check was trying to verify that the catFrag fragment (of a Cat type) matches completely the name of EnclosingType (CatOrDog). Instead I propose to use the "NodeFragmentIsAllowedOnNode" method to check if fragment can be applied to the enclosingType that can be union (our case).

Benchmark query:

    query conflictingDifferingResponses {
	    catOrDog {
		    ...catDogFrag
	    }
    }
    fragment catDogFrag on CatOrDog {
	    ...catFrag
	    ...dogFrag
    }
    fragment catFrag on Cat {
	    someValue: meowVolume
    }
    fragment dogFrag on Dog {
	    someValue: barkVolume
    }

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of fragment spreads to provide clearer error messages for unknown or invalid fragment types.
  * Enhanced error reporting in benchmark tests to log errors without stopping execution, allowing all issues to be reported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ The following benchmark measures the "overhead" of loading and resolving a Graph
 In more complete end-to-end benchmarks, we've measured up to 8x more requests per second and 8x lower p99 latency compared to Apollo Router, which is written in Rust.
 
 ```shell
-cd v2/pkg/engine
+cd v2/pkg/engine/resolve
 go test -run=nothing -bench=Benchmark_NestedBatchingWithoutChecks -memprofile memprofile.out -benchtime 3s && go tool pprof memprofile.out
 goos: darwin
 goarch: arm64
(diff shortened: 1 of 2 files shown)


jensneuse:
@ysmolski does it make sense to add any additional tests to verify that the code change is correct, or is this sufficiently covered by existing tests? I would typically expect that a code change requires changes to tests.

ysmolski:
> @ysmolski does it make sense to add any additional tests to verify that the code change is correct, or is this sufficiently covered by existing tests? I would typically expect that a code change requires changes to tests.

Yes, you are right. I have already found some tests not checking the actual error returned and fixed a bug. But I will push it to the separate PR because it is complicated issue by itself.

ysmolski:
Related issue that adds specific errors to the tests and fixes a bug: https://github.com/wundergraph/graphql-go-tools/pull/1235/

PR #1235, opened 2025-07-16
fix: check that object claims to implement interface

ysmolski:
Fragment valdidation layer calls the
"NodeFragmentIsAllowedOnInterfaceTypeDefinition" function to validate that fragment node is allowed in the interface. But it was not checking that the node actually claimed to implement the interface, only that the node contained all the fields from interface.

I have added this additional check by calling to the "NodeImplementsInterface" function.

Also tests were made more strict by adding specific errors to the invalid cases.


v2/pkg/ast/ast_node.go
```diff
@@ -500,14 +504,15 @@ func (d *Document) NodeUnionMemberRefs(node Node) (refs []int) {
 
 // Node fragments
 
-func (d *Document) NodeFragmentIsAllowedOnNode(fragmentNode, onNode Node) bool {
-	switch onNode.Kind {
+// NodeFragmentIsAllowedOnNode determines if a fragment node is valid on a target node.
+func (d *Document) NodeFragmentIsAllowedOnNode(fragment, target Node) bool {
```
ysmolski: I was thinking that the "Node" suffix is redundant in parameters. This is how it will look after simplifying them as an attempt to improve readability of these functions.

devsergiy: I think it will be better to rename `target` to `parent`

v2/pkg/ast/ast_node.go
```diff
@@ -472,7 +476,7 @@ func (d *Document) RemoveNodeFromSelectionSet(set int, node Node) (removed bool)
 
 // NodeInterfaceRefs returns the interfaces implemented by the given node (this is
 // only applicable to object kinds).
-// Returns nil if node kind is not an object kind.
+// Returns nil if the node is not an object type.
```
devsergiy: `object kind` a bit better

v2/pkg/ast/ast_node.go
```diff
@@ -486,7 +490,7 @@ func (d *Document) NodeInterfaceRefs(node Node) (refs []int) {
 
 // NodeUnionMemberRefs returns the union members of the given node (this is only
 // applicable to union kinds).
-// Returns nil if node kind is not an object kind.
+// Returns nil if the node is not an object type.
```
devsergiy: is not a union kind
old description was wrong

v2/pkg/ast/ast_node.go
```diff
@@ -516,15 +521,19 @@ func (d *Document) NodeFragmentIsAllowedOnNode(fragmentNode, onNode Node) bool {
 func (d *Document) NodeFragmentIsAllowedOnInterfaceTypeDefinition(fragmentNode, interfaceTypeNode Node) bool {
 	switch fragmentNode.Kind {
 	case NodeKindObjectTypeDefinition:
-		return d.NodeImplementsInterfaceFields(fragmentNode, interfaceTypeNode)
+		interfaceName := d.InterfaceTypeDefinitionNameBytes(interfaceTypeNode.Ref)
+		claimsImplementation := d.NodeImplementsInterface(fragmentNode, interfaceName)
```
devsergiy: I would suggest adding a helper `NodeImplementsInterfaceNode`

devsergiy: and it is enough to check only that, we should remove usage of implements interface fields completely in following PR

v2/pkg/ast/ast_node.go
```diff
@@ -516,15 +521,19 @@ func (d *Document) NodeFragmentIsAllowedOnNode(fragmentNode, onNode Node) bool {
...
+		return claimsImplementation && d.NodeImplementsInterfaceFields(fragmentNode, interfaceTypeNode)
 	case NodeKindInterfaceTypeDefinition:
-		return bytes.Equal(d.InterfaceTypeDefinitionNameBytes(fragmentNode.Ref), d.InterfaceTypeDefinitionNameBytes(interfaceTypeNode.Ref)) ||
+		fragmentName := d.InterfaceTypeDefinitionNameBytes(fragmentNode.Ref)
+		interfaceName := d.InterfaceTypeDefinitionNameBytes(interfaceTypeNode.Ref)
+		return bytes.Equal(fragmentName, interfaceName) ||
 			d.InterfaceNodeIntersectsInterfaceNode(fragmentNode, interfaceTypeNode)
```
devsergiy: we could add `bytes.Equal(fragmentName, interfaceName)` right into method `d.InterfaceNodeIntersectsInterfaceNode`

v2/pkg/ast/ast_node.go
```diff
@@ -535,25 +544,28 @@ func (d *Document) NodeFragmentIsAllowedOnUnionTypeDefinition(fragmentNode, unio
...
 
 func (d *Document) NodeFragmentIsAllowedOnObjectTypeDefinition(fragmentNode, objectTypeNode Node) bool {
 	switch fragmentNode.Kind {
 	case NodeKindObjectTypeDefinition:
 		return bytes.Equal(d.ObjectTypeDefinitionNameBytes(fragmentNode.Ref), d.ObjectTypeDefinitionNameBytes(objectTypeNode.Ref))
 	case NodeKindInterfaceTypeDefinition:
-		return d.NodeImplementsInterfaceFields(objectTypeNode, fragmentNode)
+		interfaceName := d.InterfaceTypeDefinitionNameBytes(fragmentNode.Ref)
+		claimsImplementation := d.NodeImplementsInterface(objectTypeNode, interfaceName)
+		return claimsImplementation && d.NodeImplementsInterfaceFields(objectTypeNode, fragmentNode)
```
devsergiy: It is enough to have here only `NodeImplementsInterfaceNode` check

v2/pkg/ast/ast_node.go
```diff
@@ -563,7 +575,8 @@ func (d *Document) UnionNodeIntersectsInterfaceNode(unionNode, interfaceNode Nod
 		if node.Kind != NodeKindObjectTypeDefinition {
 			continue
 		}
-		if d.NodeImplementsInterfaceFields(node, interfaceNode) {
+		claimsImplementation := d.NodeImplementsInterface(node, interfaceName)
```
devsergiy: we need only single check here `d.NodeImplementsInterface`

devsergiy (dismissed):

v2/pkg/ast/ast_node.go
ysmolski: Yes, it will be.

v2/pkg/ast/ast_node.go
ysmolski: Thanks!

v2/pkg/ast/ast_node.go
ysmolski: Ok, fixing it

v2/pkg/ast/ast_node.go
ysmolski: Alright, fixing all such places

v2/pkg/ast/ast_node.go
ysmolski: agree. fixed

v2/pkg/ast/ast_node.go
ysmolski: done

v2/pkg/ast/ast_node.go
ysmolski: done

ysmolski:
Also I have secretly renamed the parameters in those checks functions. 😊 Could not resist

devsergiy (approved):

PR #1238, opened 2025-07-17
fix: test v2 benchmarks on ci

ysmolski:
Include benchmarks into quick tests of v2.

In addition, I have attempted to simplify Makefile rules for testing. Instead of aliasing let's use "test" and "test-quick" rules directly from github workflows.

@coderabbitai summary

## Checklist

### DIFF
--- a/v2/Makefile
+++ b/v2/Makefile
@@ -1,6 +1,6 @@
 .PHONY: test
 test:
-	go test -race ./...
+	go test -race -count=1 ./...
 
 test-fresh: clean-testcache test
 
@@ -9,11 +9,7 @@ test-stability:
 
 .PHONY: test-quick
 test-quick:
-	go test -count=1 ./...
-
-.PHONY: test
-test-no-race:
-	go test -race ./...
+	go test -count=1 -run=. -bench=. -benchtime=1x ./...
 
 clean-testcache:
 	go clean -testcache
@@ -29,12 +25,6 @@ format:
 
 .PHONY: prepare-merge
 prepare-merge: format test
-
-.PHONY: ci
-ci: test
-
-.PHONY: ci-quick
-ci-full: test-quick
 
 .PHONY: generate
 generate: $(GOPATH)/bin/go-enum $(GOPATH)/bin/mockgen $(GOPATH)/bin/stringer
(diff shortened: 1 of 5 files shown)


v2/pkg/netpoll/fd_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build linux || darwin || netbsd || freebsd || openbsd || dragonfly
```
devsergiy: maybe ! windows?

v2/Makefile
```diff
@@ -1,6 +1,6 @@
 .PHONY: test
 test:
-	go test -race ./...
+	go test -race -count=1 ./...
```
devsergiy: we do not need count=1 here

v2 is updated regularly, so we could run tests in parallel
execution is rarely updated, so it is required there, because otherwise test results will be cached, because the execution directory is not changed, but we are running the execution tests on each v2 update

v2/Makefile
```diff
@@ -9,11 +9,7 @@ test-stability:
 
 .PHONY: test-quick
 test-quick:
-	go test -count=1 ./...
-
-.PHONY: test
-test-no-race:
-	go test -race ./...
+	go test -count=1 -run=. -bench=. -benchtime=1x ./...
```
devsergiy: same here

v2/pkg/netpoll/fd_test.go
ysmolski: Yep, should work here.

v2/Makefile
ysmolski: Right, thanks for explaining this!

v2/Makefile
ysmolski: Ok.. 

But I am not 100% satisfied how tests run and look with benchmarks being tested. Specifically, test cache is not workingm and noisy output from `v2/pkg/engine/datasource`. Looking into it...

ysmolski: I think I found a solution to this problem. See the update Makefile.

devsergiy: do you still want to cleanup duplicated aliases?

ysmolski: Yeah I prefer because it already lead to some inconsistencies, but can revert back if you prefer the old names. I might not know the history and all the use cases. Just let me know!

devsergiy (approved):

PR #1240, opened 2025-07-18
fix: merge infline fragment and field selections together

ysmolski:
This change fixes normalization for cases like the following:

    query fragmentsWithFields {
	    dog {
		    extras {
			    ... on DogExtra {
				    string
			    }
		    }
		    extras {
			    ... on DogExtra {
				    string
				    string1
			    }
		    }
	    }
    }

I have marged two different visitors and their tests into a single visitor. EnterSelectionSets from two stages were merged into one, but with the following differences:

    * Internal loop was optimized,
    * Early return happens only when something is merged.

@coderabbitai summary

## Checklist


v2/pkg/astnormalization/inline_fragment_selection_merging.go
```diff
@@ -54,33 +56,88 @@ func (f *inlineFragmentSelectionMergeVisitor) mergeInlineFragments(left, right i
 	return true
 }
 
+func (f *inlineFragmentSelectionMergeVisitor) fieldsCanMerge(left, right int) bool {
+	leftName := f.operation.FieldNameBytes(left)
+	rightName := f.operation.FieldNameBytes(right)
+	leftAlias := f.operation.FieldAliasBytes(left)
+	rightAlias := f.operation.FieldAliasBytes(right)
+
+	if !bytes.Equal(leftName, rightName) || !bytes.Equal(leftAlias, rightAlias) {
```
jensneuse, line 65: Can we add a func to ast Document that takes left + right field ref and compares them character by character by using len comparison and index access of input raw bytes? I think it would be more efficient than pulling out 4 byte slices and using bytes.Equal

devsergiy: do you have benchmark? 😊

I believe bytes.Equal doing exactly what have you described

jensneuse: You have to copy bytes around a lot which could be avoided.

devsergiy: why copy? it is still just a pointer to doc.Input.RawBytes

ysmolski: Just for my own education, I followed what happens here. No allocation is spotted, just pointers in the form of slices are returned and handled in two comparisons. 

The only possible issue/tradeoff I see that data locality (cpu cache friendliness) is not so good because those slices could point to distant parts of `doc.Input.RawBytes`. But this issue cannot be easily solved in place and is a question of general design.

devsergiy: It is best what we could get now, we allocate document body once and then use only parts of it without allocations

in this case we do not care so much about memory alignment

PR #1243, opened 2025-07-18
feat: add subscriptions hook

alepane21:
The demo_mode will serve a demo execution config if is not found elsewhere (controlplane, static execution config).
In this PR the flag is improved so that, when the demo_mode is enabled, the router can start also if no graph token is set.

## Summary by CodeRabbit

* **New Features**
  * Added pre-subscription startup hooks for subscriptions to emit initial events or abort startup.
  * Enhanced subscription context with a callable update emitter so hooks can push updates into the subscription stream.

* **Tests**
  * Added tests validating hook invocation, error propagation from hooks, and update emission behavior (including heavy and concurrent emission scenarios).

## Checklist


v2/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -1944,6 +1946,15 @@ func (s *SubscriptionSource) AsyncStart(ctx *resolve.Context, id uint64, input [
 	if options.Body.Query == "" {
 		return resolve.ErrUnableToResolve
 	}
+	for _, fn := range s.onSubscriptionStartFns {
+		events, err := fn(ctx)
+		if err != nil {
+			return err
+		}
+		for _, event := range events {
+			updater.Update(event)
```
jensneuse: Is it the right approach to call the updater.Update func or should we just handle the event directly?

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -1963,6 +1974,15 @@ func (s *SubscriptionSource) Start(ctx *resolve.Context, input []byte, updater r
 	if options.Body.Query == "" {
 		return resolve.ErrUnableToResolve
 	}
+	for _, fn := range s.onSubscriptionStartFns {
```
jensneuse: This code seems to be a duplicate to L1949

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
```diff
@@ -8902,6 +8902,174 @@ func TestSanitizeKey(t *testing.T) {
...
+		return subscriptionSource
+	}
+
+	t.Run("should call the onSubscriptionStartFns at subscription source Start", func(t *testing.T) {
+		ctx := resolve.NewContext(context.Background())
+		defer ctx.Context().Done()
+
+		updater := &testSubscriptionUpdater{}
+
+		startFnCalled := make(chan *resolve.Context, 1)
```
jensneuse: I'd generally use empty struct for this (struct{})

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
```diff
@@ -8902,6 +8902,174 @@ func TestSanitizeKey(t *testing.T) {
...
+	t.Run("if the onSubscriptionStartFns returns a message, the message should be sent to the client", func(t *testing.T) {
+		ctx := resolve.NewContext(context.Background())
+		defer ctx.Context().Done()
+
+		updater := &testSubscriptionUpdater{}
+
+		source := newSubscriptionSource(ctx.Context())
+		source.onSubscriptionStartFns = []OnSubscriptionStartFn{
+			func(ctx *resolve.Context) ([][]byte, error) {
+				return [][]byte{[]byte(`{"data":{"messageAdded":{"text":"hello world!","createdBy":"injection"}}}`)}, nil
```
jensneuse: - what behaviour do we expect when an "errors" field is in the response?
- should the event be wrapped in "data" or should the event be in the root?

v2/pkg/engine/datasource/graphql_datasource/configuration.go
```diff
@@ -103,6 +104,11 @@ type SingleTypeField struct {
 	FieldName string
 }
 
+// OnSubscriptionStartFn defines a hook function that is called when a subscription starts.
+// It receives the resolve context and can return initial events or an error.
+// If an error is returned, the subscription will not start.
+type OnSubscriptionStartFn func(ctx *resolve.Context) ([][]byte, error)
```
jensneuse: Would it be simpler to reason about this if we add a func on resolve.Context to "emit" events instead of returning a slice of slice of bytes here?

e.g. ctx.EmitEvent([]byte(`{}`))

jensneuse (changes requested):

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
alepane21: I changed the behaviour. Now is calling ctx.EmitSubscriptionUpdate.

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
alepane21: the test is not using this anymore

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
alepane21: I think that the data field should behave exactly like when it arrives from the subgraph. But than we also allow to return an error, to add an easy way to make an error be raised. This also allows router to have custom errors that can specify the kind of error.

v2/pkg/engine/datasource/graphql_datasource/configuration.go
alepane21: I've added the EmitEvent on the ctx!

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
alepane21: There is still some lines duplicated, but those are about error and flow management, and I think they are better positioned outside of the helper method.

alepane21:
@coderabbitai full review

v2/pkg/engine/resolve/datasource.go
```diff
@@ -26,3 +26,16 @@ type AsyncSubscriptionDataSource interface {
...
+// It is used to call a function when a subscription is started.
+// This is useful for data sources that need to do some work when a subscription is started,
+// e.g. to establish a connection to the data source or to emit updates to the client.
+// The function is called with the context and the input of the subscription.
+// The function is called before the subscription is started and can be used to emit updates to the client.
+type SubscriptionDataSourceHookable interface {
+	// OnSubscriptionStart is called when a new subscription is created
+	// If close is true, the subscription is closed.
+	// If an error is returned, the error is propagated to the client.
+	OnSubscriptionStart(ctx *Context, input []byte) (close bool, err error)
```
StarpTech: This interface doesn't reflect the ADR.

v2/pkg/engine/resolve/resolve.go
```diff
@@ -590,6 +600,12 @@ func (r *Resolver) handleAddSubscription(triggerID uint64, add *addSubscription)
 		s.heartbeat = true
 	}
 
+	// Set the subscription updater to the resolver to allow external code to emit updates
+	// on this subscription
+	add.ctx.SetSubscriptionUpdater(func(data []byte) {
```
StarpTech: This doesn't look right. Events must be emitted through the events channel.  This can cause cause issues because the writer is not synchronised.

StarpTech (changes requested):

v2/pkg/engine/resolve/datasource.go
alepane21: I fixed the name to be better aligned with the interface described in the ADR. Still it will not be like what the ADR described, because this is the interface on a datasource, but the ADR describe the custom module interface that is in the router (https://github.com/wundergraph/cosmo/pull/2059/, still working on it)

v2/pkg/engine/resolve/resolve.go
alepane21: Ok, so now this has been changed. I'm running the hooks inside the worker to avoid slowing the main loop to much. The updates are then sent directly using the executeSubscriptionUpdate, without using the events (that would have been complex, because events are going to be sent to all the subscriptions of a trigger, but I want them to be sent only to the current subscription) or even the workChan, to avoid to block the subscription worker if a modules return more messages than the buffer size.

v2/pkg/engine/resolve/resolve_test.go
```diff
@@ -5434,6 +5443,185 @@ func TestResolver_ResolveGraphQLSubscription(t *testing.T) {
...
+				messagesDroppedFromHook.Add(1)
+			}
+
+			// start a go routine to send the updates after the source started emitting messages
+			go func() {
+				// Wait for the first message to arrive before sending updates
+				select {
+				case <-firstMessageArrived:
+					for i := 1; i < int(messagesToSendFromHook); i++ {
+						done := ctx.TryEmitSubscriptionUpdate([]byte(fmt.Sprintf(`{"data":{"counter":%d}}`, i+20000)))
```
jensneuse: I'd rename done to "accepted", just to be more clear, but it's ok.

jensneuse (approved):
Overall looks good to me. Update @StarpTech on why we changed the API a bit.

v2/pkg/engine/datasource/graphql_datasource/configuration.go
```diff
@@ -119,6 +127,7 @@ type SubscriptionConfiguration struct {
 	// these headers by itself.
 	ForwardedClientHeaderRegularExpressions []RegularExpression
 	WsSubProtocol                           string
+	SubscriptionOnStartFns                  []SubscriptionOnStartFn
```
Noroth: nit:
You are already in `SubscriptionConfiguration` so you are repeating the `Subscription` name and you can give the field a more precise name

```go
	StartupHooks                  []SubscriptionOnStartFn
```

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -1989,3 +1991,13 @@ func (s *SubscriptionSource) UniqueRequestID(ctx *resolve.Context, input []byte,
 	}
 	return s.client.UniqueRequestID(ctx, options, xxh)
 }
+
+func (s *SubscriptionSource) SubscriptionOnStart(ctx *resolve.Context, input []byte) (close bool, err error) {
```
Noroth: nit: Maybe `OnSubscriptionStart` sounds a bit better? What do you think?

v2/pkg/engine/resolve/resolve.go
```diff
@@ -757,7 +822,42 @@ func (r *Resolver) handleRemoveClient(id int64) {
...
+	}
+	skip, err := sub.resolve.Filter.SkipEvent(ctx, data, r.triggerUpdateBuf)
+	if err != nil {
+		r.asyncErrorWriter.WriteError(ctx, err, sub.resolve.Response, sub.writer)
+		return true
+	}
+	if skip {
+		return true
+	}
+	return false
```
Noroth:
```suggestion
	return skip
```

v2/pkg/engine/resolve/resolve.go
```diff
@@ -569,6 +570,56 @@ func (r *Resolver) handleTriggerComplete(triggerID uint64) {
...
+		return
+	}
+
+	close, errStartHook := hook.SubscriptionOnStart(initialHookCtx, add.input)
+	if errStartHook != nil {
+		r.asyncErrorWriter.WriteError(initialHookCtx, errStartHook, add.resolve.Response, add.writer)
+	}
+	if close {
+		r.closeTrigger(triggerID, SubscriptionCloseKindNormal)
+		return
```
Noroth: I think this return can be removed

v2/pkg/engine/datasource/graphql_datasource/configuration.go
alepane21: I like the name, thanks!

v2/pkg/engine/resolve/resolve.go
alepane21: Fixed thanks ... I think it was inside a loop before the last refactor 

Doh

v2/pkg/engine/resolve/resolve.go
alepane21: Fixed!

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
alepane21: This is the same name of the hooks, and they are based on a standard decided for the new modules https://github.com/wundergraph/cosmo/pull/2059/files#diff-bd315de2a568701e2cf2a0874fe961bfd9cc53d99e4c6677c4466e58baa010d8R83, so I think it is going to be better if we leave it as it is.

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -1992,3 +1994,13 @@ func (s *SubscriptionSource) UniqueRequestID(ctx *resolve.Context, input []byte,
...
+	for _, fn := range s.subscriptionOnStartFns {
+		close, err = fn(ctx, input)
```
StarpTech: What's the motivation to provide a separate way to close the subscription? I think any error should result in not starting the subscription.

v2/pkg/engine/resolve/resolve.go
```diff
@@ -766,33 +874,13 @@ func (r *Resolver) handleTriggerUpdate(id uint64, data []byte) {
 		fmt.Printf("resolver:trigger:update:%d\n", id)
 	}
 
-	for c, s := range trig.subscriptions {
```
StarpTech, line 781: Could you please reiterate why this refactor / change is necessary for the hook?

v2/pkg/engine/resolve/resolve.go
```diff
@@ -590,6 +633,28 @@ func (r *Resolver) handleAddSubscription(triggerID uint64, add *addSubscription)
 		s.heartbeat = true
 	}
 
+	// Clone the context to avoid modifying the original context.
+	// This is important because the original context is changed in the parent thread
+	// and could cause data races.
+	initialHookCtx := add.ctx.clone(add.ctx.ctx)
+
+	// Send handleAddSubscriptionInitialHook as the first work sent to the worker to process the hooks
```
StarpTech: I think we're on the wrong path. This is a lot of refactoring and coordination for adding the start hook.

StarpTech (changes requested):

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
alepane21: I removed the close option, explained the initial idea here: https://github.com/wundergraph/cosmo/pull/2059#discussion_r2284533391

v2/pkg/engine/resolve/resolve.go
alepane21: We needed an update to be run only on a single subscription (so that the writemessage would write only to a single subscription!), and not on all the subscription associated to a trigger. To avoid copying-pasting the code I extracted the code in a different method.

v2/pkg/engine/resolve/resolve.go
alepane21: Changed as discussed! Let me know if you have other ideas, thanks!

v2/pkg/engine/resolve/resolve.go
alepane21: Change removed after last refactor.

alepane21:
@coderabbitai full review

dkorittki:
Closed, as we do the actual review when merging branch `topic/streams-v1` into `master` branch

PR #1260, opened 2025-08-01
fix: generate query plans for subscriptions

ysmolski:
WIP

@coderabbitai summary

## Checklist


v2/pkg/engine/resolve/fetchtree_test.go
```diff
@@ -0,0 +1,142 @@
...
+		queryPlan := fetches.QueryPlan()
+		actual := queryPlan.PrettyPrint()
+
+		expected := `
+QueryPlan {
+  Subscription {
+    Primary: {
+      Fetch(service: "country") {
+        {
+            countryUpdated {
```
jensneuse, line 114: Overall looks good, I'm just wondering if we should write this as `subscription countryUpdated`.
Just having the root field could be interpreted as a query.
I was also thinking if it would make sense to prefix mutations with `mutation`.
Query can be just a selection set because it's kind of well known that just having a selection set without query, mutation or subscription is understood as a query.

jensneuse (approved):
I'm approving because overall looks good. Just a question if we should render subscription root fields differently or not, which I'm not sure about. I'm curious what @devsergiy thinks.

v2/pkg/engine/resolve/fetchtree_test.go
ysmolski: In this case, surrounding block is `Subscription {}`, it can be concluded that it is the subscription and not the query. But for other cases like mutation vs query it is not conclusive just from the query plan. Maybe @devsergiy has some ideas how we can streamline query plans.

PR #1261, opened 2025-08-05
fix: add an option to ignore skip/include directives

ysmolski:
During normalization this option enforces directiveIncludeSkipVisitor to keep nodes and delete directives instead.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable option to control whether nodes affected by `@skip` and `@include` directives are preserved or removed during operation normalization.
* **Tests**
  * Introduced new test cases to verify the behavior of preserving nodes when processing `@skip` and `@include` directives.

## Checklist

### DIFF
--- a/v2/pkg/astnormalization/directive_include_skip.go
+++ b/v2/pkg/astnormalization/directive_include_skip.go
@@ -73,7 +73,7 @@ func (d *directiveIncludeSkipVisitor) handleSkip(ref int) {
 		return
 	}
 	if !d.keepNodes && bool(skip) {
-		d.handleRemoveNode()
+		d.removeParentNode()
 	} else {
 		d.operation.RemoveDirectiveFromNode(d.Ancestors[len(d.Ancestors)-1], ref)
 	}
@@ -104,7 +104,7 @@ func (d *directiveIncludeSkipVisitor) handleInclude(ref int) {
 	if d.keepNodes || bool(include) {
 		d.operation.RemoveDirectiveFromNode(d.Ancestors[len(d.Ancestors)-1], ref)
 	} else {
-		d.handleRemoveNode()
+		d.removeParentNode()
 	}
 }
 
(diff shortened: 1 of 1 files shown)


v2/pkg/astnormalization/directive_include_skip.go
```diff
@@ -62,7 +76,7 @@ func (d *directiveIncludeSkipVisitor) handleSkip(ref int) {
 	default:
 		return
 	}
-	if skip {
+	if !d.keepNodes && bool(skip) {
```
devsergiy, line 75: maybe just add this check into handleRemoveNode?

v2/pkg/astnormalization/directive_include_skip.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/lexer/literal"
 )
 
+// directiveIncludeSkip registers a visitor to handle @include and @skip directives.
+// It deletes nodes that are evaluated as unused by the directives.
```
devsergiy, line 14: does it make sense to reuse `directiveIncludeSkipKeepNodes` here with a bool param?

```
directiveIncludeSkip {
  directiveIncludeSkipKeepNodes(false)
}
```
}

v2/pkg/astnormalization/astnormalization.go
```diff
@@ -209,7 +216,11 @@ func (o *OperationNormalizer) setupOperationWalkers() {
 
 	directivesIncludeSkip := astvisitor.NewWalker(8)
 	preventFragmentCycles(&directivesIncludeSkip)
-	directiveIncludeSkip(&directivesIncludeSkip)
+	if o.options.ignoreSkipInclude {
+		directiveIncludeSkipKeepNodes(&directivesIncludeSkip)
+	} else {
+		directiveIncludeSkip(&directivesIncludeSkip)
```
devsergiy: we could use single register function here if it will have a bool param

```
directiveIncludeSkipKeepNodes(&walker, o.options.ignoreSkipInclude)
```
```

File: v2/pkg/astnormalization/directive_include_skip.go (hunk 1 above)
Comment (ysmolski) [reply v2/pkg/astnormalization/directive_include_skip.go:75]: It will not work there because we want to delete directives too.

File: v2/pkg/astnormalization/directive_include_skip.go (hunk 2 above)
Comment (ysmolski) [reply v2/pkg/astnormalization/directive_include_skip.go:14]: I could try that, but tests would not work since they accept this kind of function:
```
type registerNormalizeFunc func(walker *astvisitor.Walker)
```
I have noticed that we do not use options with normalizers and validators anywhere...

v2/pkg/astnormalization/astnormalization.go
ysmolski: I have tried something similar to what you had in mind.

v2/pkg/astnormalization/directive_include_skip.go
ysmolski: Ok, fixed it.

PR #1269, opened 2025-08-12
feat: improved subscription heartbeats

endigma:
## Summary by CodeRabbit

* **New Features**
  * Added explicit subscription heartbeat signaling support to response writers.

* **Bug Fixes**
  * Heartbeats now stop promptly on client disconnect or resolver shutdown, preventing stray heartbeat errors.

* **Performance**
  * Heartbeat emission streamlined to a single writer call, reducing overhead and improving idle stability.

* **Refactor**
  * Consolidated and renamed subscription heartbeat configuration for clearer setup and consistent behavior.

* **Tests**
  * Subscription tests updated to record and verify heartbeat events; test writers now support heartbeat signaling.

## Checklist

### DIFF
--- a/v2/pkg/engine/resolve/resolve.go
+++ b/v2/pkg/engine/resolve/resolve.go
@@ -139,8 +139,8 @@ type ResolverOptions struct {
 	ResolvableOptions ResolvableOptions
 	// AllowedCustomSubgraphErrorFields defines which fields are allowed in the subgraph error when in passthrough mode
 	AllowedSubgraphErrorFields []string
-	// SubHeartbeatInterval defines the interval in which a heartbeat is sent to all subscriptions (whether or not this does anything is determined by the subscription response writer)
-	SubHeartbeatInterval time.Duration
+	// SubscriptionHeartbeatInterval defines the interval in which a heartbeat is sent to all subscriptions (whether or not this does anything is determined by the subscription response writer)
+	SubscriptionHeartbeatInterval time.Duration
 	// MaxSubscriptionFetchTimeout defines the maximum time a subscription fetch can take before it is considered timed out
 	MaxSubscriptionFetchTimeout time.Duration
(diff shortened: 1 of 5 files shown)


v2/pkg/engine/resolve/resolve.go
```diff
@@ -518,24 +514,16 @@ func (r *Resolver) handleHeartbeat(sub *sub, data []byte) {
 		fmt.Printf("resolver:heartbeat:subscription:%d\n", sub.id.SubscriptionID)
 	}
 
-	if _, err := sub.writer.Write(data); err != nil {
```
StarpTech, line 521: I like the approach to move it to the router 👍

v2/pkg/engine/resolve/resolve.go
```diff
@@ -518,24 +514,16 @@ func (r *Resolver) handleHeartbeat(sub *sub, data []byte) {
...
-			return
-		}
-		r.asyncErrorWriter.WriteError(sub.ctx, err, nil, sub.writer)
-	}
-	err := sub.writer.Flush()
-	if err != nil {
-		// If flush fails (e.g. client disconnected), remove the subscription.
+	if err := sub.writer.Heartbeat(); err != nil {
+		// If heartbeat fails (e.g. client disconnected), remove the subscription.
 		_ = r.AsyncUnsubscribeSubscription(sub.id)
```
StarpTech, line 519: The question here is, should we really abort the whole subscription for a heartbeat write error?

endigma: This just retains the error from before, a write error is almost always EOF or context cancelled or etc, meaning the client is gone

StarpTech: Before, we unsubscribed only on client cancellation or flush error.

endigma: I think this is an acceptable simplification, if writing or flushing the heartbeat fails I can't imagine many cases where we wouldn't want to fail the subscription.

PR #1276, opened 2025-08-14
fix: always close response body

ysmolski:
Close properly body in the "newWSConnectionHandler" and "dial" functions.

Bonus part:
* Fix leak in subscriptionClient.dial/newWSConnectionHandler.
* Rename local "connectionInitMessage" var to avoid collision with global var.
* Fixed 1st flaky test in TestClientToSubgraphPingPong and improved runtime of both tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket connection cleanup and upgrade handling to prevent resource leaks, drain upgrade responses for connection reuse, and make subscription initialization/acknowledgement more robust.

* **Tests**
  * Added parameterized tests for many WebSocket upgrade failure scenarios and invalid accept headers.
  * Strengthened tests to explicitly close HTTP responses, validate ping/pong and payload timing, and tighten unsubscribe/server-completion checks.

### DIFF
--- a/v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client_test.go
+++ b/v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client_test.go
@@ -1894,14 +1894,15 @@ func TestClientToSubgraphPingPong(t *testing.T) {
 	if runtime.GOOS == "windows" {
 		t.Skip("Skipping test on Windows as it's not reliable")
 	}
+	t.Parallel()
 
 	t.Run("client sends ping message after configured interval", func(t *testing.T) {
 		t.Parallel()
 
-		// Create channel to signal server done
-		serverDone := make(chan struct{})
-		pingReceived := make(chan struct{}) // closed when the server receives a ping
-		payloadSend := make(chan struct{})  // closed when the server sends a payload
+		serverDone := make(chan struct{}) // to signal server done
+		// buffered channels and non-blocking send to avoid double-close panics if events repeat
+		pingReceived := make(chan struct{}, 1) // signaled when the server receives a ping
+		payloadSend := make(chan struct{}, 1)  // signaled when the server sends a payload
 
 		// Create test server that will handle the WebSocket connection
 		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
@@ -1962,12 +1963,18 @@ func TestClientToSubgraphPingPong(t *testing.T) {
 				switch messageStr {
 				case `{"type":"ping"}`:
 					receivedPing = true
-					close(pingReceived)
+					select {
+					case pingReceived <- struct{}{}:
+					default:
+					}
 					err = conn.Write(r.Context(), websocket.MessageText, []byte(`{"type":"pong"}`))
 					assert.NoError(t, err)
 					err = conn.Write(r.Context(), websocket.MessageText, []byte(`{"id":"1","type":"next","payload":{"data":{"messageAdded":{"text":"after ping-pong"}}}}`))
 					assert.NoError(t, err)
-					close(payloadSend)
+					select {
+					case payloadSend <- struct{}{}:
+					default:
+					}
 				case `{"id":"1","type":"complete"}`:
 					receivedComplete = true
 				}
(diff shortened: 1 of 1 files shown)


v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client_test.go
```diff
@@ -2012,6 +2013,8 @@ func TestClientToSubgraphPingPong(t *testing.T) {
 		select {
 		case <-pingReceived:
 			t.Log("Ping received successfully")
+			// don't unsubscribe immediately, give the server time to send a payload
+			time.Sleep(50 * time.Millisecond)
```
endigma: If we want to wait on this, is it possible to add a third channel "payloadReceived" so this isn't timing sensitive?

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client_test.go
```diff
@@ -2342,6 +2345,218 @@ func TestClientClosesConnectionOnPingTimeout(t *testing.T) {
...
+				var upgradeErr *UpgradeRequestError
+				require.ErrorAs(t, err, &upgradeErr)
+				require.Equal(t, tc.statusCode, upgradeErr.StatusCode)
+				require.Equal(t, server.URL, upgradeErr.URL)
+			} else {
+				assert.NoError(t, err, "Expected no error for status code %d", tc.statusCode)
+			}
+
+			// Give some time for any async cleanup
+			time.Sleep(50 * time.Millisecond)
```
endigma: Any way we can wait on this with a channel and timeout instead of sleeping?

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -486,6 +486,8 @@ func (u *UpgradeRequestError) Error() string {
 }
 
 func (c *subscriptionClient) newWSConnectionHandler(requestContext, engineContext context.Context, options GraphQLSubscriptionOptions, updater resolve.SubscriptionUpdater) (*connection, error) {
+	// Any failure will be stored here, needed for defer().
```
endigma:
```suggestion
	// Any failure will be stored here, needed for deferred body closer.
```

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -496,35 +498,43 @@ func (c *subscriptionClient) newWSConnectionHandler(requestContext, engineContex
...
+		}
+	}()
+
+	initMsg, err := c.getConnectionInitMessage(requestContext, options.URL, options.Header)
 	if err != nil {
 		return nil, err
 	}
 
 	if len(options.InitialPayload) > 0 {
-		connectionInitMessage, err = jsonparser.Set(connectionInitMessage, options.InitialPayload, "payload")
```
endigma, line 505: Does this need to be renamed? It's fine, but it seems unnecessary

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -580,9 +590,16 @@ func (c *subscriptionClient) dial(ctx context.Context, options GraphQLSubscripti
 
 	upgradeResponse, err := c.httpClient.Do(req)
 	if err != nil {
+		// GotConn is not called on error.
```
endigma: What is the significance of this comment?

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
ysmolski: There is a global variable named `connectionInitMessage` that was shadowed in this function.

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
ysmolski: Applied. Thanks

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
ysmolski: I think it can be removed, left over from my investigations. Removing

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
endigma: Ah

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client_test.go
ysmolski: Good idea, I am going to fix the other tests that has 2 seconds wait time

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client_test.go
ysmolski: ah. removing that since we are not doing leaks detection here.

PR #1282, opened 2025-08-21
feat: add support of field selection reasons extensions

ysmolski:
The PropagateFetchReasons option of Resolver enables router to send to subgraphs an extension that includes reasons why fields where included in the request. 

This data is sent only for the fields marked with `@openfed__requireFetchReason` in the schema.
Directive is implemented in the router repo.

Router part: https://github.com/wundergraph/cosmo/pull/2160

## Summary by CodeRabbit

* **New Features**
  * Opt-in per-fetch "fetch reasons" metadata (which fields were requested, why, and by which subgraphs) with public accessors and a toggle to propagate this metadata to upstream subgraphs.

* **Improvements**
  * More deterministic planning via stabilized ordering and inverse-dependency tracking.
  * Indexing recognizes fields that carry fetch-reason semantics.
  * Loader can inject fetch-reasons into upstream requests when enabled.

* **Tests**
  * New and updated tests validating collection and propagation of fetch-reason metadata.

* **Documentation**
  * Clarified planner configuration comments.

* **Chores**
  * Minor benchmark and formatting tweaks.

## Checklist


v2/pkg/engine/resolve/fetch.go
```diff
@@ -387,6 +413,17 @@ type FetchDependencyOrigin struct {
 	IsRequires bool `json:"isRequires"`
 }
 
+// RequestedField explains who requested a specific (typeName, fieldName) combination.
+// A field can be requested by the user and/or by one or more subgraphs, with optional reasons.
+type RequestedField struct {
+	TypeName         string   `json:"__typename"`
+	FieldName        string   `json:"field"`
```
ysmolski: not sure what is better: `__typename` or `typeName` as the json key.

v2/pkg/engine/plan/visitor.go
```diff
@@ -46,6 +47,7 @@ type Visitor struct {
 	skipFieldsRefs               []int
 	fieldRefDependsOnFieldRefs   map[int][]int
 	fieldDependencyKind          map[fieldDependencyKey]fieldDependencyKind
+	fieldRefAllowsFieldRefs      map[int][]int // inverse of fieldRefDependsOnFieldRefs
```
devsergiy: maybe `fieldRefDependencyForFieldRefs`?

v2/pkg/engine/plan/visitor.go
```diff
@@ -54,13 +56,13 @@ type Visitor struct {
 	indirectInterfaceFields      map[int]indirectInterfaceField
 	pathCache                    map[astvisitor.VisitorKind]map[int]string
 
-	// plannerFields tells which fields are planned on which planners
-	// map plannerID -> []fieldRef
+	// plannerFields maps plannerID to fieldRefs planned on this planner.
```
devsergiy, line 59: I think it is better to keep an explicit explanation of what the key and the value are in a map

devsergiy: Even better will be explanation like this
```
map[plannerID][]fieldRef
```

v2/pkg/engine/plan/visitor.go
```diff
@@ -632,7 +634,10 @@ func (v *Visitor) LeaveField(ref int) {
 	}
 }
 
+// skipField returns true if the field is a dependency and should not be included in the response.
```
devsergiy: Maybe like this?
```
// skipField returns true if the field was added by the query planner as a dependency 
// for another field and should not be included in the response. 
// If it returns false, the field was requested by the client.
```

v2/pkg/engine/plan/visitor.go
```diff
@@ -632,7 +634,10 @@ func (v *Visitor) LeaveField(ref int) {
...
+// Consequently, if we don't skip the field, the client requested it in the query.
 func (v *Visitor) skipField(ref int) bool {
+	// If this grows, switch to map[int]struct{} for O(1).
```
devsergiy: should it be a TODO?

v2/pkg/engine/resolve/fetch.go
```diff
@@ -387,6 +418,17 @@ type FetchDependencyOrigin struct {
 	IsRequires bool `json:"isRequires"`
 }
 
+// FetchReason explains who requested a specific (typeName, fieldName) combination.
+// A field can be requested by the user and/or by one or more subgraphs, with optional reasons.
+type FetchReason struct {
+	TypeName      string   `json:"typename"`
+	FieldName     string   `json:"field"`
+	FromSubgraphs []string `json:"from_subgraphs,omitempty"`
```
devsergiy: Initially, it was `BySubgraphs`; why is `From` better?

devsergiy: Don't we request from the current subgraph because it was requested by other subgraphs? or maybe `for` subgraphs

v2/pkg/engine/resolve/loader.go
```diff
@@ -1596,6 +1598,23 @@ func (l *Loader) executeSourceLoad(ctx context.Context, fetchItem *FetchItem, so
...
+		fetchReasons := fetchItem.Fetch.FetchReasons()
+		if len(fetchReasons) > 0 {
+			var encoded []byte
+			encoded, res.err = json.Marshal(fetchReasons)
+			if res.err != nil {
+				res.err = errors.WithStack(res.err)
+				return
+			}
+			// We expect that body.extensions is an object
+			input, res.err = jsonparser.Set(input, encoded, "body", "extensions", "fetch_reasons")
```
devsergiy, line 1611: Will it work if extensions are not present yet?

v2/pkg/engine/resolve/resolve.go
```diff
@@ -110,44 +110,63 @@ type ResolverOptions struct {
...
 	SubscriptionHeartbeatInterval time.Duration
+
 	// MaxSubscriptionFetchTimeout defines the maximum time a subscription fetch can take before it is considered timed out
 	MaxSubscriptionFetchTimeout time.Duration
+
 	// ApolloRouterCompatibilitySubrequestHTTPError is a compatibility flag for Apollo Router, it is used to handle HTTP errors in subrequests differently
 	ApolloRouterCompatibilitySubrequestHTTPError bool
+
+	// PropagateFetchReasons enables sending to upstream subgraphs (only) the "fetch_reasons"
+	// extension that explains why each field was requested. This flag does not expose the data to clients.
```
devsergiy: we could add `request extension`

v2/pkg/engine/resolve/loader.go
ysmolski: Yes, it does work. My tests do not have extensions created elsewhere and it works.

v2/pkg/engine/plan/visitor.go
ysmolski: It (key and value explanation) is included into this phrase "plannerFields maps plannerID to fieldRefs". 
In general, "T maps A to B" is equivalent to `T map[A]B`

v2/pkg/engine/resolve/loader.go
ysmolski: In case when it was present, it will merge into the "extensions" object.

v2/pkg/engine/resolve/fetch.go
ysmolski: I am returning "By" prefix

v2/pkg/engine/plan/visitor.go
ysmolski: ack

v2/pkg/engine/plan/visitor.go
ysmolski: Added

v2/pkg/engine/plan/visitor.go
ysmolski: Yeah. It is a tricky stuff. I did not want it to be similar to the other field "fieldRefDependsOnFieldRefs"...

v2/pkg/engine/plan/visitor.go
```diff
@@ -2,6 +2,7 @@ package plan
 
 import (
 	"bytes"
+	"cmp"
```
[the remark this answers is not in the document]
ysmolski: @coderabbitai. It is not an issue. We do not support Go < 1.23 anyway. Do not raise such concerns in the future

v2/pkg/engine/plan/visitor.go
ysmolski: What about `fieldRefDependants`? Dependant is the inverse of "depends on".

v2/pkg/engine/resolve/resolve.go
ysmolski: fixed

v2/pkg/engine/plan/visitor.go
```diff
@@ -1392,3 +1397,103 @@ func (v *Visitor) resolveFetchDependencies(fetchID int) []resolve.FetchDependenc
...
 }
+
+func (v *Visitor) buildFetchReasons(fetchID int) []resolve.FetchReason {
+	fields, ok := v.plannerFields[fetchID]
+	if !ok {
+		return nil
+	}
+	dsConfig := v.planners[fetchID].DataSourceConfiguration()
+
+	type typedField struct {
```
devsergiy, line 1408: you could reuse KeyConditionCoordinate and rename it to FieldCoordinate

v2/pkg/engine/plan/visitor.go
```diff
@@ -1392,3 +1397,103 @@ func (v *Visitor) resolveFetchDependencies(fetchID int) []resolve.FetchDependenc
...
+			continue
+		}
+
+		byUser := !v.skipField(fieldRef)
+		dependants, ok := v.fieldRefDependants[fieldRef]
+
+		var subgraphs []string
+		var isKey, isRequires bool
+
+		if ok {
```
devsergiy, line 1431: should ok be `hasDependants` or `dependants, ok := v.fieldRefDependants[fieldRef]` be moved right before this line?

devsergiy (approved):

v2/pkg/engine/plan/visitor.go
ysmolski: Oh thanks. I have renamed and used it.

v2/pkg/engine/plan/visitor.go
ysmolski: Yep, left over from old revisions. I have moved this assigment right into the if condition.

v2/pkg/engine/plan/visitor.go
ysmolski: I will addres this in the next PR if you don't mind. Chaining PR's in GH is a pain.

v2/pkg/engine/plan/visitor.go
ysmolski: I will addres this in the next PR if you don't mind. Chaining PR's in GH is a pain.

PR #1293, opened 2025-09-08
fix: refactor CoordinateDependencies, FetchReasons

ysmolski:
Move them into the "FetchInfo" struct. That struct is ideal for storing derived data
for a Fetch and passing it to the later stages of the engine.

Add the BuildFetchReasons option to the planner, that allows other parts to enable building the FetchReason list.

## Summary by CodeRabbit

* **New Features**
  * Propagates fetch-reasons to subgraphs; outgoing requests can include extensions.fetch_reasons when enabled.
  * Exposes richer per-fetch and per-field metadata (source, operation type, coordinates, provenance) and a toggle to enable building fetch-reason data.

* **Refactor**
  * Consolidates fetch- and dependency-related metadata into a unified fetch-info model and centralizes fetch-reason tracking and propagation.

* **Tests**
  * Updated test harness and cases to cover fetch-reason generation and propagation.

* **Chores**
  * Simplifies lint/format commands to use default tool configuration.

## Checklist

### DIFF
--- a/v2/pkg/engine/plan/visitor.go
+++ b/v2/pkg/engine/plan/visitor.go
@@ -1345,7 +1345,7 @@ func (v *Visitor) configureFetch(internal *objectFetchConfiguration, external re
 		return singleFetch
 	}
 	singleFetch.Info.FetchReasons = v.buildFetchReasons(internal.fetchID)
-	if singleFetch.Info.FetchReasons == nil {
+	if len(singleFetch.Info.FetchReasons) == 0 {
 		return singleFetch
 	}
 


ysmolski:
@coderabbitai review

v2/pkg/engine/plan/visitor.go
```diff
@@ -1333,11 +1333,23 @@ func (v *Visitor) configureFetch(internal *objectFetchConfiguration, external re
 			OperationType:  internal.operationType,
 			QueryPlan:      external.QueryPlan,
 		}
-	}
-
-	if !v.Config.DisableIncludeFieldDependencies {
-		singleFetch.CoordinateDependencies = v.resolveFetchDependencies(internal.fetchID)
-		singleFetch.FieldFetchReasons = v.buildFetchReasons(internal.fetchID)
+		if !v.Config.DisableIncludeInfo {
```
devsergiy: could we use early return here?

```
if v.Config.DisableIncludeInfo {
  return singleFetch
}
```

devsergiy: and down the line of if nesting

v2/pkg/engine/resolve/fetch.go
```diff
@@ -21,8 +21,7 @@ type Fetch interface {
 	FetchKind() FetchKind
 	Dependencies() *FetchDependencies
 	DataSourceInfo() DataSourceInfo
-	DependenciesCoordinates() []FetchDependency
-	FetchReasons() []FetchReason
+	GetInfo() *FetchInfo
```
devsergiy: could it be `FetchInfo`?

v2/pkg/engine/resolve/fetch.go
```diff
@@ -21,8 +21,7 @@ type Fetch interface {
 	FetchKind() FetchKind
 	Dependencies() *FetchDependencies
 	DataSourceInfo() DataSourceInfo
```
devsergiy: I believe we could remove this function, as it is used in a single plac,e and we could obtain this info directly from FetchInfo

v2/pkg/engine/plan/visitor.go
```diff
@@ -1333,11 +1333,23 @@ func (v *Visitor) configureFetch(internal *objectFetchConfiguration, external re
...
+				singleFetch.Info.FetchReasons = v.buildFetchReasons(internal.fetchID)
+				if singleFetch.Info.FetchReasons != nil {
+					dsConfig := v.planners[internal.fetchID].DataSourceConfiguration()
+					indices := make([]int, 0, len(singleFetch.Info.FetchReasons))
+					for i, fr := range singleFetch.Info.FetchReasons {
+						if dsConfig.RequiresFetchReason(fr.TypeName, fr.FieldName) {
+							indices = append(indices, i)
+						}
+					}
+					singleFetch.Info.RequireFetchReason = indices
```
devsergiy: What if we add an additional property `PropagateFetchReasons` to the Info? should not be a lot of overhead, as we precisely specify which type+field requires it

v2/pkg/engine/resolve/fetch.go
ysmolski: Okay. Renaming

v2/pkg/engine/plan/visitor.go
ysmolski: Yes, it can be done and that field could be omited from JSON using tags. But I will have to do filtering anyway in the loader by copying records with `PropagateFetchReasons` set to true. I wish there was an option to avoid that copying...

v2/pkg/engine/plan/visitor.go
ysmolski: I see what you mean. Replacing nested ifs...

v2/pkg/engine/resolve/loader.go
```diff
@@ -1599,19 +1599,27 @@ func (l *Loader) executeSourceLoad(ctx context.Context, fetchItem *FetchItem, so
...
-				return
+			if len(filteredReasons) > 0 {
+				var encoded []byte
+				encoded, res.err = json.Marshal(filteredReasons)
+				if res.err != nil {
+					res.err = errors.WithStack(res.err)
+					return
+				}
+				// We expect that body.extensions is an object
+				input, res.err = jsonparser.Set(input, encoded, "body", "extensions", "fetch_reasons")
```
jensneuse: I don't want to be annoying but I think this approach is wrong. We have an existing architecture to render the input. Now we add post processing to extend it. I understand that this makes things simpler. However, this is not a universal solution. E.g. if we want to add fetch reasons to gRPC messages, we will now have to parse the input and extract the information again. Wouldn't it be cleaner if we make it part of the input template and just render it into the input? No post processing needed. And if we want to add fetch reasons to gRPC, we can update the gRPC templating pipeline and provide fetch reasons statically typed instead of just in input. my 2 cents

devsergiy: I believe it should not be a part of this issue, as the code for propagating arbitrary request extensions was already there

Could be part of a bigger refactoring of making input an object instead of a string

jensneuse: I'm fine with that. In this case, let's create a ticket for that to not lose track.

v2/pkg/engine/plan/visitor.go
ysmolski: After Slack discussion, I am adding a field that will hold `PropagatedFetchReasons`

v2/pkg/engine/resolve/fetch.go
ysmolski: Yes, removing it

v2/pkg/engine/resolve/loader.go
ysmolski: @jens, I have created corresponding issue, please mark as resolved :)

execution/Makefile
```diff
@@ -15,10 +15,10 @@ update-fixtures:
 
 .PHONY: format
 format:
-	golangci-lint fmt --config ../.golangci.yml
+	golangci-lint fmt
```
devsergiy, line 19: q: why have you dropped configuration from lint args?

v2/Makefile
```diff
@@ -15,11 +15,11 @@ update-fixtures:
 
 .PHONY: format
 format:
-	golangci-lint fmt --config ../.golangci.yml
+	golangci-lint fmt
```
devsergiy, line 19: same q here

execution/Makefile
devsergiy: config file has formatter rules on the imports order, as this cmd should be run from this folder I don't believe that it will pick up that settings from the repository root

v2/pkg/engine/plan/visitor.go
```diff
@@ -1325,19 +1325,37 @@ func (v *Visitor) configureFetch(internal *objectFetchConfiguration, external re
...
+	if v.Config.DisableIncludeFieldDependencies {
+		return singleFetch
+	}
+	singleFetch.Info.CoordinateDependencies = v.resolveFetchDependencies(internal.fetchID)
+
+	if !v.Config.BuildFetchReasons {
+		return singleFetch
+	}
+	singleFetch.Info.FetchReasons = v.buildFetchReasons(internal.fetchID)
+	if singleFetch.Info.FetchReasons != nil {
```
devsergiy: could also be 
```
if singleFetch.Info.FetchReasons == nil {
  return singleFetch
```

v2/pkg/engine/resolve/fetch.go
```diff
@@ -429,12 +370,28 @@ type FetchReason struct {
...
+	// and which fields from other fetches they depend on.
+	// This information is useful to understand why a fetch depends on other fetches,
+	// and how multiple dependencies lead to a chain of fetches
+	CoordinateDependencies []FetchDependency
+
+	// FetchReasons contains provenance for reasons why particular fields were fetched.
+	// If this structure is built, then all the fields are processed.
+	FetchReasons []FetchReason
+
+	// PropagatedFetchReasons holds those FetchReasons that will be propagated
```
devsergiy, line 392: Could we extend a comment that this will contain information only for the fields which was marked as requiresFetchReason stored in ds configuration

execution/Makefile
ysmolski: I think it was included here: https://github.com/wundergraph/graphql-go-tools/commit/02cda5031355aad38368ae100c6148ca8e9969e8#diff-0789ae4864e2c190021398e4605231919334c5b495badd5cfe92c0ad50f2f6f9
It is redundant because that tool picks config from the parent dir anyway:

```
$ golangci-lint fmt -v
INFO golangci-lint has version 2.4.0 built with go1.25.0 from 43d03392 on 2025-08-13T23:36:29Z
INFO [config_reader] Config search paths: [./ /Users/ys/d/ggt/v2 /Users/ys/d/ggt /Users/ys/d /Users/ys /Users /]
INFO [config_reader] Used config file ../.golangci.yml
```

devsergiy: ah, cool, didn't know that
Could you add a comment about that?

v2/Makefile
ysmolski: same as above and applied in general for all the golangci-lint command.

v2/pkg/engine/resolve/fetch.go
ysmolski: Done

v2/pkg/engine/plan/visitor.go
ysmolski: fixed

execution/Makefile
ysmolski: ok, added

jensneuse (approved):

devsergiy (approved):

PR #1297, opened 2025-09-17
feat: validate presence of optional @requires dependencies

ysmolski:
In GraphQL Federation, when a field has @requires dependencies on
external nullable field and those dependencies fail to resolve, 
the entire entity resolution would lead to malformed data being sent 
to later fetches.

For example, 2 subgraphs:

	# Subgraph 1
	type Entity @key(fields: "id") {
		id: ID
		info: String
	}

	# Subgraph 2, where fullInfo requires 
	# the "info" field to be resolved first.
	type Query {
		entity: [Entity]
	}
	type Entity @key(fields: "id") {
		id: ID
		info: String @external
		fullInfo: String @requires(fields: "info")
	}

If the fetch from Subgraph1 returned the null value for 
the "info" field along with error pointing to that field, then 
this bit of information should not be used in the fetch
from Subgraph 2 that provides info to get the value of
the "fullInfo" field.

This change implements validation of optional @requires dependencies 
in GraphQL Federation. The core feature allows the system to 
gracefully handle scenarios where entities fail to resolve their 
@requires dependencies, particularly for nullable fields.

This change introduces selective handling where only nullable @requires fields 
are treated as "optional" - if they fail, the entity is marked as "tainted" but 
processing continues for other entities.

## Summary by CodeRabbit

- **New Features**
  - Optional runtime validation to omit entities when nullable external "@requires" dependencies fail.
  - Taint-tracking to detect and ignore tainted entities from failed dependency fetches.
  - Planner can emit __typename in nested selections to support requires-driven fetches; fetch metadata now records field nullability.
- **Tests**
  - Expanded suites for federation scenarios, required-fields logic, tainted-object detection, and conditional response routing for tests.


v2/pkg/engine/resolve/loader.go
```diff
@@ -546,75 +571,91 @@ func (l *Loader) mergeResult(fetchItem *FetchItem, res *result, items []*astjson
...
+				if l.handleOptionalRequiresDeps && res.postProcessing.SelectResponseDataPath != nil {
+					taintedIndices = l.getTaintedIndices(fetchItem.Fetch, responseData, responseErrors)
+				}
+				if len(taintedIndices) > 0 {
+					// Override errors with generic error about missing deps.
+					err = l.renderErrorsFailedToFetch(fetchItem, res, missingRequiresDependencies)
+				} else {
+					// Look for errors in the response and merge them into the "errors" array.
+					err = l.mergeErrors(res, fetchItem, responseErrors, errs)
+				}
```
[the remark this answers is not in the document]
ysmolski: @coderabbitai You are kind of right, but this approach is too verbose. Can you think of other ways to remove errors for tainted indexes from the responseErrors?

v2/pkg/engine/plan/configuration.go
```diff
@@ -39,6 +39,13 @@ type Configuration struct {
 	// It may be enabled by some other components of the engine.
 	// It requires DisableIncludeInfo and DisableIncludeFieldDependencies set to false.
 	BuildFetchReasons bool
+
+	// ValidateRequiredExternalFields validates nullable external "@requires" dependencies.
+	// When a subgraph returns a null value with an error for a field set specified in
```
devsergiy: maybe `When a subgraph entity fetch`?

v2/pkg/engine/plan/configuration.go
```diff
@@ -39,6 +39,13 @@ type Configuration struct {
...
+	// the "@requires" directive, any later resolver that depends on it should not receive such an
```
devsergiy: `any following fetch`?

v2/pkg/engine/plan/node_selection_visitor.go
```diff
@@ -46,6 +46,8 @@ type nodeSelectionVisitor struct {
 	hasUnresolvedFields bool // hasUnresolvedFields is used to determine if we need to run the planner again. We should set it to true in case we have unresolved fields
 
 	rewrittenFieldRefs []int
+
+	enforceTypenameForRequired bool
```
devsergiy: missing comment what the field is doing

v2/pkg/engine/plan/required_fields_visitor.go
```diff
@@ -45,6 +45,10 @@ type addRequiredFieldsConfiguration struct {
 	allowTypename                bool
 	typeName                     string
 	fieldSet                     string
+
+	// enforceTypenameForRequired enforces an addition of __typename to selection sets used
+	// in the "requires" key.
+	enforceTypenameForRequired bool
```
devsergiy: maybe `enforceTypenameInNestedSelections`?

v2/pkg/engine/plan/visitor.go
```diff
@@ -1507,3 +1520,16 @@ func (v *Visitor) buildFetchReasons(fetchID int) []resolve.FetchReason {
 	})
 	return reasons
 }
+
+// fieldDefinitionRef returns the definition reference of a field in a given type or ast.InvalidRef if not found.
+func (v *Visitor) fieldDefinitionRef(typeName string, fieldName string) int {
+	node, ok := v.Definition.Index.FirstNodeByNameStr(typeName)
```
devsergiy: you could use a shortcut `v.Definition.NodeByNameStr`

v2/pkg/engine/resolve/loader.go
```diff
@@ -357,17 +365,60 @@ func (l *Loader) resolveSingle(item *FetchItem) error {
...
 			break
 		}
 		items = l.selectItems(items, path[i])
 	}
-	return items
+	return l.filterNonTainted(items)
+}
+
+// filterNonTainted filters out taintedEntities from the given items list.
+func (l *Loader) filterNonTainted(items []*astjson.Value) []*astjson.Value {
```
devsergiy: probably `filterTainted` because we preserve items which is ok

v2/pkg/engine/resolve/loader.go
```diff
@@ -716,7 +795,116 @@ func (l *Loader) appendSubgraphError(res *result, fetchItem *FetchItem, value *a
...
+				typeName := unsafebytes.BytesToString(entity.GetStringBytes("__typename"))
+				if typeName == "" {
+					break
+				}
+				coord := GraphCoordinate{TypeName: typeName, FieldName: field}
+				if _, ok := requestedForRequires[coord]; !ok {
+					break
+				}
+				indices = append(indices, index)
+				errors.Del(strconv.Itoa(errorIdx))
```
devsergiy: I would not remove the original error from the errors. it should be controlled only by error propagation settings

v2/pkg/engine/resolve/loader.go
```diff
@@ -716,7 +795,116 @@ func (l *Loader) appendSubgraphError(res *result, fetchItem *FetchItem, value *a
...
+		errorPath := candidate.Get("path")
+		if astjson.ValueIsNull(errorPath) || errorPath.Type() != astjson.TypeArray {
+			continue
+		}
+		pathItems := errorPath.GetArray()
+		if len(pathItems) == 0 {
+			continue
+		}
+		for i, item := range pathItems {
+			if unsafebytes.BytesToString(item.GetStringBytes()) == "_entities" {
```
devsergiy: we could unwrap if with `!= "_entities"`

v2/pkg/engine/resolve/loader.go
```diff
@@ -716,7 +795,116 @@ func (l *Loader) appendSubgraphError(res *result, fetchItem *FetchItem, value *a
...
+				if len(pathItems)-i <= 2 {
```
devsergiy: please add a comment what `<=2` means

v2/pkg/engine/resolve/loader.go
```diff
@@ -357,17 +365,60 @@ func (l *Loader) resolveSingle(item *FetchItem) error {
...
+
+// isTaintedEntity checks if the given `item` is considered isTaintedEntity in the Loader context.
+// Not only the item is being considered, but also its elements if the item is an array,
+// or its values if the item is an object.
+func (l *Loader) isTaintedEntity(item *astjson.Value) bool {
+	_, ok := l.taintedEntities[item]
+	if ok {
+		return true
+	}
+	switch item.Type() {
```
devsergiy: Are you sure that we need to look inside the values?

selectItems should return objects at the end

v2/pkg/engine/resolve/loader.go
```diff
@@ -716,7 +795,116 @@ func (l *Loader) appendSubgraphError(res *result, fetchItem *FetchItem, value *a
...
+			continue
+		}
+		for i, item := range pathItems {
+			if unsafebytes.BytesToString(item.GetStringBytes()) == "_entities" {
+				if len(pathItems)-i <= 2 {
+					break
+				}
+				// We have the full path to the failed item.
+				// Verify that it is null and extract the enclosing typename.
+				field := unsafebytes.BytesToString(pathItems[len(pathItems)-1].GetStringBytes())
```
devsergiy: example
```
path: ["_entities",0,"nested","a"]
field: "a"
```

v2/pkg/engine/resolve/loader.go
```diff
@@ -716,7 +795,116 @@ func (l *Loader) appendSubgraphError(res *result, fetchItem *FetchItem, value *a
...
+				entity, index := extractEntityIndex(data, pathItems[i+1:len(pathItems)-1])
```
devsergiy: example
```
path: ["_entities",0,"nested","a"]
relativePath: [0,"nested","a"]
```

devsergiy: returned value is not an entity - it's more like `dataForPath`

devsergiy (changes requested):

v2/pkg/engine/resolve/loader.go
ysmolski: Sometimes items are selected deep and merged deep inside of resolvable.data.
It was discovered while writing this test: https://github.com/wundergraph/cosmo/pull/2230/files#diff-a6bdfba4e8514db6043afb1d613de7a04c17c60c0445b129cd160627093903e2R919

v2/pkg/engine/plan/required_fields_visitor.go
ysmolski: renamed to `addTypenameInNestedSelections`

v2/pkg/engine/resolve/loader.go
ysmolski: "filter" is kinda ambiguous. Renamed to `removeTainted`

v2/pkg/engine/resolve/loader.go
ysmolski: To expand on the test above. It happens when items are selected from deep levels of the resolvable.data:
```json
{
  "products" : [ {
    "__typename" : "Consultancy",
    "upc" : "consultancy",
    "lead" : {
      "__typename" : "Employee",
      "id" : 1
    }
  }, {
    "__typename" : "Cosmo",
    "upc" : "cosmo",
    "lead" : {
      "__typename" : "Employee",
      "id" : 2
    }
  }, {
    "__typename" : "SDK"
  } ]
}
```
In this case the following items are selected for the fetch:
```json
Item 0:
      "__typename" : "Employee",
      "id" : 1
Item 1:
      "__typename" : "Employee",
      "id" : 2
```

v2/pkg/engine/resolve/loader.go
ysmolski: Ok, let me check how it would look like. Most probably, I would want to change the error that I add.

ysmolski: Ok, I have fixed it to keep errors.

v2/pkg/engine/resolve/loader.go
ysmolski: renamed it

v2/pkg/engine/resolve/loader.go
ysmolski: One of those fetches contains null and 
```{
      "__typename" : "Employee",
      "id" : 1,
      "isAvailable" : null
    }
``` 
is marked as tainted.
The later fetch selects item:
```
{
    "__typename" : "Consultancy",
    "upc" : "consultancy",
    "lead" : {
      "__typename" : "Employee",
      "isAvailable" : null
    }
  }
```
And if when selecting, we do not look deeply, we use this item for the fetch.

v2/pkg/engine/resolve/loader.go
```diff
@@ -1132,8 +975,14 @@ func (l *Loader) optionallyRewriteErrorPaths(fetchItem *FetchItem, values []*ast
 				newPath := make([]string, 0, len(pathPrefix)+len(pathItems)-i)
 				newPath = append(newPath, pathPrefix...)
 				for j := i + 1; j < len(pathItems); j++ {
-					// BUG: for pathItems containing integers, it will append empty strings
-					newPath = append(newPath, unsafebytes.BytesToString(pathItems[j].GetStringBytes()))
+					switch pathItems[j].Type() {
+					case astjson.TypeString:
+						newPath = append(newPath, unsafebytes.BytesToString(pathItems[j].GetStringBytes()))
+					case astjson.TypeNumber:
+						newPath = append(newPath, strconv.Itoa(pathItems[j].GetInt()))
```
devsergiy, line 982: if we under "_entities" - I believe we need to skip next integer path item - because it will be index of concrete entity

right now we can't map it back to original location of object, so it will be more simpe to skip it

devsergiy (changes requested):

execution/engine/execution_engine_test.go
```diff
@@ -4547,6 +4556,702 @@ func TestExecutionEngine_Execute(t *testing.T) {
 			withFetchReasons(),
 		))
 	})
+
+	t.Run("execute operation with requires and external", func(t *testing.T) {
```
devsergiy: please wrap these cases into

```
t.Run("validation of optional requires deps", func(t *testing.T) {
}
```

PR #1312, opened 2025-10-07
fix: propagate fetch reasons for interface-related fields

ysmolski:
Interfaces were not handled before at all. 
Now, if interface field is marked with special directive,
then fields on implementing type are considered as marked too.

You can find more details in the comments.

## Summary by CodeRabbit

- Bug Fixes
  - Improved propagation of fetch reasons across interfaces, objects and fragments so subrequest payloads reliably include all relevant implementing types and avoid duplicates.

- Tests
  - Expanded coverage for interface/object/fragment combinations, multiple implementers, single-implementer cases, and duplicate-avoidance scenarios.

- Refactor
  - Internal fetch-reason propagation logic reworked for clearer, deduplicated results and consistent sorting.

- Chores (breaking)
  - Simplified public metadata shape by removing a previously exported fetch-reason field from root-node metadata (may require client updates).

### DIFF
--- a/v2/pkg/engine/plan/visitor.go
+++ b/v2/pkg/engine/plan/visitor.go
@@ -1555,6 +1555,8 @@ func (v *Visitor) getPropagatedReasons(fetchID int, fetchReasons []resolve.Fetch
 			propagated[i].ByUser = propagated[i].ByUser || reason.ByUser
 			if len(reason.BySubgraphs) > 0 {
 				propagated[i].BySubgraphs = append(propagated[i].BySubgraphs, reason.BySubgraphs...)
+				slices.Sort(propagated[i].BySubgraphs)
+				propagated[i].BySubgraphs = slices.Compact(propagated[i].BySubgraphs)
 				propagated[i].IsKey = propagated[i].IsKey || reason.IsKey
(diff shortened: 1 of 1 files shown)


v2/pkg/engine/plan/visitor.go
```diff
@@ -1533,3 +1525,97 @@ func (v *Visitor) fieldDefinitionRef(typeName string, fieldName string) int {
...
+		if !exists {
+			continue
+		}
+
+		// Special case when the field belongs to an object, and this field is not in the lookup.
+		// If this object implements an interface that has a corresponding field in the lookup,
+		// then propagate it.
+		if typeNode.Kind == ast.NodeKindObjectTypeDefinition && !fieldInLookup {
+			objectDef := v.Definition.ObjectTypeDefinitions[typeNode.Ref]
+			for _, interfaceRef := range objectDef.ImplementsInterfaces.Refs {
```
devsergiy: will it be better to use name `interfaceTypeNameRef`?

ysmolski: Hmm. I have renamed to `interfaceTypeRef`

PR #1323, opened 2025-10-16
feat: add support for field resolvers

Noroth:
## Summary by CodeRabbit

* **New Features**
  * Field resolver support expanded: Product (shippingEstimate, recommendedCategory) and Category (productCount, popularityScore, categoryMetrics, subcategories).
  * New schema types and inputs: Subcategory, CategoryMetrics, ProductCountFilter, SubcategoryItemFilter, ShippingEstimateInput, ShippingDestination enum, and openfed__FieldSet scalar.

* **Improvements**
  * Execution engine now builds and runs dependency-aware, parallel RPC call plans for resolver-driven and federated flows, with richer response merging.

* **Tests & Tooling**
  * New tests, benchmarks, mock service handlers, and mapping generation helpers; package Readme added.

## Checklist


Noroth:
@coderabbitai review

v2/pkg/engine/datasource/grpc_datasource/grpc_datasource.go
```diff
@@ -80,74 +88,93 @@ func NewDataSource(client grpc.ClientConnInterface, config DataSourceConfig) (*D
...
+				results[index] = resultData{
+					kind:         serviceCall.RPC.Kind,
+					response:     response,
+					responsePath: serviceCall.RPC.ResponsePath,
+				}
+				mu.Unlock()
+
+				return nil
+			})
+		}
```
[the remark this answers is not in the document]
Noroth: This was fixed in Go 1.22

v2/pkg/engine/datasource/grpc_datasource/grpc_datasource_test.go
```diff
@@ -27,6 +27,79 @@ import (
...
+	})
+	require.NoError(b, err)
+
+	b.ReportAllocs()
+	b.ResetTimer()
+	for b.Loop() {
+		output := new(bytes.Buffer)
+		err = ds.Load(context.Background(), []byte(`{"query":"`+query+`","body":`+variables+`}`), output)
+		require.NoError(b, err)
+	}
```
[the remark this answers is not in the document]
Noroth: This was added with Go 1.24
https://go.dev/blog/testing-b-loop

v2/pkg/engine/datasource/grpc_datasource/compiler.go
```diff
@@ -388,23 +575,291 @@ func (p *RPCCompiler) resolveServiceName(methodName string) (string, bool) {
...
+		return nil
+	}
+
+	if fd.IsList() {
+		return p.resolveListDataForPath(msg.List(), fd, path)
+	}
+
+	return p.resolveDataForPath(msg.Message(), path)
+
+}
```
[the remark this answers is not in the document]
Noroth: This is already handled in `resolveDataForPath`

v2/pkg/engine/datasource/grpc_datasource/compiler.go
```diff
@@ -388,23 +575,291 @@ func (p *RPCCompiler) resolveServiceName(methodName string) (string, bool) {
...
+
+			result = append(result, values...)
+		default:
+			result = append(result, item)
+		}
+	}
+
+	return result
+}
+
```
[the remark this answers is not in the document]
Noroth: Go 1.22 introduced range over integers
https://tip.golang.org/doc/go1.22

ysmolski:
Holy cheese! Even though it is big, most of it is autogen. I will try to review it, just let me know when you are done with coderabbitai

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -452,12 +472,19 @@ func (r *rpcPlanningContext) newMessageFromSelectionSet(enclosingTypeNode ast.No
...
 func (r *rpcPlanningContext) resolveFieldMapping(typeName, fieldName string) string {
-	grpcFieldName, ok := r.mapping.ResolveFieldMapping(typeName, fieldName)
-	if !ok {
-		return fieldName
+	if grpcFieldName, ok := r.mapping.FindFieldMapping(typeName, fieldName); ok {
+		return grpcFieldName
+	}
+
+	return fieldName
+}
```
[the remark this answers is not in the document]
Noroth: I think it is okay to panic here. We should always provide a mapping. If the mapping is nil something is completely wrong.

v2/pkg/engine/datasource/grpc_datasource/compiler.go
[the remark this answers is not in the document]
Noroth: This is fine in this case. Logically we can only have either a list or a message here at this point in time. The only other option would be when the parent call would return a single scalar value, but in this case we couldn't have a field resolver as it would require it be inside of an object type. Therefore the parent function must return either a message or a list

v2/pkg/engine/datasource/grpc_datasource/fetch_test.go
```diff
@@ -0,0 +1,178 @@
...
+				MethodName: "Method6",
+			},
+		},
+	}
+	graph := NewDependencyGraph(executionPlan)
+	for b.Loop() {
+		_ = graph.TopologicalSortResolve(func(nodes []FetchItem) error {
+			return nil
+		})
+	}
```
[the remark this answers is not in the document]
Noroth: Same as below, was added in Go 1.24

v2/pkg/engine/datasource/grpc_datasource/fetch.go
```diff
@@ -0,0 +1,144 @@
...
+
+	for index, call := range executionPlan.Calls {
+		graph.nodes[index] = call.DependentCalls
+		graph.fetches[index] = FetchItem{
+			ID:               index,
+			Plan:             &call,
+			ServiceCall:      nil,
+			DependentFetches: call.DependentCalls,
+		}
+	}
```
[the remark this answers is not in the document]
Noroth: Fixed in Go 1.22

v2/pkg/engine/datasource/grpc_datasource/json_builder.go
```diff
@@ -217,6 +220,88 @@ func (j *jsonBuilder) mergeEntities(left *astjson.Value, right *astjson.Value) (
...
+	if len(resolvedValues) < len(responseValues) {
+		return fmt.Errorf("length of values doesn't match the length of the result array, expected %d, got %d", len(responseValues), len(resolvedValues))
+	}
+
+	for i := range responseValues {
+		responseValues[i].Set(elementName, resolvedValues[i].Get(elementName))
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
Noroth: No you're wrong here.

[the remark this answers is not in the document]
Noroth: resolved.GetArray(resolveResponsePath) is valid.

v2/pkg/engine/datasource/grpc_datasource/compiler.go
```diff
@@ -14,6 +16,11 @@ import (
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/operationreport"
 )
 
+const (
+	// InvalidRef is a constant used to indicate that a reference is invalid.
+	InvalidRef = -1
```
ysmolski, line 21: you could use `ast.InvalidRef` instead

v2/pkg/engine/datasource/grpc_datasource/fetch.go
```diff
@@ -0,0 +1,144 @@
...
+
+	return graph
+}
+
+func (g *DependencyGraph) TopologicalSortResolve(resolver func(nodes []FetchItem) error) error {
```
ysmolski, line 59: This is heavy function with recursive traversal. Could you comment it what it does?

As I have understood from the code, it arranges calls into levels (chunks) so calls on each level can be executed in parallel. Level 0 does not depends on anything and Level 1 depends on level 0, and so on.

v2/pkg/engine/datasource/grpc_datasource/compiler.go
Noroth: I had it before. I started using my own type because it's not related to ast.Document, it's more like a similar principle.

v2/pkg/engine/datasource/grpc_datasource/fetch.go
Noroth: Yes you got that correct. I can add some comments.

v2/pkg/engine/datasource/grpc_datasource/compiler.go
```diff
@@ -99,6 +121,41 @@ type Document struct {
 	Methods  []Method  // All methods from all services in the document
 }
 
+// newNode creates a new node in the document.
+func (d *Document) newNode(ref int, name string, kind NodeKind) {
+	h := xxhash.Sum64String(name)
```
jensneuse: better to use pool.Hash

v2/pkg/engine/datasource/grpc_datasource/compiler.go
```diff
@@ -325,19 +406,129 @@ func (p *RPCCompiler) ConstructExecutionPlan(operation, schema *ast.Document) (*
...
+// 	}
+
+// 	for invocation := range resultChan {
+// 		invocations = append(invocations, invocation)
+// 	}
+
+// 	return invocations, nil
+// }
+
+func (p *RPCCompiler) CompileFetches(graph *DependencyGraph, fetches []FetchItem, inputData gjson.Result) ([]ServiceCall, error) {
```
jensneuse, line 458: I'd prefer if we could all standardize on astjson

v2/pkg/engine/datasource/grpc_datasource/compiler.go
```diff
@@ -388,23 +579,298 @@ func (p *RPCCompiler) resolveServiceName(methodName string) (string, bool) {
...
+			items = append(items, p.resolveUnderlyingListItems(itemsList.Get(i), nestingLevel-1)...)
+		}
+
+		return items
+	}
+
+	result := make([]protoref.Value, itemsList.Len())
+	for i := 0; i < itemsList.Len(); i++ {
+		result[i] = itemsList.Get(i)
+	}
```
jensneuse:
```suggestion
	n := itemsList.Len()
	result := make([]protoref.Value, n)
	for i := 0; i < n; i++ {
		result[i] = itemsList.Get(i)
	}
```

v2/pkg/engine/datasource/grpc_datasource/configuration.go
```diff
@@ -6,11 +6,17 @@ import (
...
+	RPCConfig | ResolveRPCMapping
+	// TODO: add requires RPC mapping
+}
+
 type (
 	// RPCConfigMap is a map of RPC names to RPC configurations
 	// The key is the field name in the GraphQL operation type (query, mutation, subscription).
 	// The value is the RPC configuration for that field.
-	RPCConfigMap map[string]RPCConfig
+	RPCConfigMap[T RPCMapping] map[string]T
```
jensneuse, line 19: trpc hihi

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -537,10 +564,469 @@ func (r *rpcPlanningContext) buildField(enclosingTypeNode ast.Node, fd int, fiel
...
+
+			if err != nil {
+				return nil, err
+			}
+
+			contextMessage.Fields = append(contextMessage.Fields, field)
+		}
+
+		fieldArgsMessage.Fields = make(RPCFields, 0, len(resolvedField.fieldArguments))
+		for _, fieldArgument := range resolvedField.fieldArguments {
```
jensneuse: it's faster to just range the index and then index access the items

v2/pkg/engine/datasource/grpc_datasource/execution_plan_field_resolvers_test.go
```diff
@@ -0,0 +1,894 @@
...
+											{
+												Name:        "id",
+												TypeName:    string(DataTypeString),
+												JSONPath:    "id",
+												ResolvePath: buildPath("categories.id"),
+											},
+											{
+												Name:        "name",
+												TypeName:    string(DataTypeString),
+												JSONPath:    "name",
```
jensneuse: I think it would be better (and more efficient) to align the formats here with what astjson needs so that we can use astjson at execution without transforming again

v2/pkg/engine/datasource/grpc_datasource/execution_plan_test.go
```diff
@@ -49,6 +50,18 @@ func runTest(t *testing.T, testCase testCase) {
...
+	pathElements := strings.Split(path, ".")
+	pathItems := make([]ast.PathItem, 0, len(pathElements))
+	for _, element := range pathElements {
+		pathItems = append(pathItems, ast.PathItem{
+			Kind:      ast.FieldName,
+			FieldName: []byte(element),
+		})
+	}
+	return pathItems
+}
```
jensneuse, line 76:
```suggestion
func buildPath(path string) ast.Path {
    // Copy the whole path once, then take subslices—no per-segment []byte allocations.
    b := make([]byte, len(path))
    copy(b, path)

    // Count segments
    n := 1
    for i := 0; i < len(b); i++ {
        if b[i] == '.' {
            n++
        }
    }

    items := make([]ast.PathItem, n)
    start, seg := 0, 0
    for i := 0; i <= len(b); i++ {
        if i == len(b) || b[i] == '.' {
            items[seg] = ast.PathItem{
                Kind:      ast.FieldName,
                FieldName: b[start:i], // subslice into the single buffer
            }
            seg++
            start = i + 1
        }
    }
    return items
}
```

v2/pkg/engine/datasource/grpc_datasource/fetch.go
```diff
@@ -0,0 +1,175 @@
...
+		dependencies = append(dependencies, dependency)
+	}
+
+	return dependencies, nil
+}
+
+// SetFetchData sets the service call for a given index.
+func (g *DependencyGraph) SetFetchData(index int, serviceCall *ServiceCall) {
+	g.mu.Lock()
+	g.fetches[index].ServiceCall = serviceCall
```
jensneuse, line 161: why does it need a mux?

v2/pkg/engine/datasource/grpc_datasource/grpc_datasource.go
```diff
@@ -80,74 +88,93 @@ func NewDataSource(client grpc.ClientConnInterface, config DataSourceConfig) (*D
...
+				// The number of entities per type must match the number of lookup keys in the variablese
+				if serviceCall.RPC.Kind == CallKindEntity {
+					err = builder.validateFederatedResponse(response)
+					if err != nil {
+						return err
+					}
+				}
+
+				mu.Lock()
+				results[index] = resultData{
```
jensneuse, line 141: you can solve this lock free, create a list of results (pointers to results), pass each result pointer to the fetching goroutine, as no other goroutine touches that result item, it's safe to write to before the errgroup ends, we do this in the "engine loader"

v2/pkg/engine/datasource/grpc_datasource/json_builder.go
```diff
@@ -217,6 +220,88 @@ func (j *jsonBuilder) mergeEntities(left *astjson.Value, right *astjson.Value) (
 	return entities, nil
 }
 
+func (j *jsonBuilder) mergeWithPath(base *astjson.Value, resolved *astjson.Value, path ast.Path) error {
+	resolvedValues := resolved.GetArray(resolveResponsePath)
+
+	searchPath := path[:len(path)-1]
+	elementName := path[len(path)-1].FieldName.String()
```
jensneuse, line 232: are you sure both paths are always long enough?

v2/pkg/grpctest/cmd/mapping_helper/main.go
```diff
@@ -0,0 +1,149 @@
+package main
```
jensneuse, line 3: add docs, what's the purpose of this? how to use? why?

jensneuse (changes requested):

v2/pkg/engine/datasource/grpc_datasource/compiler.go
ysmolski: I have to agree with Jens about deprecating gjson in our codebase. Gjson is not a good library to continue using. They broke their API in minor versions and now we are stuck on some very old version and the only way forward with it is to get rid of gjson and use astjson instead. I suggest to get get rid of `gjson` completely in the future.

v2/pkg/grpctest/cmd/mapping_helper/main.go
Noroth: There is a readme. It is basically just a helper tool for me because I always created the mapping from hand for my tests.

v2/pkg/engine/datasource/grpc_datasource/json_builder.go
Noroth: Well if not something is really wrong in the planner. But I can add a length check and return a proper error to prevent a panic.

v2/pkg/engine/datasource/grpc_datasource/compiler.go
Noroth: Yeah I agree here. I already wanted to get rid of it. I would prefer to do it in a separate PR to reduce the scope of this one. But I can pick up on that next

v2/pkg/engine/datasource/grpc_datasource/compiler.go
```diff
@@ -122,6 +185,18 @@ type Message struct {
...
 }
 
+// FieldByName returns a field by its name.
+// Returns nil if no field with the given name exists.
+func (m *Message) FieldByName(name string) *Field {
+	for index, field := range m.Fields {
+		if field.Name == name {
+			return &m.Fields[index]
+		}
+	}
```
ysmolski: Is this method used in the hotpath? I am always suspicous to see search loops like this. What if the number of Fields would be large? The same applies to `ServiceByName` and `MethodByName`

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -2,16 +2,24 @@ package grpcdatasource
...
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/astvisitor"
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/plan"
 )
 
 const (
 	// knownTypeOptionalFieldValueName is the name of the field that is used to wrap optional scalar values
 	// in a message as protobuf scalar types are not nullable.
 	knownTypeOptionalFieldValueName = "value"
+
+	// resolverContextDirectiveName is the name of the directive that is used to configure the resolver context.
```
ysmolski: I could not find documentation on what does "connect__fieldResolver" do. Is it documented somewhere else?

v2/pkg/grpctest/cmd/mapping_helper/main.go
ysmolski: I suggest to add one-two sentences like you have just explained to prevent frustrations for others.

v2/pkg/engine/datasource/grpc_datasource/compiler.go
Noroth: Good point. I can improve those messages since I added the nodes.
For the fields it's tricky. Those are not going into the nodes. I would probably leave it for now as this method is only used for one case atm.

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
Noroth: It will be documented in protographic and in Cosmo docs.

ysmolski: fair enough

v2/pkg/engine/datasource/grpc_datasource/compiler.go
ysmolski: Okay. I have not checked deeply if this is used in the hotpath. What I have seen so far, was not using those from deeply nested loops.

v2/pkg/engine/datasource/grpc_datasource/compiler.go
ysmolski: Sure, separate PR is good enough. Also it does have to be astjson if there a way to implement it even more simple, like using regular structs meaning I am not insisting on specific implementation.

v2/pkg/engine/datasource/grpc_datasource/fetch.go
```diff
@@ -0,0 +1,171 @@
...
+	return dependencies, nil
+}
+
+// SetFetchData sets the service call for a given index.
+func (g *DependencyGraph) SetFetchData(index int, serviceCall *ServiceCall) {
+	g.fetches[index].ServiceCall = serviceCall
+}
+
+// initializeSlice initializes a slice with a given length and a given value.
+func initializeSlice[T any](len int, zero T) []T {
```
StarpTech: Move to a util/helper module?

Noroth: We don't have a dedicated package for utils. I put it into a `utils.go`

v2/pkg/engine/datasource/grpc_datasource/compiler.go
```diff
@@ -387,44 +556,313 @@ func (p *RPCCompiler) resolveServiceName(methodName string) (string, bool) {
...
+		return nil, fmt.Errorf("context field not found in message %s", rpcMessage.Name)
+	}
+
+	contextField := rootMessage.Descriptor().Fields().ByNumber(protoref.FieldNumber(contextSchemaField.Number))
+	if contextField == nil {
+		return nil, fmt.Errorf("context field not found in message %s", inputMessage.Name)
+	}
+
+	contextList := p.newEmptyListMessageByName(rootMessage, contextSchemaField.Name)
+	contextData := p.resolveContextData(context[0], contextRPCField) // TODO handle multiple contexts (resolver requires another resolver)
```
devsergiy, line 601: should we check the length of the context to avoid panic?

v2/pkg/engine/datasource/grpc_datasource/compiler.go
```diff
@@ -387,44 +556,313 @@ func (p *RPCCompiler) resolveServiceName(methodName string) (string, bool) {
...
+	if argsRPCField == nil {
+		return nil, fmt.Errorf("field_args field not found in message %s", rpcMessage.Name)
+	}
+
+	args, err := p.buildProtoMessage(argsMessage, argsRPCField.Message, data)
+	if err != nil {
+		return nil, err
+	}
+
+	// // Set the key list
```
devsergiy: what is the meaning of key?

v2/pkg/engine/datasource/grpc_datasource/compiler.go
```diff
@@ -387,44 +556,313 @@ func (p *RPCCompiler) resolveServiceName(methodName string) (string, bool) {
...
+	}
+
+	contextValues := make([]map[string]protoref.Value, 0)
+	for _, field := range contextField.Message.Fields {
+		values := p.resolveContextDataForPath(context.ServiceCall.Output, field.ResolvePath)
+
+		contextValues = slices.Grow(contextValues, len(values))
+
+		for index, value := range values {
+			if index >= len(contextValues) {
```
devsergiy, line 646: a bit unclear for me, if I got it right, we have for example 3 fields let's say each of them has 10 items

so we will grow slice 3 times by 10 items

but we reuse indexes, so actually we need only 10 items in a context values

probably it also makes sense to comment what this code is doing

v2/pkg/engine/datasource/grpc_datasource/compiler.go
```diff
@@ -387,44 +556,313 @@ func (p *RPCCompiler) resolveServiceName(methodName string) (string, bool) {
...
+
+			contextValues[index][field.Name] = value
+		}
+
+	}
+
+	return contextValues
+}
+
+func (p *RPCCompiler) resolveContextDataForPath(message protoref.Message, path ast.Path) []protoref.Value {
```
devsergiy, line 659: lacks the method comment

v2/pkg/engine/datasource/grpc_datasource/compiler.go
```diff
@@ -387,44 +556,313 @@ func (p *RPCCompiler) resolveServiceName(methodName string) (string, bool) {
...
+
+	if fd.IsList() {
+		return p.resolveListDataForPath(msg.List(), fd, path)
+	}
+
+	return p.resolveDataForPath(msg.Message(), path)
+
+}
+
+func (p *RPCCompiler) resolveListDataForPath(message protoref.List, fd protoref.FieldDescriptor, path ast.Path) []protoref.Value {
```
devsergiy, line 681: lacks the method comment

v2/pkg/engine/datasource/grpc_datasource/compiler.go
```diff
@@ -387,44 +556,313 @@ func (p *RPCCompiler) resolveServiceName(methodName string) (string, bool) {
...
+		}
+
+		return p.resolveDataForPath(field.Message(), path[1:])
+	default:
+		return []protoref.Value{field}
+	}
+}
+
+// getMessageField gets the field from the message by its name.
+// It also handles nested lists and nullable lists.
```
devsergiy: > It also handles nested lists and nullable lists.
is it true currently?

v2/pkg/engine/datasource/grpc_datasource/compiler.go
```diff
@@ -387,44 +556,313 @@ func (p *RPCCompiler) resolveServiceName(methodName string) (string, bool) {
...
+	if fd == nil {
+		return nil
+	}
+
+	listMsg := msg.Get(fd)
+	if !listMsg.IsValid() {
+		return nil
+	}
+
+	itemsValue := listMsg.Message().Get(listMsg.Message().Descriptor().Fields().ByNumber(1))
```
devsergiy: >ByNumber(1)

what 1 means?

v2/pkg/engine/datasource/grpc_datasource/compiler.go
```diff
@@ -387,44 +556,313 @@ func (p *RPCCompiler) resolveServiceName(methodName string) (string, bool) {
...
 		return nil
 	}
 
+	return p.resolveUnderlyingListItems(listFieldValue, nestingLevel)
+
+}
+
+func (p *RPCCompiler) resolveUnderlyingListItems(value protoref.Value, nestingLevel int) []protoref.Value {
+	msg := value.Message()
+	fd := msg.Descriptor().Fields().ByNumber(1)
```
devsergiy: > ByNumber(1)

what 1 means?

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
devsergiy, line 336: as we always should provide the mapping

should it be an error when mapping is nil?

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -530,17 +560,480 @@ func (r *rpcPlanningContext) buildField(enclosingTypeNode ast.Node, fd int, fiel
...
+	}
+
+	field.Message = fieldMessage
+	return field, nil
+}
+
+// buildMessageFromInputObjectType builds a message from an input object type definition.
+func (r *rpcPlanningContext) buildMessageFromInputObjectType(node *ast.Node) (*RPCMessage, error) {
+	if node.Kind != ast.NodeKindInputObjectTypeDefinition {
+		return nil, fmt.Errorf("unsupported type: %s", node.Kind)
```
devsergiy: should it be an incorrect type for this method? why unsupported?

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -530,17 +560,480 @@ func (r *rpcPlanningContext) buildField(enclosingTypeNode ast.Node, fd int, fiel
...
+		}
+
+		message.Fields = append(message.Fields, field)
+	}
+
+	return message, nil
+}
+
+func (r *rpcPlanningContext) buildMessageFieldFromInputValueDefinition(ivdRef int, node *ast.Node) (RPCField, error) {
+	ivd := r.definition.InputValueDefinitions[ivdRef]
```
devsergiy: will it be better to use longer, more readable names?

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -530,17 +560,480 @@ func (r *rpcPlanningContext) buildField(enclosingTypeNode ast.Node, fd int, fiel
...
+	ivdType := r.definition.Types[ivd.Type]
+
+	underlyingTypeNode, found := r.nodeByTypeRef(ivd.Type)
```
devsergiy: resolveUnderlyingType bypases lists, is it okay here?
e.g. if input type field is a list?

devsergiy: ah, I see, it is handled in buildInputMessageField

worth a comment

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -530,17 +560,480 @@ func (r *rpcPlanningContext) buildField(enclosingTypeNode ast.Node, fd int, fiel
...
+	fieldRef               int
+	fieldDefinitionTypeRef int
+	requiredFieldSelection int
+	responsePath           ast.Path
+
+	contextFields  []contextField
+	fieldArguments []fieldArgument
+}
+
+func (r *rpcPlanningContext) setResolvedField(walker *astvisitor.Walker, fd int, fieldArgs []int, fieldPath ast.Path, resolvedField *resolvedField) error {
```
devsergiy: it is unclear what `fd` argument is, without going through the set of all methods calls - to actually see that it is a FieldDefinitionRef

e.g. is it file descriptor, is it field, or field definition

devsergiy: suggest to use a longer name like FieldDefinitionRef

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -530,17 +560,480 @@ func (r *rpcPlanningContext) buildField(enclosingTypeNode ast.Node, fd int, fiel
...
+			r.definition.FieldDefinitionNameBytes(fd),
+			r.operation.ArgumentNameBytes(fieldArgRef),
+		)
+
+		if argDefRef == ast.InvalidRef {
+			return nil, fmt.Errorf("unable to resolve argument input value definition for argument %s", fieldArg)
+		}
+
+		jsonValue := fieldArg
+		if fieldType == ast.ValueKindVariable {
```
devsergiy: we probably should never have other than variable kind, should we raise an error in case of other type?

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -530,17 +560,480 @@ func (r *rpcPlanningContext) buildField(enclosingTypeNode ast.Node, fd int, fiel
...
+			jsonPath:              jsonValue,
+		})
+
+	}
+
+	return result, nil
+
+}
+
+func (r *rpcPlanningContext) getFieldsFromContext(parentNode ast.Node, contextRef int) ([]int, error) {
```
devsergiy: will suggest to use fully qualified method name `getFieldsFromContextDirective(parentNode ast.Node, contextDirectiveRef int)`

otherwise it feels like we are getting smth from go context

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -530,17 +560,480 @@ func (r *rpcPlanningContext) buildField(enclosingTypeNode ast.Node, fd int, fiel
...
+func (r *rpcPlanningContext) filterIDFieldsFunc(o ast.ObjectTypeDefinition, fd int) func(yield func(int) bool) {
+	fieldRefs := o.FieldsDefinition.Refs
+	return func(yield func(int) bool) {
+		for _, ref := range fieldRefs {
+			if ref == fd {
+				continue
+			}
+
+			typeName := r.definition.FieldDefinitionTypeNameString(ref)
+			if typeName != "ID" {
```
devsergiy: do we allow only fields with the `ID` type in context fields?

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -530,17 +560,480 @@ func (r *rpcPlanningContext) buildField(enclosingTypeNode ast.Node, fd int, fiel
...
+
+			if !yield(ref) {
+				return
+			}
+		}
+	}
+}
+
+// nodeByTypeRef is a helper function to resolve the underlying type node for a given type reference.
+func (r *rpcPlanningContext) nodeByTypeRef(typeRef int) (ast.Node, bool) {
```
devsergiy: should this be the ast helper? like `ResolveNodeFromTypeRef`?

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -530,17 +560,480 @@ func (r *rpcPlanningContext) buildField(enclosingTypeNode ast.Node, fd int, fiel
...
+}
+
+const (
+	resultFieldName    = "result"
+	contextFieldName   = "context"
+	fieldArgsFieldName = "field_args"
+)
+
+func (r *rpcPlanningContext) newResolveRPCCall(config *resolveRPCCallConfig) (RPCCall, error) {
+	resolveConfig := config.resolveConfig
```
devsergiy, line 996: Do we reassign for shorter names or to copy values?

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -51,9 +59,24 @@ type RPCExecutionPlan struct {
...
+)
+
 // RPCCall represents a single call to a gRPC service method.
 // It contains all the information needed to make the call and process the response.
 type RPCCall struct {
+	// Kind of call, used to decide how to execute the call
+	// This is used to identify the call type and execution behaviour.
+	Kind CallKind
 	// DependentCalls is a list of calls that must be executed before this call
 	DependentCalls []int
```
devsergiy, line 81: missed that during previuos reviews

do you think RPCCall should have an explicit `CallID`? It still will be the initial index, the same as for fetches in resolver, but it will be a bit more obvious

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor.go
```diff
@@ -109,18 +147,48 @@ func (r *rpcPlanVisitor) EnterOperationDefinition(ref int) {
 //
 // TODO handle field arguments to define resolvers
```
devsergiy: could we remove todo now?

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor.go
```diff
@@ -109,18 +147,48 @@ func (r *rpcPlanVisitor) EnterOperationDefinition(ref int) {
...
+	// As we check that we are inside of a field we can safely access the second to last type definition.
+	parentTypeNode := r.walker.TypeDefinitions[len(r.walker.TypeDefinitions)-2]
+	fieldDefinitionRef, exists := r.definition.NodeFieldDefinitionByName(parentTypeNode, r.operation.FieldNameBytes(ancestor.Ref))
+	if !exists {
+		return
+	}
+
+	argument := r.operation.ArgumentValue(ref)
+	jsonPath := r.operation.ArgumentNameString(ref)
+	if argument.Kind == ast.ValueKindVariable {
```
devsergiy, line 171: should we raise an error here too? Or do we handle it gracefully?

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor.go
```diff
@@ -290,8 +405,18 @@ func (r *rpcPlanVisitor) EnterField(ref int) {
 
 // LeaveField implements astvisitor.FieldVisitor.
 func (r *rpcPlanVisitor) LeaveField(ref int) {
+	r.fieldPath = r.fieldPath.RemoveLastItem()
+	r.resolvedFieldIndex = ast.InvalidRef
```
devsergiy, line 408: q: should we pop to previous resolved field?
e.g. imaging following nesting
```graphql
query ($filter1, $filter2){
   entities {
     id
     filteredNestedEntities(filter: $filter1) {
        id
        filteredDoubleNestedEntities(filter: $filter2) { # here on leave we set invalid ref to resolved field index
          name
        }
        someOtherField # but this field should be added to top resolved field 
     }
   }
}
```

Do we support such nested resolving?
Note: this could be a regular call - because subgraph owns all entities

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor_federation.go
```diff
@@ -45,10 +46,15 @@ type rpcPlanVisitorFederation struct {
...
 
-	plan             *RPCExecutionPlan
-	subgraphName     string
-	currentCall      *RPCCall
-	currentCallIndex int
+	plan         *RPCExecutionPlan
+	subgraphName string
+	currentCall  *RPCCall
+
+	relatedCallID      int
```
devsergiy: relatedCall - is it a call that needs dependency, or which provides dependency?

v2/pkg/engine/datasource/grpc_datasource/fetch.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	// After setting up the call hierarchy, we are grouping the calls by their level.
+	chunks := make(map[int][]FetchItem)
+	for callIndex, chunkIndex := range callHierarchyRefs {
+		chunks[chunkIndex] = append(chunks[chunkIndex], g.fetches[callIndex])
+	}
+
+	// We are iterating over the chunks and resolving the calls in the correct order.
+	for i := 0; i < len(chunks); i++ {
```
devsergiy, line 126: do not like that this method is sorting and calling callback, aka doing fetches

devsergiy: I would prefer to have this method be responsible only for sorting, and how to execute is the other thing

Also, with this approach, you probably can't cache the order of execution

v2/pkg/engine/datasource/grpc_datasource/grpc_datasource.go
```diff
@@ -80,74 +87,90 @@ func NewDataSource(client grpc.ClientConnInterface, config DataSourceConfig) (*D
...
-			err := d.cc.Invoke(errGrpCtx, methodName, invocation.Input, invocation.Output)
-			if err != nil {
-				return err
-			}
+		results := make([]resultData, len(serviceCalls))
+		errGrp, errGrpCtx := errgroup.WithContext(ctx)
+
+		// make gRPC calls
+		for index, serviceCall := range serviceCalls {
+			errGrp.Go(func() error {
```
devsergiy, line 118: So we will have parallel calls only within the current chunk of fetches? And chunks are always serial?

v2/pkg/engine/datasource/grpc_datasource/json_builder.go
```diff
@@ -217,6 +220,92 @@ func (j *jsonBuilder) mergeEntities(left *astjson.Value, right *astjson.Value) (
...
+	}
+
+	for i := range responseValues {
+		responseValues[i].Set(elementName, resolvedValues[i].Get(elementName))
+	}
+
+	return nil
+}
+
+func (j *jsonBuilder) flattenObject(value *astjson.Value, path ast.Path) ([]*astjson.Value, error) {
```
devsergiy, line 268: Could you clarify what flatten means here? Maybe some examples in tests?

v2/pkg/engine/datasource/grpc_datasource/compiler.go
Noroth: This is already done in `CompileNode`

v2/pkg/engine/datasource/grpc_datasource/compiler.go
Noroth: Probably an old comment. Will update

v2/pkg/engine/datasource/grpc_datasource/compiler.go
Noroth: Good catch. The intention was to grow the slice for all the values but it doesn't make sense in this case

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
Noroth: Yeah I removed `context` from the naming and instead refer to the actual directive name.

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
Noroth: This is only a fallback function when no directive was provided. I'm actually still wondering if we want to have a fallback or not. (This is being validated in protographic so we can't have invalid protobuf).

The expectation is that you always provide `@connect__fieldResolver` but if you omit it we would look for an `ID` field and use this.

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
Noroth: To shorten it. But all the values are pointers so I only copy the pointer address here.

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
Noroth: No I wouldn't do that. I'm using a dependency graph to have a topological sort, so basically the dependencies define the call IDs and it will provide me with any call that can be parallelized during any batch. So even if we have call id 9 and 10. they might come first.

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor.go
Noroth: I can do that in a separate PR because some tests currently rely on it.

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor_federation.go
Noroth: It basically defines your parent call. If a call uses this, the call depends on it.

v2/pkg/engine/datasource/grpc_datasource/grpc_datasource.go
Noroth: Some chunks depend on each other. So we first have to retrieve the data from the subgraph before we can provide the context data. But we are storing the information so basically all chunks that have can resolve their dependencies can be executed in parallel.

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor.go
Noroth: Yeah with the current logic I only select all the fields in the nested message of the resolved field. As discussed I will create a follow up PR to implement this as the PR is already quite huge.

v2/pkg/engine/datasource/grpc_datasource/json_builder.go
Noroth: added some comments

v2/pkg/engine/datasource/grpc_datasource/compiler.go
devsergiy: consider moving to the beginning of this method, as it will be visible that you check it

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor_federation.go
devsergiy: should it be parent call then?

PR #1331, opened 2025-10-23
feat: september 2025 spec description support for `OperationDocument` and `Fragment`

asoorm:
This PR implements description support for executable definitions according to the GraphQL September 2025 specification, enabling descriptions for both operations and fragments.

OperationDefinition
* Add Description field to OperationDefinition struct with helper methods
* Update parser to handle descriptions before query/mutation/subscription keywords
* Update printer to output operation descriptions

FragmentDefinition
* Add Description field to FragmentDefinition struct with helper methods
* Update parser to handle descriptions before fragment keyword
* Update printer to output fragment descriptions

## Summary by CodeRabbit

* **New Features**
  * Added support for parsing and handling descriptions on GraphQL operations (queries, mutations, subscriptions) and fragments. Descriptions are now preserved and properly rendered when outputting GraphQL documents, supporting both single-line and block-string description formats.

* **Tests**
  * Expanded test coverage for description parsing and printing across various operation and fragment scenarios.


v2/pkg/astprinter/astprinter.go
```diff
@@ -276,6 +276,11 @@ func (p *printVisitor) LeaveArgument(ref int) {
 
 func (p *printVisitor) EnterOperationDefinition(ref int) {
```
devsergiy, line 277: Could you add a test for printer?

v2/pkg/astparser/testdata/operation_with_description.graphql
```diff
@@ -0,0 +1,39 @@
+"""
```
devsergiy, line 1: do not see where this fixture is used

It is worth adding parsing this whole file, to make sure that each operation from multioperation document will get the description

devsergiy (changes requested):

v2/pkg/astparser/parser_test.go
```diff
@@ -1986,6 +1986,122 @@ func TestParser_Parse(t *testing.T) {
...
+					if !query.Description.IsDefined {
+						panic("want description to be defined")
+					}
+					description := doc.Input.ByteSliceString(query.Description.Content)
+					if description != "Fetches a user by ID" {
+						panic(fmt.Errorf("want 'Fetches a user by ID', got '%s'", description))
+					}
+					if doc.Input.ByteSliceString(query.Name) != "GetUser" {
+						panic("want GetUser")
+					}
```
ysmolski, line 2006: @devsergiy is this style of testing with panics a legacy thing?

ysmolski (approved):
LGTM, but wait for Sergiy's approval before the merge.

v2/pkg/astparser/parser_test.go
devsergiy: better to use testify for the new tests, but as the rest of the tests in this file use panics, we could leave it as is for now

PR #1332, opened 2025-10-23
fix: omit introspection queries from complexity limits

ysmolski:
This is disabled by default and should stay like that for now. There is a constant that can enable it again, but it is not intended to be tweaked by the clients.

## Summary by CodeRabbit

* **Bug Fixes**
  * Operation complexity now correctly skips GraphQL introspection fields when configured, improving calculation accuracy.

* **Tests**
  * Added tests covering node-skipping behavior and cases where skipped fields contribute zero to node count, complexity, and depth.

### DIFF
--- a/v2/pkg/middleware/operation_complexity/operation_complexity.go
+++ b/v2/pkg/middleware/operation_complexity/operation_complexity.go
@@ -20,7 +20,10 @@ To help the algorithm understand the schema make use of these two directives:
 should be used as a Node multiplier.
 
 "nodeCountSkip" indicates that the algorithm should skip this Node.
-It can be used to allowlist certain query paths, e.g. for introspection.
+It can be used to allowlist certain query paths.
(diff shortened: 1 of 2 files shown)


v2/pkg/middleware/operation_complexity/operation_complexity.go
```diff
@@ -45,6 +48,12 @@ var (
 	nodeCountSkip     = []byte("nodeCountSkip")
 )
 
+const (
+	skipIntrospection = true
+	dashSchema        = "__schema"
```
jensneuse: nit: doubleUnderscoreSchema, doubleUnderscoreType, OR simpler __schemaLiteral, __typeLiteral

ysmolski: thanks, you are right.

PR #1336, opened 2025-10-25
feat: improve memory management & request deduplication

jensneuse:
- memory management improved through arenas
- moved request deduplication from transport layer to engine
- UniqueRequestID is no longer needed, we pre-compute a hash for the headers outside of the Engine and can supply it via HeadersBuilder interface through the RequestContext to the engine

## Summary by CodeRabbit

* **New Features**
  * Global inbound and per‑subgraph single‑flight deduplication to reduce duplicate work.
  * Subscription triggers now include source metadata and propagate request headers to data sources.

* **Improvements**
  * Data sources and HTTP client now accept request headers and return response payloads directly.
  * Reduced allocations and improved performance via arena‑based pooling.
  * HTTP client: Content‑Length handling and per‑request size hints.
  * New header‑builder and context support for per‑subgraph headers.

* **Refactor**
  * Removed parallel‑list fetch path.

* **Chores**
  * Dependency updates to enable arena pooling.

### DIFF
--- a/v2/pkg/engine/resolve/context.go
+++ b/v2/pkg/engine/resolve/context.go
@@ -26,7 +26,7 @@ type Context struct {
 	// the resolver will use the new name to look up the old name to render the variable in the query.
 	RemapVariables map[string]string
 
-	VariablesHash uint64
+	VariablesHash    uint64
 	Files            []*httpclient.FileUpload
 	Request          Request
 	RenameTypeNames  []RenameTypeName


v2/pkg/engine/datasource/httpclient/nethttpclient.go
```diff
@@ -130,21 +130,32 @@ func respBodyReader(res *http.Response) (io.Reader, error) {
...
 
 	request, err := http.NewRequestWithContext(ctx, string(method), string(url), body)
 	if err != nil {
-		return err
+		return nil, err
+	}
+
+	if baseHeaders != nil {
+		request.Header = baseHeaders
 	}
```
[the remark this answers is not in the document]
jensneuse: baseHeaders are already cloned

v2/pkg/engine/resolve/loader.go
```diff
@@ -1643,11 +1573,96 @@
...
+	// Perform the actual load
+	err := l.loadByContextDirect(ctx, source, headers, input, res)
+	if err != nil {
+		item.err = err
+		return err
+	}
+
+	item.response = res.out
+	return nil
+}
```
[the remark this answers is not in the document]
jensneuse: no

v2/pkg/engine/resolve/resolve.go
```diff
@@ -233,9 +244,9 @@
 	return resolver
 }
 
-func newTools(options ResolverOptions, allowedExtensionFields map[string]struct{}, allowedErrorFields map[string]struct{}) *tools {
+func newTools(options ResolverOptions, allowedExtensionFields map[string]struct{}, allowedErrorFields map[string]struct{}, sf *SingleFlight) *tools {
 	return &tools{
-		resolvable: NewResolvable(options.ResolvableOptions),
+		resolvable: NewResolvable(nil, options.ResolvableOptions),
 		loader: &Loader{
 			propagateSubgraphErrors:                      options.PropagateSubgraphErrors,
```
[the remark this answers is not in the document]
jensneuse: passing nil as arena is ok

v2/pkg/engine/resolve/subgraph_request_singleflight.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	// If not exists, acquire a write lock to create the item
+	shard.mu.Lock()
+	// Double-check if the item was created while acquiring the write lock
+	item, exists = shard.items[sfKey]
+	if exists {
+		shard.mu.Unlock()
+		return sfKey, fetchKey, item, true
+	}
```
endigma: Is this second check worth having? these checks would be microseconds apart, the possible mutex contention and channel delay probably outweighs the performance benefit

if we wanted to wait, we might as well implement some kind of "waiting" function that has a more sensible interval to actually catch one, but wouldn't it be faster just to use this one to create the item and hope the later one picks up this key?

v2/pkg/engine/resolve/subgraph_request_singleflight.go
```diff
@@ -0,0 +1,192 @@
...
+type SubgraphRequestSingleFlight struct {
+	shards  []singleFlightShard
+	xxPool  *sync.Pool
+	cleanup chan func()
+}
+
+type singleFlightShard struct {
+	mu    sync.RWMutex
+	items map[uint64]*SingleFlightItem
+	sizes map[uint64]*fetchSize
```
endigma: Nit: If these are read-heavy in usage which is what it seems to be under load, sync.Map is probably more efficient

v2/pkg/engine/resolve/subgraph_request_singleflight.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/cespare/xxhash/v2"
+)
+
+// SubgraphRequestSingleFlight is a sharded, goroutine safe single flight implementation to de-duplicate subgraph requests
+// It's hashing the input and adds the pre-computed subgraph headers hash to avoid collisions
+// In addition to single flight, it provides size hints to create right-sized buffers for subgraph requests
+type SubgraphRequestSingleFlight struct {
+	shards  []singleFlightShard
+	xxPool  *sync.Pool
+	cleanup chan func()
```
endigma: This channel seems unused?

v2/pkg/engine/resolve/subgraph_request_singleflight.go
```diff
@@ -0,0 +1,192 @@
...
+	sizes map[uint64]*fetchSize
+}
+
+const defaultSingleFlightShardCount = 4
+
+// SingleFlightItem is used to communicate between leader and followers
+// If an Item for a key doesn't exist, the leader creates and followers can join
+type SingleFlightItem struct {
+	// loaded will be closed by the leader to indicate to followers when the work is done
+	loaded chan struct{}
```
endigma, line 30: When is this channel read from? I can't find it but I'm on GitHub not in an IDE

v2/pkg/engine/resolve/subgraph_request_singleflight.go
```diff
@@ -0,0 +1,192 @@
+package resolve
```
endigma, line 1: Things in this file seem self contained enough to possibly be moved into their own module, which I think would improve readability since the reader wouldn't have to scan so hard for anything being used in or from the rest of the "resolve" module.

v2/pkg/engine/resolve/inbound_request_singleflight.go
```diff
@@ -0,0 +1,138 @@
+package resolve
```
endigma, line 1: Maybe this can go with `subgraph_request_singleflight.go` into another module

v2/pkg/engine/resolve/arena_test.go
```diff
@@ -0,0 +1,261 @@
...
+
+	"github.com/wundergraph/go-arena"
+)
+
+func TestNewArenaPool(t *testing.T) {
+	pool := NewArenaPool()
+
+	require.NotNil(t, pool, "NewArenaPool returned nil")
+	assert.Equal(t, 0, len(pool.pool), "expected empty pool")
+	assert.Equal(t, 0, len(pool.sizes), "expected empty sizes map")
```
endigma: Nit: `assert.Len()` is recommended for this case by [testifylint](https://github.com/Antonboom/testifylint?tab=readme-ov-file#len).

v2/pkg/astvisitor/visitor.go
```diff
@@ -1391,6 +1398,11 @@ func (w *Walker) Walk(document, definition *ast.Document, report *operationrepor
 	} else {
 		w.Report = report
 	}
+	if w.arena == nil {
+		w.arena = arena.NewMonotonicArena(arena.WithMinBufferSize(64))
```
ysmolski, line 1402: Where does the magic number `64` come from? 
Does the arena being initilazied here and not in the `NewWalker` because of cases when we create a walker but never "walk" it? If it is not the case we could eliminate branching by moving initialization to the `NewWalker`

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
```diff
@@ -8866,10 +8864,10 @@ func TestLoadFiles(t *testing.T) {
 		input = httpclient.SetInputBodyWithPath(input, variables, "variables")
 		input = httpclient.SetInputBodyWithPath(input, query, "query")
 		input = httpclient.SetInputURL(input, []byte(serverUrl))
-		buf := bytes.NewBuffer(nil)
 
 		ctx := context.Background()
-		require.NoError(t, src.LoadWithFiles(ctx, input, []*httpclient.FileUpload{httpclient.NewFileUpload(f.Name(), fileName, "variables.file")}, buf))
+		_, err = src.LoadWithFiles(ctx, nil, input, []*httpclient.FileUpload{httpclient.NewFileUpload(f.Name(), fileName, "variables.file")})
+		require.NoError(t, err)
```
ysmolski:
```suggestion
		got, err := src.LoadWithFiles(ctx, nil, input, []*httpclient.FileUpload{httpclient.NewFileUpload(f.Name(), fileName, "variables.file")})
		require.NoError(t, err)
		require.Equal(t, []byte{}, got)

```

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource_test.go
```diff
@@ -8924,11 +8921,11 @@ func TestLoadFiles(t *testing.T) {
...
 
 		ctx := context.Background()
-		require.NoError(t, src.LoadWithFiles(ctx, input,
+		_, err = src.LoadWithFiles(ctx, nil, input,
 			[]*httpclient.FileUpload{
 				httpclient.NewFileUpload(f1.Name(), file1Name, "variables.files.0"),
-				httpclient.NewFileUpload(f2.Name(), file2Name, "variables.files.1")},
-			buf))
+				httpclient.NewFileUpload(f2.Name(), file2Name, "variables.files.1")})
+		require.NoError(t, err)
```
ysmolski:
```suggestion
		got, err := src.LoadWithFiles(ctx, nil, input,
			[]*httpclient.FileUpload{
				httpclient.NewFileUpload(f1.Name(), file1Name, "variables.files.0"),
				httpclient.NewFileUpload(f2.Name(), file2Name, "variables.files.1")})
		require.NoError(t, err)
		require.Equal(t, []byte{}, got)
```

v2/pkg/engine/datasource/grpc_datasource/grpc_datasource_test.go
```diff
@@ -149,12 +146,10 @@ func Test_DataSource_Load(t *testing.T) {
...
 	require.NoError(t, err)
 
-	output := new(bytes.Buffer)
-
-	err = ds.Load(context.Background(), []byte(`{"query":"`+query+`","variables":`+variables+`}`), output)
+	output, err := ds.Load(context.Background(), nil, []byte(`{"query":"`+query+`","variables":`+variables+`}`))
 	require.NoError(t, err)
 
-	fmt.Println(output.String())
+	fmt.Println(string(output))
```
ysmolski: This debugging stuff should be commented out.
```suggestion
	// fmt.Println(string(output))
```

v2/pkg/engine/datasource/grpc_datasource/json_builder.go
```diff
@@ -186,11 +189,10 @@ func (j *jsonBuilder) mergeValues(left *astjson.Value, right *astjson.Value) (*a
 // This function ensures that entities are placed in the correct positions in the final response
 // array based on their original representation order, which is critical for GraphQL federation.
 func (j *jsonBuilder) mergeEntities(left *astjson.Value, right *astjson.Value) (*astjson.Value, error) {
-	root := astjson.Arena{}
 
 	// Create the response structure with _entities array
-	entities := root.NewObject()
-	entities.Set(entityPath, root.NewArray())
+	entities := astjson.ObjectValue(j.jsonArena)
```
ysmolski, line 197: After the recent update to astjson, all kinds of the dynamic values should be created in the Arena, right?

v2/pkg/engine/datasource/httpclient/nethttpclient.go
```diff
@@ -293,14 +308,13 @@ func DoMultipartForm(
...
 		hasWrittenFileName = true
 
-		fmt.Fprintf(fileMap, `"%d":["%s"]`, i, file.variablePath)
+		_, _ = fmt.Fprintf(fileMap, `"%d":["%s"]`, i, file.variablePath)
 
 		key := fmt.Sprintf("%d", i)
-		_, _ = h.WriteString(file.Path())
 		temporaryFile, err := os.Open(file.Path())
 		tempFiles = append(tempFiles, temporaryFile)
 		if err != nil {
```
ysmolski: `defer` below only happens when there was no early return before. If for some file we return early, and some fiels were previsouly added to `tempFiles` then we will not remove close/remove those files. Also the if check should happen before adding the temporaryFile to the slice.

v2/pkg/engine/datasource/httpclient/nethttpclient.go
```diff
@@ -309,7 +323,7 @@ func DoMultipartForm(
 
 	multipartBody, contentType, err := multipartBytes(formValues, files)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	defer func() {
```
ysmolski, line 334: I guess this defer should be moved before the loop.

v2/pkg/engine/datasource/introspection_datasource/source_test.go
```diff
@@ -27,13 +27,18 @@ func TestSource_Load(t *testing.T) {
...
 
 			actualResponse := &bytes.Buffer{}
-			require.NoError(t, json.Indent(actualResponse, buf.Bytes(), "", "  "))
-			goldie.Assert(t, fixtureName, actualResponse.Bytes())
+			require.NoError(t, json.Indent(actualResponse, responseData, "", "  "))
+			// Trim the trailing newline that json.Indent adds
+			responseBytes := actualResponse.Bytes()
+			if len(responseBytes) > 0 && responseBytes[len(responseBytes)-1] == '\n' {
+				responseBytes = responseBytes[:len(responseBytes)-1]
+			}
```
ysmolski, line 40: 👍 
finally, we have it.

v2/pkg/engine/resolve/arena.go
```diff
@@ -0,0 +1,107 @@
+package resolve
```
ysmolski: What about moving this into separate module at the `v2/pkg/engine/` path, so that other parts of the engine might benefit from it? (Although I do not see immediate condidate to use it yet.) This pool is a good abstration that is not connected tightly to the "resolve". 

Or maybe even it could live in go-arena...

v2/pkg/engine/resolve/fetch.go
```diff
@@ -227,27 +226,6 @@ func (*EntityFetch) FetchKind() FetchKind {
 	return FetchKindEntity
 }
 
-// The ParallelListItemFetch can be used to make nested parallel fetches within a list
```
ysmolski, line 230: I kind of do not get why you removed it. It was not used or was performing poorly?

ysmolski (commented):
39/55 seen

v2/pkg/engine/resolve/loader.go
```diff
@@ -1840,7 +1897,10 @@ func (l *Loader) compactJSON(data []byte) ([]byte, error) {
 		return nil, err
 	}
 	out := dst.Bytes()
-	v, err := astjson.ParseBytesWithoutCache(out)
+	// don't use arena here or segfault
```
ysmolski:
```suggestion
	// Don't use arena here to avoid segfaults.
```

v2/pkg/engine/resolve/resolve.go
```diff
@@ -69,6 +73,20 @@ type Resolver struct {
 	heartbeatInterval time.Duration
 	// maxSubscriptionFetchTimeout defines the maximum time a subscription fetch can take before it is considered timed out
 	maxSubscriptionFetchTimeout time.Duration
+
+	// resolveArenaPool is the arena pool dedicated for Loader & Resolvable
+	// ArenaPool automatically adjusts arena buffer sizes per workload
+	// resolving & response buffering are very different tasks
+	// as such, it was best to have two arena pools in terms of memory usage
+	// A single pool for both was much less efficient
```
ysmolski: Personal nit. I would love to see punctuation here. Without dots and commas it is harder to read this text fast. I have to stop multiple times and re-read lines to identify sentences. It applies not only to this place but to other places with multi-line comments. 

```suggestion
	// resolveArenaPool is the arena pool dedicated for Loader & Resolvable.
	// ArenaPool automatically adjusts arena buffer sizes per workload.
	// Resolving & response buffering are very different tasks;
	// as such, it was best to have two arena pools in terms of memory usage.
	// A single pool for both was much less efficient.
```

v2/pkg/engine/resolve/loader.go
```diff
@@ -180,6 +171,19 @@ type Loader struct {
...
+	// Only use from main goroutine
+	// Don't Reset or Release, the Resolver handles this
+	// Disclaimer: When parsing json into the arena, the underlying bytes must also be allocated on the arena!
+	// This is very important to "tie" their lifecycles together
+	// If you're not doing this, you will see segfaults
+	// Example of correct usage in func "mergeResult"
+	jsonArena arena.Arena
+	// sf is the SubgraphRequestSingleFlight object shared across all client requests
+	// it's thread safe and can be used to de-duplicate subgraph requests
+	sf *SubgraphRequestSingleFlight
```
ysmolski: How about this name?

```suggestion
	jsonArena arena.Arena
	
	// singleFlight is the SubgraphRequestSingleFlight object shared across all client requests.
	// It's thread safe and can be used to de-duplicate subgraph requests.
	singleFlight *SubgraphRequestSingleFlight
```

v2/pkg/engine/resolve/subgraph_request_singleflight.go
ysmolski: it is being read in the loader.go

v2/pkg/engine/resolve/subgraph_request_singleflight.go
```diff
@@ -0,0 +1,192 @@
...
+	return s
+}
+
+// GetOrCreateItem generates a single flight key (100% identical fetches) and a fetchKey (similar fetches, collisions possible but unproblematic)
+// and return a SingleFlightItem as well as an indication if it's shared or not
+// If shared == false, the caller is a leader
+// If shared == true, the caller is a follower
+// item.sizeHint can be used to create an optimal buffer for the fetch in case of a leader
+// item.err must always be checked
+// item.response must never be mutated
```
ysmolski: Slight cosmetics fixes. Also why collisions are unproblematic? Is it because some other query will be waiting additional time until the leader is done, and that time is negligible? 

```suggestion
// GetOrCreateItem returns a single flight key (100% identical fetches), 
// a fetchKey (similar fetches, collisions possible but unproblematic),
// a SingleFlightItem, and an indication if it is shared or not.
// If not shared, the caller is a leader, otherwise it is a follower.
// item.sizeHint can be used to create an optimal buffer for the fetch in case of a leader.
// item.err must always be checked.
// item.response must never be mutated.
```

v2/pkg/engine/resolve/fetch.go
jensneuse: it's in fact not used at all, it's legacy and everyone else was afraid of deleting xD

v2/pkg/engine/resolve/subgraph_request_singleflight.go
jensneuse: that's worth a benchmark test

v2/pkg/astvisitor/visitor.go
jensneuse: We very rarely have situations where subgraph responses are smaller, so it's less buffer grow operations until we find the right buffer size, at the same time it's still small enough so that we're not over-allocating

v2/pkg/engine/resolve/subgraph_request_singleflight.go
jensneuse: that's correct, removed it

v2/pkg/engine/resolve/subgraph_request_singleflight.go
jensneuse: I clarified that fetchKey is only used for size hints

v2/pkg/engine/datasource/grpc_datasource/json_builder.go
jensneuse: yes, that's the case

ysmolski (approved):
I agree with the approach taken. The only big wish it to improbethe organization of the code mentioned in the comments.

v2/pkg/engine/resolve/subgraph_request_singleflight.go
jensneuse: I've simplified the code by using sync.Map

v2/pkg/engine/resolve/subgraph_request_singleflight.go
jensneuse: sync.Map is faster and the code is less verbose

v2/pkg/engine/resolve/subgraph_request_singleflight.go
jensneuse: this is not possible because we'd run into a cyclic dependency, unfortunately we've some tight coupling here, I could decouple it but I'm against as it would make the func calls more complex and currently I'm just passing around resolve.Context which makes things quite simple to handle

v2/pkg/engine/resolve/inbound_request_singleflight.go
jensneuse: same reason as this:
https://github.com/wundergraph/graphql-go-tools/pull/1336#discussion_r2537631714

v2/pkg/engine/resolve/arena_test.go
jensneuse: changed

v2/pkg/engine/resolve/arena.go
jensneuse: done

endigma (approved):

v2/pkg/engine/datasource/httpclient/nethttpclient.go
```diff
@@ -244,45 +269,46 @@ func makeHTTPRequest(client *http.Client, ctx context.Context, url, method, head
...
 	}
 
 	var tempFiles []*os.File
 
+	defer func() {
+		for _, file := range tempFiles {
+			if err := file.Close(); err != nil {
+				continue
+			}
+			if err = os.Remove(file.Name()); err != nil {
```
StarpTech: I suggest always removing the file regardless of whether it can be closed. A close error might not always point to a non-existent file.

```go
_ = file.Close()
_ = os.Remove(file.Name())
```

v2/pkg/engine/datasource/pubsub_datasource/pubsub_nats.go
```diff
@@ -77,40 +54,44 @@ type NatsPublishDataSource struct {
...
 	if err := s.pubSub.Publish(ctx, publishConfiguration); err != nil {
-		_, err = io.WriteString(out, `{"success": false}`)
-		return err
+		return []byte(`{"success": false}`), err
 	}
 
-	_, err = io.WriteString(out, `{"success": true}`)
-	return err
+	return []byte(`{"success": true}`), nil
 }
```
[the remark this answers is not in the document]
StarpTech: Makes sense @jensneuse

v2/pkg/engine/resolve/loader.go
```diff
@@ -1643,11 +1589,120 @@ func (l *Loader) setTracingInput(fetchItem *FetchItem, input []byte, trace *Data
...
+	}
+	info := fetchItem.Fetch.FetchInfo()
+	if info == nil {
+		return nil, 0
+	}
+	return l.ctx.HeadersForSubgraphRequest(info.DataSourceName)
+}
+
+// singleFlightAllowed returns true if the specific GraphQL Operation is a Query
+// even if the root operation type is a Mutation or Subscription
```
StarpTech, line 1647: What does that mean?

v2/pkg/engine/resolve/loader.go
```diff
@@ -1643,11 +1589,120 @@ func (l *Loader) setTracingInput(fetchItem *FetchItem, input []byte, trace *Data
...
+		}
+
+		res.out = item.response
+		return nil
+	}
+
+	// helps the http client to create buffers at the right size
+	ctx = httpclient.WithHTTPClientSizeHint(ctx, item.sizeHint)
+
+	defer l.singleFlight.Finish(item)
```
StarpTech, line 1707: Sure, that we finish it after possible returns?

v2/pkg/engine/resolve/loader.go
jensneuse: In GraphQL Federation, a Mutation can return an Entity, which will trigger a subgraph Query to fetch additional entity fields. In this case, single flight (request deduplication) is not allowed for the root but allowed for the sub query.

StarpTech: Got it. Thanks for the explanation.

v2/pkg/engine/resolve/resolve.go
```diff
@@ -1083,6 +1179,24 @@ func (r *Resolver) AsyncUnsubscribeClient(connectionID int64) error {
...
+		keyGen := pool.Hash64.Get()
+		_, _ = keyGen.Write(input)
+		var b [8]byte
+		binary.LittleEndian.PutUint64(b[:], headerHash)
+		_, _ = keyGen.Write(b[:])
+		triggerID = keyGen.Sum64()
+		pool.Hash64.Put(keyGen)
+		return header, triggerID
+	}
+	return nil, 0
```
dkorittki: Is it intentional not to return an error here? Effectively we cannot create a real trigger id because we lack a way to get headers and fallback to  0. That would mean if someone forgets to add a subgraph header builder to the context, it silently uses 0 as the trigger id. Could it happen that we add subscriptions to the same trigger 0 which should not be grouped together?

There are test cases in resolve_test.go which call `(Async)ResolveGraphQLSubscription` but the context they pass miss a header builder. The resulting trigger id is `0`.  If this behaviour is not intented they pass by accident because they are all designed to use the same trigger for their subscriptions. No tests actually verifies that multiple subscriptions to **different** triggers work.

PR #1341, opened 2025-10-31
feat: add composite type support for field resolvers

Noroth:
This PR improves the existing functionality for field resolvers.
Currently field resolvers do not support returning any form of composite types (interfaces and unions). This feature will be provided in this PR. 

The planner will now determine the correct selection set and identify, whether the field selections are part of an inline fragment. The calling logic will always inline all fragments provided to the datasource, therefore we do not need to handle fragments and spreads.

## Summary by CodeRabbit

* **New Features**
  * Resolver-driven execution planning with nested field-resolver support and enhanced composite-type handling; new GraphQL fields/types (product mascotRecommendation/stockStatus/productDetails, category mascot/categoryStatus/metrics, TestContainer/TestDetails, animal/owner/breed data) and matching RPCs.

* **Tests**
  * Large expansion of test coverage for nested resolvers, composite/interface/union scenarios, federation/entity lookups, execution plan generation, stack utilities, lookups, mappings, and nullable-field cases.

* **Chores**
  * Updated mock services, RPC mappings/protos, test helpers, and internal utilities to support the above.

## Checklist

### DIFF
--- a/v2/pkg/engine/datasource/grpc_datasource/grpc_datasource.go
+++ b/v2/pkg/engine/datasource/grpc_datasource/grpc_datasource.go
@@ -102,8 +102,6 @@ func (d *DataSource) Load(ctx context.Context, input []byte, out *bytes.Buffer) 
 	defer arena.Reset()
 	root := arena.NewObject()
 
-	failed := false
-
 	if err := d.graph.TopologicalSortResolve(func(nodes []FetchItem) error {
 		serviceCalls, err := d.rc.CompileFetches(d.graph, nodes, variables)
 		if err != nil {
@@ -149,7 +147,6 @@ func (d *DataSource) Load(ctx context.Context, input []byte, out *bytes.Buffer) 
 		}
 
 		if err := errGrp.Wait(); err != nil {
-			failed = true
 			return err
 		}
 
@@ -161,13 +158,12 @@ func (d *DataSource) Load(ctx context.Context, input []byte, out *bytes.Buffer) 
 				root, err = builder.mergeValues(root, result.response)
 			}
 			if err != nil {
-				out.Write(builder.writeErrorBytes(err))
 				return err
 			}
 		}
 
 		return nil
-	}); err != nil || failed {
+	}); err != nil {
 		out.Write(builder.writeErrorBytes(err))
 		return nil
 	}


v2/pkg/grpctest/mockservice_resolve.go
```diff
@@ -0,0 +1,772 @@
...
+		if int(threshold.GetValue()) > baseScore {
+			results = append(results, &productv1.ResolveCategoryPopularityScoreResult{
+				PopularityScore: nil,
+			})
+		} else {
+			results = append(results, &productv1.ResolveCategoryPopularityScoreResult{
+				PopularityScore: &wrapperspb.Int32Value{Value: int32(baseScore)},
+			})
+		}
+	}
```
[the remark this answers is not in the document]
Noroth: I think you're wrong here. Because generated proto types in Go will always check the receiver for null when using the functions.

v2/pkg/engine/datasource/grpc_datasource/execution_plan_field_resolvers_test.go
```diff
@@ -892,3 +892,1016 @@ func TestExecutionPlanFieldResolvers(t *testing.T) {
...
+																			JSONPath:      "barkVolume",
+																		},
+																	},
+																},
+															},
+														},
+													},
+												},
+											},
+										},
```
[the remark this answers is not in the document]
Noroth: This is actually correct. There are two different kinds of paths here. One is reading from a proto message which uses snake_cased fields, the other path is for the JSON response which uses camel case.

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor_federation.go
```diff
@@ -389,6 +399,35 @@ func (r *rpcPlanVisitorFederation) LeaveField(ref int) {
...
+		return
+	}
+
+	r.resolvedFields = append(r.resolvedFields, resolvedField)
+	r.fieldResolverAncestors.push(len(r.resolvedFields) - 1)
+	r.fieldPath = r.fieldPath.WithFieldNameItem(r.operation.FieldNameBytes(ref))
+
+	// In case of nested fields with arguments, we need to increment the related call ID.
+	r.parentCallID++
+}
```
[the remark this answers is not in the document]
Noroth: currently a field resolver must have arguments. We do not yet support defining them without

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -895,33 +1036,153 @@ type resolveRPCCallConfig struct {
...
+		field, err := r.buildRequiredField(parentTypeNode, fieldRef)
+		if err != nil {
+			return nil, err
+		}
+
+		message.Fields = append(message.Fields, field)
+	}
+
+	message.Fields = slices.Clip(message.Fields)
 	return message, nil
```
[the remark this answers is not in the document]
Noroth: This is actually correct. This function will be called if the type that is returned from a resolved field is a complex type. This can either be a simple Graphql object type or a composite type like and interface or a union. In those cases we need to make a difference. Because a normal selection set doesn't require any additional information and the composite selection set needs to provide additional information like the inline fragment condition type. 

We can't have both in this case.

v2/pkg/grpctest/mockservice_resolve.go
```diff
@@ -0,0 +1,772 @@
...
+
+	for _, ctx := range req.GetContext() {
+		// Calculate normalized score: (value / baseline) * 100
+		// This gives a percentage relative to the baseline
+		normalizedScore := (ctx.GetValue() / baseline) * 100.0
+
+		results = append(results, &productv1.ResolveCategoryMetricsNormalizedScoreResult{
+			NormalizedScore: normalizedScore,
+		})
+	}
```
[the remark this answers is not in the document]
Noroth: I doubt that

```
func (x *ResolveCategoryMetricsNormalizedScoreRequest) GetFieldArgs() *ResolveCategoryMetricsNormalizedScoreArgs {
	if x != nil {
		return x.FieldArgs
	}
	return nil
}
```

v2/pkg/grpctest/util.go
```diff
@@ -0,0 +1,149 @@
...
+			},
+		}
+	}
+
+	return &productv1.ListOfListOfUser{
+		List: &productv1.ListOfListOfUser_List{
+			Items: results,
+		},
+	}
+}
```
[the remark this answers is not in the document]
Noroth: Improved it with the use of the generated protobuf getters

v2/pkg/grpctest/util.go
```diff
@@ -0,0 +1,149 @@
...
+			},
+		}
+	}
+
+	return &productv1.ListOfListOfCategory{
+		List: &productv1.ListOfListOfCategory_List{
+			Items: results,
+		},
+	}
+}
```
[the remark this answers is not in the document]
Noroth: Improved it with the use of the generated protobuf getters

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor_federation.go
```diff
@@ -187,9 +187,29 @@ func (r *rpcPlanVisitorFederation) EnterSelectionSet(ref int) {
...
+			resolvedField := &r.resolvedFields[resolvedFieldAncestor]
+
+			resolvedField.memberTypes = memberTypes
+			resolvedField.fieldsSelectionSetRef = ast.InvalidRef
+
+			r.planCtx.enterResolverCompositeSelectionSet(compositType, ref, resolvedField)
+			return
+		}
+
+		// TODO: handle nested resolved fields.
```
SkArchon: I'm assuming its a separate PR, just highlighting in case its not the case

v2/pkg/engine/datasource/grpc_datasource/util.go
```diff
@@ -8,3 +8,42 @@ func initializeSlice[T any](len int, zero T) []T {
 	}
 	return s
 }
+
+// stack is a generic LIFO (Last In First Out) data structure that stores elements of type T.
+type stack[T any] []T
```
SkArchon, line 13: Is there a better place that we could place this with more "visibility" / "more general" rather than `grpc_datasource/util.go‎`? This is under the assumption we would want other folks to not create another "stack" type again.

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -718,6 +716,77 @@ func (r *rpcPlanningContext) buildInputMessageField(typeRef int, fieldName, json
 	return field, nil
 }
 
+func (r *rpcPlanningContext) buildFieldMessage(fieldTypeNode ast.Node, fieldRef int) (*RPCMessage, error) {
+	field := r.operation.Fields[fieldRef]
+	if !field.HasSelections {
+		return nil, nil
+	}
+
+	fieldRefs := make([]int, 0)
```
jensneuse: why not init with size 0?

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -727,20 +796,92 @@ func (r *rpcPlanningContext) resolveServiceName(subgraphName string) string {
...
+// isFieldResolver checks if a field is a field resolver.
+func (r *rpcPlanningContext) isFieldResolver(fieldRef int, isRootField bool) bool {
+	if isRootField {
+		return false
+	}
+
+	return len(r.operation.FieldArguments(fieldRef)) > 0
+}
+
+// getCompositeType returns the most recent composite type from the type definitions.
```
ysmolski: Not sure about "the most recent composite type from the type definitions". It is somehow sounds off considering the body the function that just checks `.Kind` field.

PS: I cannot do multiline comments in this file. GH has bug.

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -890,38 +1031,159 @@ func (r *rpcPlanningContext) filterIDFieldsFunc(o ast.ObjectTypeDefinition, fiel
...
+
+			field.Message = message
+		}
+
+		result = append(result, field)
+	}
+	return result, nil
+}
+
+func (r *rpcPlanningContext) fieldDefinitionForType(fieldName, typeName string) (ref int, exists bool) {
```
ysmolski: `exists` is kind of redundant here. I guess this design follows from the redundant `exists` in this function:
```go
func (d *Document) NodeFieldDefinitionByName(node Node, fieldName ByteSlice) (definition int, exists bool) {
	for _, i := range d.NodeFieldDefinitions(node) {
		if bytes.Equal(d.Input.ByteSlice(d.FieldDefinitions[i].Name), fieldName) {
			return i, true
		}
	}
	return InvalidRef, false
}
```
It would be much more simple if we just returned the InvalidRef without redundant boolean value in both functions. This redundancy seems to be contagious. ...Unless there is a use case I am not aware about.

I remember adding similar function to the visitor not long ago. It looks like there is a repeating pattern that maybe moved to the ast.Document as a method. It is up to you if you want to do it in this PR.

```go
// fieldDefinitionRef returns the definition reference of a field in a given type or ast.InvalidRef if not found.
func (v *Visitor) fieldDefinitionRef(typeName string, fieldName string) int {
	node, ok := v.Definition.NodeByNameStr(typeName)
	if !ok {
		return ast.InvalidRef
	}
	defRef, ok := v.Definition.NodeFieldDefinitionByName(node, []byte(fieldName))
	if !ok {
		return ast.InvalidRef
	}
	return defRef
}
```

v2/pkg/engine/datasource/grpc_datasource/util.go
ysmolski: I am not sure about this wrapper. I cannot rationalize introducing this abstraction, except for having a little be nicer names essentially the same functionality as a simple slice. I think accessing slice with indexing operation is simple enough that there is no need to wrap them.

For example, this method:
```go
// peek returns the top element of the stack without removing it.
// Note: This function will panic if called on an empty stack.
func (a *stack[T]) peek() T {
	return (*a)[len(*a)-1]
}
```
It hides potential panic, thus being obscure, introduces accidental complexity.

Noroth: I changed it because I thought reading large appends and removing indices sometimes and sometimes not was way harder to read than having functions that actually sound like that. I personally think it's way more readable than what I had before

ysmolski: I see. Then we could keep it as experiment in grpd_datasrouce and observer for now.

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
Noroth: Good point. I think the function looked differently when I first implemented it and I never updated the comment.

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
Noroth: I removed exists

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -718,6 +716,77 @@ func (r *rpcPlanningContext) buildInputMessageField(typeRef int, fieldName, json
...
+
+		if message.FieldSelectionSet == nil {
+			message.FieldSelectionSet = make(RPCFieldSelectionSet)
+		}
+
+		message.FieldSelectionSet.Add(typeName, fields...)
+	}
+
+	for _, fieldRef := range fieldRefs {
+		field, err := r.buildRequiredField(fieldTypeNode, fieldRef)
```
ysmolski, line 786: Could we encounter nested field resolvers here? We do not check it here like we do in buildFieldResolverTypeMessage:
```go
func (r *rpcPlanningContext) buildFieldResolverTypeMessage(typeName string, resolverField *resolverField) (*RPCMessage, error) {
    // ...
    for _, fieldRef := range fieldRefs {
        if r.isFieldResolver(fieldRef, false) {
            continue  
        }
        // ...
    }
}
```
Let's say we have this:
```graphql
interface SearchResult {
  id: ID!
  relevance(algorithm: String!): Float!  # Field resolver
}

type Product implements SearchResult {
  id: ID!
  relevance(algorithm: String!): Float!
  name: String!
}

query {
  search {
    ... on Product {
      name
      relevance(algorithm: "v2")  # This will be included in the message. Is it okay?
    }
  }
}
```

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -890,38 +1032,164 @@ func (r *rpcPlanningContext) filterIDFieldsFunc(o ast.ObjectTypeDefinition, fiel
...
+	}
+
+	// If the field is a message type and has selections, we need to build a nested message.
+	if field.ProtoTypeName == DataTypeMessage && r.operation.FieldHasSelections(fieldRef) {
+		fieldTypeNode, found := r.definition.ResolveNodeFromTypeRef(r.definition.FieldDefinitionType(fieldDef))
+		if !found {
+			return RPCField{}, fmt.Errorf("unable to build required field: unable to resolve field type node for field %s", fieldName)
+		}
+
+		message, err := r.buildFieldMessage(fieldTypeNode, fieldRef)
```
ysmolski, line 1139: Both `buildRequiredField` and `buildCompositeField` call `buildFieldMessage` recursively. And `buildFieldMessage` calls back to `buildRequiredField` and `buildCompositeField`. Is it possible to get infinite recursion in this chain?

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -890,38 +1032,164 @@ func (r *rpcPlanningContext) filterIDFieldsFunc(o ast.ObjectTypeDefinition, fiel
...
+
+		field, err := r.buildField(inlineFragmentNode, fieldDef, r.operation.FieldNameString(fieldRef), "")
+		if err != nil {
+			return nil, err
+		}
+
+		if field.ProtoTypeName == DataTypeMessage && r.operation.FieldHasSelections(fieldRef) {
+			fieldTypeNode, found := r.definition.ResolveNodeFromTypeRef(r.definition.FieldDefinitionType(fieldDef))
+			if !found {
+				return nil, fmt.Errorf("unable to build required field: unable to resolve field type node for field %s", r.operation.FieldNameString(fieldRef))
```
ysmolski: The error message says "unable to build required field" but this is in buildCompositeField

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor.go
```diff
@@ -405,7 +409,7 @@ func (r *rpcPlanVisitor) EnterField(ref int) {
 // LeaveField implements astvisitor.FieldVisitor.
 func (r *rpcPlanVisitor) LeaveField(ref int) {
 	r.fieldPath = r.fieldPath.RemoveLastItem()
-	r.resolvedFieldIndex = ast.InvalidRef
+	r.fieldResolverAncestors.pop()
```
ysmolski: For all fields that we leave, we pop the item. But we add to the stack only from "enterFieldResolver". Maybe we should pop only when we leave fieldResolver fields? Current behavior violates the basic stack invariant.
Additionally (or more like compensating maybe), we do SkipNode when we are inside of fieldResolver, that makes LeaveField never called:
```
	if r.fieldResolverAncestors.len() > 0 {
		// We don't want to call LeaveField here because we ignore the field entirely.
		r.walker.SkipNode()
		return
	}
```

It is hard to understand/reason what is going with stack because of this.

I decided to figure out how it works anyway.

```graphql
{ user { posts(limit: 10) { title } } }
```

1. EnterField(user) - regular field, no push
2. LeaveField(user) - pop() called (but stack is empty, no-op)
3. EnterField(posts) - field resolver, push (stack: [0])
4. EnterField(title) - inside field resolver, SkipNode() called
5. LeaveField(title) is SKIPPED
6. LeaveField(posts) - pop() called (stack: [])

It means:
- `user` doesn't push but tries to pop (no-op because empty)
- `posts` pushes and pops (correct)
- `title` doesn't push and doesn't pop (because LeaveField is skipped)

What will happen if we have nested field resolvers?

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -890,38 +1032,164 @@ func (r *rpcPlanningContext) filterIDFieldsFunc(o ast.ObjectTypeDefinition, fiel
...
+		if !found {
+			return RPCField{}, fmt.Errorf("unable to build required field: unable to resolve field type node for field %s", fieldName)
+		}
+
+		message, err := r.buildFieldMessage(fieldTypeNode, fieldRef)
+		if err != nil {
+			return RPCField{}, err
+		}
+
+		field.Message = message
```
ysmolski, line 1144: this could be nil. is it okay?

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -890,38 +1032,164 @@ func (r *rpcPlanningContext) filterIDFieldsFunc(o ast.ObjectTypeDefinition, fiel
...
+			if !found {
+				return nil, fmt.Errorf("unable to build required field: unable to resolve field type node for field %s", r.operation.FieldNameString(fieldRef))
+			}
+
+			message, err := r.buildFieldMessage(fieldTypeNode, fieldRef)
+			if err != nil {
+				return nil, err
+			}
+
+			field.Message = message
```
ysmolski, line 1182: this could be nil. is it okay?

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -890,38 +1032,164 @@ func (r *rpcPlanningContext) filterIDFieldsFunc(o ast.ObjectTypeDefinition, fiel
...
+				return nil, err
+			}
+
+			message.FieldSelectionSet[fragmentSelection.typeName] = fields
+		}
+
+		return message, nil
 	}
+
+	if resolverField.fieldsSelectionSetRef == ast.InvalidRef {
```
ysmolski, line 1086: What does this represent? When is it used? This should be documented.

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor.go
```diff
@@ -72,15 +72,15 @@ type rpcPlanVisitorConfig struct {
...
-		resolvedFieldIndex: ast.InvalidRef,
-		fieldPath:          make(ast.Path, 0),
+		walker:                 &walker,
+		plan:                   &RPCExecutionPlan{},
+		subgraphName:           cases.Title(language.Und, cases.NoLower).String(config.subgraphName),
+		mapping:                config.mapping,
+		operationFieldRef:      ast.InvalidRef,
+		resolverFields:         make([]resolverField, 0),
+		parentCallID:           ast.InvalidRef,
+		fieldResolverAncestors: newStack[int](0),
```
ysmolski, line 82: Capacity is 0. Does it mean by default we expect no field resolvers? So statistically it is rare to se them and additional allocations are worth it?

Noroth: This is something that will change in the future. Currently we can't really assume it

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
Noroth: Good point. This should not happen. I updated it and added a test

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
Noroth: We're going recursively on the selection sets. So these should end at some point.

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor.go
Noroth:
```
EnterField(user) - regular field, no push
LeaveField(user) - pop() called (but stack is empty, no-op)
EnterField(posts) - field resolver, push (stack: [0])
EnterField(title) - inside field resolver, SkipNode() called
LeaveField(title) is SKIPPED
LeaveField(posts) - pop() called (stack: [])
```

This logic is not quite correct. I'll break it down a bit more

```graphql
query { user { posts(limit: 10) { title } } }

<- EnterDocument
query A <- Enter Operation Definition
{  <- EnterSelectionSet
  user <- EnterField("user") { <- EnterSelectionSet 
    posts <- EnterField("posts") (limit: 10) <- EnterArgument/LeaveArgument  
      {  <- EnterSelectionSet
        title  <- EnterField("title"), LeaveField("title")
       } <- LeaveSelectionSet
<- LeaveField("posts")
   }  <- LeaveSelectionSet
<- LeaveField("user")
} <-LeaveSelectionSet
<-LeaveDocument
```

To summarize it. We need to make sure when we are traversing inside of a field resolver, that we retrieve all the nested resolvers from the root resolver. At the same time we ignore all fields that are not field resolvers.

The information about field resolvers will be stored separately. We need to create additional RPC calls that need to be invoked in order, therefore we look for nested fields that are field resolvers. To prevent the path from being incorrect we skip LeaveField for all non resolved fields. As for nested fields wie will continue to call EnterField before LeaveField will be called.

imagine:
```
fieldResolver1 -> push stack[0] -> EnterSelection 
  -> field1(non-resolver) skip 
  -> fieldResolver2(nested fields) push stack[1] -> EnterSelection 
    -> fieldResolver3(resolver - no nested fields) push stack[2] 
    -> LeaveField(fieldResolver3) pop stack[1] 
  -> LeaveField(fieldResolver2) pop stack[0] 
-> LeaveField(fieldResolver1) pop stack = empty
```

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor.go
ysmolski: If you are okay with this, then me too :)

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
ysmolski: Yeah, I guess resolvability depends on the visitor and the structure of ast. So it should be fine. I could not think of any example to break it...

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
ysmolski: Thank you!

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor.go
ysmolski: So we ignore all fields that are not fields resolvers there. That means that leaveField only happens for fieldResolvers.

Could you put your summary into comments to accompany EnterField and SkipNode? I think this is not very trivial to understand from the first glance at the code, so some explanation will be useful there later.

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
Noroth: Well technically this cannot be nil, because the surrounding loop logic validates that the field has selections.
But you're right on one point - the `buildFieldMessage` should rather return an error if we don't have any selections. Currently I think as the function is only called by other functions that already validate that the field has selections the additional check in `buildFieldMessage` doesn't really do anything.

ysmolski (approved):
LGTM

v2/pkg/engine/datasource/grpc_datasource/execution_plan_test.go
```diff
@@ -102,7 +102,7 @@ func TestQueryExecutionPlans(t *testing.T) {
 	}{
 		{
 			name:    "Should include typename when requested",
-			query:   `query UsersWithTypename { users { __typename id name } }`,
+			query:   `query UsersWithTypename { users { __typename id __typename name } }`,
```
ysmolski, line 105: is double `__typename` intended?

v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -890,38 +1044,176 @@ func (r *rpcPlanningContext) filterIDFieldsFunc(o ast.ObjectTypeDefinition, fiel
...
+		if err != nil {
+			return nil, err
+		}
+
+		message.Fields = append(message.Fields, field)
 	}
+
+	message.Fields = slices.Clip(message.Fields)
 	return message, nil
 }
```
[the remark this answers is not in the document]
Noroth: This is actually fine because in GraphQL we can have fields that do not have to be declared in an inline fragment. E.g. __typename or fields that are defined in an interface.
Therefore these fields will be outside of the selection set of the inline fragment and need to be handled

v2/pkg/engine/datasource/grpc_datasource/execution_plan_test.go
Noroth: Ah yes this is intended. Just making sure that it doesn't matter where the typename is specified.

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor.go
```diff
@@ -211,8 +211,6 @@ func (r *rpcPlanVisitor) EnterSelectionSet(ref int) {
 			}
 			resolvedField := &r.resolverFields[resolverFieldAncestor]
 			resolvedField.memberTypes = memberTypes
-			resolvedField.fieldsSelectionSetRef = ast.InvalidRef
```
dkorittki: Technically this means `resolvedField.fieldsSelectionSetRef` is now 0 instad of -1 but doesn't this logically mean we declare this fields selection set reference as valid whereas before it was invalid? Was it a bug?

v2/pkg/engine/datasource/grpc_datasource/grpc_datasource.go
```diff
@@ -169,7 +168,8 @@ func (d *DataSource) Load(ctx context.Context, input []byte, out *bytes.Buffer)
 
 		return nil
 	}); err != nil || failed {
-		return err
+		out.Write(builder.writeErrorBytes(err))
```
dkorittki, line 167: There are other places inside that if condition which, too, write to `out` in error conditions. So this would lead to double writes to `out` with probably the same error. Probably not intended right?

v2/pkg/engine/datasource/grpc_datasource/execution_plan_visitor.go
Noroth: This is actually fine. I make sure in `enterResolverCompositeSelectionSet` to always handle the `fieldSelectionSetRef`.

v2/pkg/engine/datasource/grpc_datasource/grpc_datasource.go
Noroth: Good catch. We should write the error bytes only once.

PR #1342, opened 2025-11-10
fix: add ability to exclude introspection queries from complexity limits

ysmolski:
Allow external users to configure if they want to skip intospection querues in the complexity limits.

Follow-up in Cosmo: https://github.com/wundergraph/cosmo/pull/2296

## Summary by CodeRabbit

* **New Features**
  * Per-operation option to exclude GraphQL introspection fields from complexity calculations for finer-grained analysis.

* **Breaking Changes**
  * Removed the legacy built-in complexity calculator and the direct request-level complexity call; callers must use the new estimator-based complexity pathway.

* **Tests**
  * Updated and expanded complexity tests to cover introspection-included and introspection-skipped scenarios and to validate the estimator workflow.

### DIFF
--- a/v2/pkg/middleware/operation_complexity/operation_complexity_test.go
+++ b/v2/pkg/middleware/operation_complexity/operation_complexity_test.go
@@ -506,9 +506,7 @@ func runConfig(t *testing.T, definition, operation string, expectedGlobalComplex
 
 	estimator := NewOperationComplexityEstimator(skipIntrospection)
 	actualGlobalComplexityResult, actualFieldsComplexityResult := estimator.Do(&op, &def, &report)
-	if report.HasErrors() {
-		require.NoError(t, report)
-	}
+	require.False(t, report.HasErrors())
 
 	assert.Equal(t, expectedGlobalComplexityResult.NodeCount, actualGlobalComplexityResult.NodeCount, "unexpected global node count")
 	assert.Equal(t, expectedGlobalComplexityResult.Complexity, actualGlobalComplexityResult.Complexity, "unexpected global complexity")


execution/graphql/request_test.go
```diff
@@ -90,7 +90,7 @@ func TestRequest_parseQueryOnce(t *testing.T) {
 func TestRequest_CalculateComplexity(t *testing.T) {
 	t.Run("should return error when schema is nil", func(t *testing.T) {
 		request := Request{}
-		result, err := request.CalculateComplexity(DefaultComplexityCalculator, nil)
+		result, err := request.CalculateComplexity(DefaultComplexityCalculator, nil, false)
```
jensneuse: could we work the "skipIntrospection" flag into the DefaultComplexityCalculator as an option or create a CustomComplexityCalculator that has flags? we could use the options pattern

-not a hard opinion, more like thoughts

v2/pkg/middleware/operation_complexity/operation_complexity.go
```diff
@@ -202,7 +204,7 @@ func (c *complexityVisitor) EnterField(ref int) {
 	}
 
 	typeName, fieldName, alias := c.extractFieldRelatedNames(ref, definition)
```
jensneuse, line 207: Is this making string copies? if yes we should avoid it when used in the hot path

ysmolski: It uses unsafe `unsafebytes.BytesToString` underneath

execution/graphql/request_test.go
ysmolski: Your idea has instigated some research that lead to removal of the file containing `DefaultComplexityCalculator`. We did not need that file because it was used in testing only and I just used `operation_complexity` directly.

jensneuse (approved):

ysmolski:
@coderabbitai why i cannot merge this pr?

PR #1350, opened 2025-12-05
feat: allow empty arguments for field resolvers

Noroth:
# Introduction
This PR introduces support to call field resolvers without any arguments. Currently we only check if field resolvers provide arguments with parentheses.

Field resolvers were determined whether fields have arguments. This is technically correct but instead of checking field arguments on the operation side, they have to be determined from the schema definition.

## Prerequisites

The following problem statement refer to this example schema

```graphql
type Query {
  categories: [Category!]!
}

type Category {
  id: ID!
  # The argument is nullable and can be omitted
  relevantTopic(orderBy: String): Topic! 
}

type Topic {
  id: ID!
  description: String!
} 
```

## Problem Statement

This will correctly call the field resolver
```graphql
query {
  categories {
    id
    relevantTopic("") {
      id
      description
    }
  }
}
```

This did not call the field resolver
```graphql
query {
  categories {
    id
    relevantTopic {
      id
      description
    }
  }
}
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling and validation for field resolution
  * Improved support for optional arguments in field resolvers
  * Better error messages when field definitions cannot be resolved

* **Tests**
  * Expanded test coverage for field resolver edge cases

## Checklist

### DIFF
--- a/v2/pkg/engine/datasource/grpc_datasource/grpc_datasource_test.go
+++ b/v2/pkg/engine/datasource/grpc_datasource/grpc_datasource_test.go
@@ -4539,6 +4539,33 @@ func Test_Datasource_Load_WithFieldResolvers(t *testing.T) {
 				require.Empty(t, errData)
 			},
 		},
+		{
+			name:  "Query with field resolver without parentheses (nullable parameter)",
+			query: "query CategoriesWithFieldResolverNoParens { categories { id name popularityScore } }",
+			vars:  `{"variables":{}}`,
+			validate: func(t *testing.T, data map[string]interface{}) {
+				require.NotEmpty(t, data)
+
+				categories, ok := data["categories"].([]interface{})
+				require.True(t, ok, "categories should be an array")
+				require.NotEmpty(t, categories, "categories should not be empty")
+				require.Len(t, categories, 4, "Should return 4 categories")
+
+				for _, category := range categories {
+					category, ok := category.(map[string]interface{})
+					require.True(t, ok, "category should be an object")
+					require.NotEmpty(t, category["id"])
+					require.NotEmpty(t, category["name"])
+					// popularityScore should return 50 when called without threshold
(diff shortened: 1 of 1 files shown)


v2/pkg/engine/datasource/grpc_datasource/execution_plan.go
```diff
@@ -779,11 +779,16 @@ func (r *rpcPlanningContext) buildFieldMessage(fieldTypeNode ast.Node, fieldRef
 	}
 
 	for _, fieldRef := range fieldRefs {
-		if r.isFieldResolver(fieldRef, false) {
+		fieldDefRef, found := r.definition.NodeFieldDefinitionByName(fieldTypeNode, r.operation.FieldNameBytes(fieldRef))
```
ysmolski, line 782: Not an issue in your PR, but rather how inconsitently we deal with not found references in the engine. "found" is totally redundant in some cases. In the case when it is returned from the cache wrapped, it might not be so (did not look there). I am not sure if we should address it here.
Because if that redundant "found" we have redundant check later: `|| fieldDefRef == ast.InvalidRef`

```go
func (d *Document) NodeFieldDefinitionByName(node Node, fieldName ByteSlice) (definition int, exists bool) {
	for _, i := range d.NodeFieldDefinitions(node) {
		if bytes.Equal(d.Input.ByteSlice(d.FieldDefinitions[i].Name), fieldName) {
			return i, true
		}
	}
	return InvalidRef, false
}
```

devsergiy: not sure that it is always redundant, `isFieldResolver` is called in multiple places, sometimes the code is not calling `NodeFieldDefinitionByName`

devsergiy: ah, I rechecked locally code from this PR, yeah, you are right
when we call isFieldResolver we already checked in all places that ref is valid

PR #1351, opened 2025-12-08
fix: group subgraph errors by subgraph name

SkArchon:
Right now we pass in all subgraph errors to the engine loader hook OnFinished function ([See Here](https://github.com/wundergraph/cosmo/blob/24c064853f11f0c157c5e1a8bfa289d4f73d4c50/router/core/engine_loader_hooks.go#L118)), which means that whenever any subgraph has an error, we will pass that/those errors to the OnFinished function for an irrelevant Subgraph, which leads to the following
* Incorrect status information attached to that subgraph's Engine Loader span ([See Here](https://github.com/wundergraph/cosmo/blob/24c064853f11f0c157c5e1a8bfa289d4f73d4c50/router/core/engine_loader_hooks.go#L231))
* Access logs incorrect status used ([See Here](https://github.com/wundergraph/cosmo/blob/24c064853f11f0c157c5e1a8bfa289d4f73d4c50/router/core/engine_loader_hooks.go#L224))

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Error reporting now attributes failures to specific data sources, yielding clearer, source-specific error messages and deterministic ordering when combined.
  * Aggregated error output is suppressed when no errors are present, reducing noise.

* **Tests**
  * Test suites updated to validate per-source error attribution, message details, ordering, and context initialization across authorization, rate-limit, and fetch-failure scenarios.

### DIFF
--- a/v2/pkg/engine/resolve/context.go
+++ b/v2/pkg/engine/resolve/context.go
@@ -14,6 +14,7 @@ import (
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/datasource/httpclient"
 )
 
+// Context should not ever be initialized directly, and should be initialized via the NewContext function
 type Context struct {
 	ctx              context.Context
 	Variables        *astjson.Value


v2/pkg/engine/resolve/context.go
```diff
@@ -138,11 +139,29 @@ func (c *Context) SetRateLimiter(limiter RateLimiter) {
...
+	for _, k := range keys {
+		joined = errors.Join(joined, c.subgraphErrors[k])
+	}
+	return joined
 }
 
-func (c *Context) appendSubgraphErrors(errs ...error) {
-	c.subgraphErrors = errors.Join(c.subgraphErrors, errors.Join(errs...))
+func (c *Context) appendSubgraphErrors(ds DataSourceInfo, errs ...error) {
+	if c.subgraphErrors == nil {
```
SkArchon: The context is used extensively in the router, even initialized, thus I was a bit defensive and added a check here to be safe as the context is initialized without using the NewContext method

ysmolski: Not sure about this defensive check. If something initialize context or nullify the `subgraphErrors` then we just hide potential bug. In other words, this defensive nil check suggests there might be code paths where Context is created without using NewContext(), or the map could be set to nil elsewhere.

v2/pkg/engine/resolve/context.go
```diff
@@ -138,11 +139,29 @@ func (c *Context) SetRateLimiter(limiter RateLimiter) {
 }
 
 func (c *Context) SubgraphErrors() error {
```
ysmolski, line 142: Is this the only place how router sees errors? How do we expose errors to router in general? I am asking because this interface has not changed, we just return all the errors...

SkArchon: The router calls this function in two places, in these cases it does not care about which errors belong to which subgraph, thus I maintained the same behaviour of returning all the errors.

ysmolski: Is this method the only way how router gets the errors?

v2/pkg/engine/resolve/context.go
SkArchon: Hi yes exactly, you are correct, we create the Context in the router without using `NewContext()`, these should be refactored I think, but I would not like to do it with this task.

I believe its safe to remove the defensive check, as it seems to be only used in the engine (its private at the moment), my only concern is simply future proofing it, if someone decides to use `subgraphErrors` without using NewContext().

v2/pkg/engine/resolve/context.go
SkArchon: I could have added a separate slice where we store errors just for these cases, but felt a bit excessive.

v2/pkg/engine/resolve/context.go
ysmolski: I don't understand. In the router we do not use NewContext(). That means that we need initialize the map somehow. If we remove defensive check, then what will initialize the map?

v2/pkg/engine/resolve/context.go
ysmolski: Nevermind. I found the other place where `subgraphErrors` is exposed:
```
func newResponseInfo(res *result, subgraphErrors map[string]error) *ResponseInfo {
	responseInfo := &ResponseInfo{
		StatusCode:   res.statusCode,
		Err:          subgraphErrors[res.ds.Name],
		responseBody: res.out,
	}
```

v2/pkg/engine/resolve/context.go
```diff
@@ -138,11 +139,26 @@ func (c *Context) SetRateLimiter(limiter RateLimiter) {
...
+	for k := range c.subgraphErrors {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	var joined error
+	for _, k := range keys {
+		joined = errors.Join(joined, c.subgraphErrors[k])
+	}
+	return joined
```
devsergiy, line 158: Should we put this functionality under the flag, as the joined output will slightly change?

v2/pkg/engine/resolve/loader_test.go
```diff
@@ -285,7 +285,8 @@ func TestLoader_LoadGraphQLResponseData(t *testing.T) {
 		},
 	}
 	ctx := &Context{
-		ctx: context.Background(),
+		ctx:            context.Background(),
+		subgraphErrors: make(map[string]error),
```
ysmolski: We could use NewContext() here and below.

v2/pkg/engine/resolve/ratelimit_test.go
```diff
@@ -55,7 +55,7 @@ func TestRateLimiter(t *testing.T) {
 
 		res := generateTestFederationGraphQLResponse(t, ctrl)
 
-		return res, &Context{ctx: context.Background(), Variables: nil, rateLimiter: limiter, RateLimitOptions: RateLimitOptions{Enable: true}},
+		return res, &Context{ctx: context.Background(), Variables: nil, rateLimiter: limiter, RateLimitOptions: RateLimitOptions{Enable: true}, subgraphErrors: make(map[string]error)},
```
ysmolski: Here we do not use NewContext and paying some verbosity price too. I am not sure that it is okay to initilize `subgraphErrors` like that in every test. We might not use construct Context directly...

v2/pkg/engine/resolve/context.go
```diff
@@ -138,11 +139,26 @@ func (c *Context) SetRateLimiter(limiter RateLimiter) {
...
 }
 
-func (c *Context) appendSubgraphErrors(errs ...error) {
-	c.subgraphErrors = errors.Join(c.subgraphErrors, errors.Join(errs...))
+func (c *Context) appendSubgraphErrors(ds DataSourceInfo, errs ...error) {
+	c.subgraphErrors[ds.Name] = errors.Join(c.subgraphErrors[ds.Name], errors.Join(errs...))
 }
```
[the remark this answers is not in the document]
SkArchon: Please recheck

v2/pkg/engine/resolve/context.go
SkArchon: As discussed this should be ok

v2/pkg/engine/resolve/context.go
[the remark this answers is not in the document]
SkArchon: Future code should note use `Context{}` directly.

v2/pkg/engine/resolve/context.go
SkArchon: As discussed we. changed the router to use NewContext()

ysmolski (approved):
LGTM, thank you!

v2/pkg/engine/resolve/context.go
ysmolski, line 18: One small request. Please add documentation here that refers to the `NewContext()`, to discourage use of plain `Context{}`

PR #1373, opened 2026-02-02
feat: add field argument mapping

dkorittki:
## Checklist

- [x] Provide ability to disable field arg mapping
- [x] Wait for https://github.com/wundergraph/graphql-go-tools/pull/1336 to be merged first, then basing this branch onto those changes

## Context

Extends the engines variable normalizer to map field arguments as a side product of it's actual work and return the results alongside the file upload mapping.

This adds a useful feature which the router uses, called field argument mapping. It's described in detail in the [corresponding router pull request](https://github.com/wundergraph/cosmo/pull/2385).

The principle of field argument mapping is to allow Cosmo Streams hook users to be able to access values of field arguments to filter subscription events based on this. The idea is to give access the values via operation variables.
Values of any field arguments are available via variables, either because the users already use variables or because they have been extracted from literal values via variable extraction. The map that is being build holds the path to the field argument as a dot notated string and the variable name where the argument value is hold. Later on, when hook users want to access the value, the variable name can be resolved from the argument path and we can provide the variable value.

Here's an example:

```graphql
query GetUser($userId: ID!) {
    user(id: $userId) {
        name
        orders(limit:12) {
            id
        }
    }
}`,
```

The map thats being build is:
```
map[argumentPath]variableName
"query.user.id", "userId"
"query.user.orders.limit", "a"
```

If a hook user later accesses `query.user.orders.limit` we resolve the value of variable `a` to get `12`.  
If the hook accesses `query.user.id` we can resolve `userId`, even if the engine performed variable remapping (`userId` being renamed to `a` or `b`, etc.)
This works because the variables which later are accessed are the ones from the operation context from the router and those are unaffected of variable remapping. In other words: The router config option `engine.disable_variables_remapping` does not have an effect on this. It works regardless.

The first implementation of this feature was done solely on the router. It worked without engine changes. A newly created walker would walk the operation AST and collect and map all field arguments. While this worked the problem was the introduction of a new walker during the request processing right in the hot path and it wasn't cachable either.

With this pull request the actual walking is done by an **already existing visitor** as a side product: `variablesExtractionVisitor`. This visitor is part of the variable normalizer and **already walks field arguments**. I just placed the new functionality into the visitors `EnterArgument` hook.

The variable normalizer does it's work by invoking the `NormalizeOperation` method. This method, in it's current form, returns the file upload mapping, which is cached by the router. This method has been modified to extend the return type by the field arg mapping results. Since the router already cached the result we get caching field arg mapping for free.

Field Argument Mapping can be enabled or disabled when the variable normalizer is instanciated.

```go
func NewVariablesNormalizer(withFieldArgMapping bool) *VariablesNormalizer
```

This option is then passed down to the `variablesExtractionVisitor`, where the field argument mapping is happening.
The same visitor exists for the operation normalizer. There, however, we disable field argument mapping because it's a useful feature only on the variable normalizer.

## Performance Considerations

This code hooks into an already existing walker, which walks all field arguments of an operation. The code added will allocate a string out of a slice byte via string convertion `(string(myByteSlice)` for most arguments. Also it will repeatedly grow a map, because the map initially is not allocated with a predetermined size. This could have performance implications on queries with vast amounts of field arguments but I think in reality we don't have a lot of them. Also once it's cached by the router, it does not matter anyway.

## Summary by CodeRabbit

* **Refactor**
  * Variable normalization now supports an option to enable field-argument mapping and returns a combined result with upload mappings and optional field-argument mappings; default mapping is off unless enabled.
  * Path string formatting was standardized to include inline fragment/type information in more contexts, affecting path rendering in messages and lookups.
* **Tests**
  * Expanded and updated suites to cover field-argument mapping scenarios and the new structured normalization result.


v2/pkg/ast/path.go
```diff
@@ -128,7 +128,7 @@ func (p Path) String() string {
 	return out
 }
 
-func (p Path) DotDelimitedString() string {
+func (p Path) DotDelimitedString(useFragmentRefs bool) string {
```
Noroth: I think it might be better to have a second function here. 

Add the new logic to a function that accepts a parameter, but keep the existing `DotDelimitedString` function to just call the other function with `true`.

For someone who doesn't know about this change and no docs this might be a bit confusing. Also bool parameters have the tendency to end up as code smell therefore I personally prefer functions with proper names and descriptions.

Noroth: Could also be a good time to add some tests for this. I think the path tests are missing in general.

v2/pkg/astnormalization/astnormalization.go
```diff
@@ -353,7 +353,7 @@ type VariablesNormalizer struct {
 	variablesExtractionVisitor *variablesExtractionVisitor
 }
 
-func NewVariablesNormalizer() *VariablesNormalizer {
+func NewVariablesNormalizer(withFieldArgMapping bool) *VariablesNormalizer {
```
Noroth: We should start documenting these exported functions when we make changes there.

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -782,7 +782,7 @@ func (p *Planner[T]) allowField(ref int) bool {
 	// In addition, we skip field if its path are equal to planner parent path
 	// This is required to correctly plan on datasource which has corresponding child/root node,
 	// but we don't need to add it to the query as we are in the nested request
-	currentPath := fmt.Sprintf("%s.%s", p.visitor.Walker.Path.DotDelimitedString(), fieldAliasOrName)
+	currentPath := fmt.Sprintf("%s.%s", p.visitor.Walker.Path.DotDelimitedString(true), fieldAliasOrName)
```
Noroth: we could remove `fmt` here because it's quite a lot lower than string concatenation.

we're already doing it properly in `representation_variable.go`

```go
currentPath := v.Walker.Path.DotDelimitedString(true) + "." + string(fieldName)
```

v2/pkg/engine/plan/datasource_filter_collect_nodes_visitor.go
```diff
@@ -592,9 +592,9 @@ func (f *treeBuilderVisitor) collectFieldInfo(fieldRef int) {
...
 	fieldAliasOrName := f.operation.FieldAliasOrNameString(fieldRef)
 	isTypeName := fieldName == typeNameField
-	parentPath := f.walker.Path.DotDelimitedString()
+	parentPath := f.walker.Path.DotDelimitedString(true)
 	onFragment := f.walker.Path.EndsWithFragment()
-	parentPathWithoutFragment := f.walker.Path.WithoutInlineFragmentNames().DotDelimitedString()
+	parentPathWithoutFragment := f.walker.Path.WithoutInlineFragmentNames().DotDelimitedString(true)
 
 	currentPath := fmt.Sprintf("%s.%s", parentPath, fieldAliasOrName)
 	currentPathWithoutFragments := fmt.Sprintf("%s.%s", parentPathWithoutFragment, fieldAliasOrName)
```
Noroth: same here. Also I feel like there is a lot of unnecessary copying in this function into the final result with the local variables. Doesn't matter too much but this could be optimized?

v2/pkg/ast/path.go
devsergiy: +1 

please introduce a new function

v2/pkg/engine/plan/datasource_filter_collect_nodes_visitor.go
dkorittki: I fixed the string concat. In case of variable assignments between line 592 and 597 I feel like they add to readability. The compiler should optimize it away anyway. I ran the compiler with escape analysis to see if they escape to heap and they don't. Only their values do but that's what we need to call anyway. The worst thing which could happen is they are stored on the stack but that's still super fast and shouldn't be noticable.

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
dkorittki: Fixed

v2/pkg/astnormalization/astnormalization.go
dkorittki: Fixed

v2/pkg/ast/path.go
dkorittki: Added tests in a new file path_test.go

dkorittki: Fixed

v2/pkg/astnormalization/variables_extraction.go
```diff
@@ -141,12 +160,57 @@ func (v *variablesExtractionVisitor) EnterArgument(ref int) {
...
+	}
 }
 
 func (v *variablesExtractionVisitor) EnterDocument(operation, definition *ast.Document) {
 	v.operation, v.definition = operation, definition
 	v.extractedVariables = v.extractedVariables[:0]
 	v.extractedVariableTypeRefs = v.extractedVariableTypeRefs[:0]
+	if v.fieldArgumentMapping.enabled {
+		v.fieldArgumentMapping.result = make(FieldArgumentMapping)
+	}
```
Noroth, line 176: nit: you could check the number of expected arguments in the v.operation document to pass the size to `make`?

Noroth (approved):

v2/pkg/astnormalization/astnormalization.go
```diff
@@ -382,22 +385,32 @@ func NewVariablesNormalizer() *VariablesNormalizer {
 	}
 }
 
-func (v *VariablesNormalizer) NormalizeOperation(operation, definition *ast.Document, report *operationreport.Report) []uploads.UploadPathMapping {
+// NormalizeOperation processes GraphQL operation variables.
+// It detects and removes unused variables, extracts variables from inline values
+// and coerces variable types. It modifies the operation in place and
```
devsergiy: it coerses not types, but variable values, we could also mention what part of graphql spec it is https://spec.graphql.org/September2025/#sec-Coercing-Variable-Values

v2/pkg/astnormalization/astnormalization.go
```diff
@@ -382,22 +385,32 @@ func NewVariablesNormalizer() *VariablesNormalizer {
...
+// returns metadata including field argument mappings and upload paths.
+// Field argument mapping is done only when v is configured to do so via NewVariablesNormalizer,
```
devsergiy: >when v is configured
when VariablesNormalizer

or when collecting arguments mapping enabled

v2/pkg/astnormalization/astnormalization_test.go
```diff
@@ -1060,7 +1061,330 @@ func TestVariablesNormalizer(t *testing.T) {
...
+			}
+		`
+
+		testCases := []struct {
+			name            string
+			operation       string
+			variables       string
+			expectedMapping FieldArgumentMapping
+		}{
+			{
```
devsergiy, line 1119: missing test case for complex input type, smth like this

```
type Query {
   users(paging: Paging!): [String]!
}

input Paging {
  limit: Int!
  offset: Int!
}
```

such query
```
query Search($limit: Int!, $offset: Int!) {
  users(paging: {limit: $limit, offset: $offset})
}
variables: {"limit": 10, "offset": 5}
```

will be normalized into
```
query Search($a: Paging!) {
  users(paging: $a)
}
variables: {"limit": 10, "offset": 5, "a": {"limit": 10, "offset": 5}}
```

v2/pkg/astnormalization/variables_extraction.go
```diff
@@ -141,12 +160,57 @@ func (v *variablesExtractionVisitor) EnterArgument(ref int) {
...
+}
+
+// recordFieldArgumentMapping records the currently visited field argument
+// of v in v.fieldArgumentMapping, alongside its matching variable name or literal value.
+// If varName is empty, it looks up the variable name from the operation or stores the literal value.
+func (v *variablesExtractionVisitor) recordFieldArgumentMapping(ref int, varName string) {
+	// Guard to prevent nil panics.
```
devsergiy, line 183: when it is enabled we should always have value

potentially we could just put a check here
```
if v.fieldArgumentMapping.enabled {
```

v2/pkg/astnormalization/variables_extraction.go
```diff
@@ -141,12 +160,57 @@ func (v *variablesExtractionVisitor) EnterArgument(ref int) {
...
+	if varName != "" {
+		// Variable name was provided (from extraction)
+		v.fieldArgumentMapping.result[key] = varName
+		return
+	}
+
+	if v.operation.Arguments[ref].Value.Kind != ast.ValueKindVariable {
+		// We expect the operation on this visitor to be normalized before the visitor walks it.
+		// This means that values of any field arguments should be extracted to variables.
+		// If we still land here it means the query hasn't been normalized before.
```
devsergiy: > If we still land here it means the query hasn't been normalized before.

a bit strange comment, because this visitor is doing this normalization

v2/pkg/astnormalization/field_argument_mapping.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// FieldArgumentMapping maps the path of field arguments inside an operation
+// to the name of their mapped variables.
+// Key: "rootOperationType.fieldPath.argumentName" (e.g., "query.posts.limit")
+// Value: name of variable of field argument after variable normalization.
+type FieldArgumentMapping map[string]string
+
+// VariablesNormalizerResult contains the results of variable normalization.
+type VariablesNormalizerResult struct {
```
devsergiy: what do you think about `VariablesNormalizationResult` name?

I propose to move VariablesNormalizer into its own file `variables_normalizer` or `variables_normalization` and add these types there

v2/pkg/astnormalization/astnormalization.go
```diff
@@ -353,7 +353,10 @@ type VariablesNormalizer struct {
 	variablesExtractionVisitor *variablesExtractionVisitor
 }
 
-func NewVariablesNormalizer() *VariablesNormalizer {
+// NewVariablesNormalizer creates a new variable normalizer.
+// If withFieldArgMapping is true then the normalizer creates a
+// mapping of field arguments as a "side product" when NormalizeOperation is called.
+func NewVariablesNormalizer(withFieldArgMapping bool) *VariablesNormalizer {
```
devsergiy: what do you think about introducing either an options or config argument?

e.g.
```
type VariablesNormalizerOptions struct {
  withFieldArgMapping bool
}

WithFieldArgMapping()
```

or just
```
type VariablesNormalizerConfig struct {
  WithFieldArgMapping bool
}
```

devsergiy (changes requested):

v2/pkg/astnormalization/astnormalization.go
dkorittki: This is what "when v is configured to do so via NewVariablesNormalizer" wanted to express. I'll rephrase it.

v2/pkg/astnormalization/variables_extraction.go
dkorittki: The relationship between `v.fieldArgumentMapping.enabled == true` and `v.fieldArgumentMapping.result != nil` is loose. It's enforced in `EnterDocument`. It could change by accident in the future and hypothetically we could end up with `v.fieldArgumentMapping.enabled == true` and `v.fieldArgumentMapping.result == nil`. Then `if v.fieldArgumentMapping.enabled` would not return early and we would run into a nil panic.

I did it this way to be maximum defensive. This function wants to work with `v.fieldArgumentMapping.result` so it must exist.

Will change the comments to make this intend more clear

v2/pkg/astnormalization/astnormalization.go
dkorittki: Fixed

v2/pkg/astnormalization/field_argument_mapping.go
dkorittki: Done

v2/pkg/astnormalization/variables_extraction.go
dkorittki: Fixed

v2/pkg/astnormalization/variables_extraction.go
dkorittki: Done

v2/pkg/astnormalization/astnormalization_test.go
dkorittki: Done

v2/pkg/astnormalization/astnormalization.go
dkorittki: Done

v2/pkg/astnormalization/astnormalization.go
dkorittki: Done

PR #1374, opened 2026-02-02
feat: subscriptions overhaul

endigma:
## Summary

Rewrites the subscription client from scratch with the following goals:

- Multiplexing support
- Cleaner separation between upstream/downstream connections
- Protocol correctness and testability

### New `subscriptionclient` package

`v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/`

Separates protocols, transports, and client logic:

- **`protocol/`**: `graphql-ws` and `graphql-transport-ws` behind a shared `Protocol` interface
- **`transport/`**: WebSocket and SSE behind a shared `Transport` interface
- **`client.go`**: entry point that selects the appropriate transport and protocol

`graphql_subscription_client.go` is now a thin adapter over this package.

### Resolver simplification

`v2/pkg/engine/resolve/`

The old clients used netpoll/event-loop patterns that required separate `AsyncSubscriptionDataSource` and `SubscriptionDataSource` interfaces. The new client is goroutine-based, so the two interfaces collapse into one. `AsyncSubscriptionDataSource` is removed along with its channels and workers, replaced by mutex-guarded state and a single heartbeat loop.

### `Close(kind)` replaced with `Error(data)` on response writer

`v2/pkg/engine/resolve/response.go`

The resolver no longer closes downstream connections directly. It forwards failures as-is and lets the downstream handler decide what to do.

---

## Summary by CodeRabbit

* **Chores**
  * Updated module dependencies and added project tooling/config files.

* **New Features**
  * Added a redesigned, option-driven subscription client with pluggable WS/SSE transports, protocol implementations, and runtime stats.

* **Refactor**
  * Reworked subscription APIs and resolver lifecycle for synchronous cleanup and simplified subscription handling.

* **Tests**
  * Expanded and reorganized subscription test coverage; added focused transport/protocol unit tests and removed legacy handlers/tests.

* **Bug Fixes**
  * Improved upstream error reporting, heartbeat, connection lifecycle and cleanup behavior.

## Checklist


.serena/project.yml
```diff
@@ -0,0 +1,152 @@
+# the name by which the project can be referenced within Serena
+project_name: "graphql-go-tools"
```
Noroth: Are we now committing user specific AI tool configs?

execution/engine/config_factory_federation.go
```diff
@@ -458,16 +458,16 @@ func (f *FederationEngineConfigFactory) subscriptionClient(
...
 	default:
 		// for compatibility reasons we fall back to graphql-ws protocol
 		graphqlSubscriptionClient = subscriptionClientFactory.NewSubscriptionClient(
-			httpClient,
-			streamingClient,
 			f.engineCtx,
+			graphql_datasource.WithUpgradeClient(httpClient),
+			graphql_datasource.WithStreamingClient(streamingClient),
 		)
 	}
```
Noroth: I'm either absolutely confused here, or both cases are the exact same code and the switch is unnecessary

execution/graphql/result_writer.go
```diff
@@ -39,7 +39,7 @@ func (e *EngineResultWriter) Heartbeat() error {
 	return nil
 }
 
-func (e *EngineResultWriter) Close(_ resolve.SubscriptionCloseKind) {
+func (e *EngineResultWriter) Error(_ []byte) {
```
Noroth: For future people taking a look at this code. Can we get some comments on the functions and the type. This will help understand what this writer is used for.

I mean for the whole file. We need to start documenting areas that are missing documentation.
Empty `Error` functions look like mocking but this doesn't look like a test writer.

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/common/message.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+)
+
+var ErrConnectionClosed = errors.New("connection closed")
+
+// MessageType identifies the kind of message delivered on a subscription channel.
+type MessageType int
```
Noroth: could be uint8

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/common/message.go
```diff
@@ -0,0 +1,47 @@
...
+	Errors     json.RawMessage `json:"errors,omitempty"`
+	Extensions json.RawMessage `json:"extensions,omitempty"`
+}
+
+type Request struct {
+	Query         string          `json:"query"`
+	OperationName string          `json:"operationName,omitempty"`
+	Variables     json.RawMessage `json:"variables,omitempty"`
+	Extensions    json.RawMessage `json:"extensions,omitempty"`
+}
```
Noroth, line 50: Add some doc comments

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/common/options.go
```diff
@@ -0,0 +1,55 @@
...
+
+type WSSubprotocol string
+
+const (
+	SubprotocolAuto               WSSubprotocol = ""                     // Auto, negotiated with the server
+	SubprotocolGraphQLTransportWS WSSubprotocol = "graphql-transport-ws" // Modern subprotocol
+	SubprotocolGraphQLWS          WSSubprotocol = "graphql-ws"           // Legacy subprotocol, deprecated
+)
+
+func (s WSSubprotocol) Subprotocols() []string {
```
Noroth, line 25: Doc comment would be good here as well.

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/graphql_transport_ws.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+type incomingMessage struct {
+	ID      string          `json:"id,omitempty"`
+	Type    string          `json:"type"`
+	Payload json.RawMessage `json:"payload,omitempty"`
+}
+
+type GraphQLTransportWS struct {
+	AckTimeout time.Duration
```
Noroth: Is this supposed to be exported? This could be changed during runtime.

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/graphql_transport_ws.go
```diff
@@ -0,0 +1,174 @@
...
+
+// Init implements Protocol.
+func (p *GraphQLTransportWS) Init(ctx context.Context, conn *websocket.Conn, payload map[string]any) error {
+	initMsg := outgoingMessage{
+		Type: gtwsTypeConnectionInit,
+	}
+	if payload != nil {
+		initMsg.Payload = payload
+	}
+	timeout := p.AckTimeout
```
Noroth: AckTimeout could technically be negative and you wouldn't apply you default timeout

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/graphql_transport_ws.go
```diff
@@ -0,0 +1,174 @@
...
+		case gtwsTypePing:
+			if err := p.Pong(ackCtx, conn); err != nil {
+				return fmt.Errorf("pre-init pong: %w", err)
+			}
+			continue
+		default:
+			return fmt.Errorf("%w: got %q", ErrAckNotReceived, ackMessage.Type)
+		}
+	}
+}
```
Noroth: Do you provide any docs for the protocol implementations for other people to look up?

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/graphql_ws.go
```diff
@@ -0,0 +1,175 @@
...
+	initMsg := outgoingMessage{
+		Type: gwsTypeConnectionInit,
+	}
+	if payload != nil {
+		initMsg.Payload = payload
+	}
+	timeout := p.AckTimeout
+	if timeout == 0 {
+		timeout = 30 * time.Second
+	}
```
Noroth: Same here, could be negative

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/protocol.go
```diff
@@ -0,0 +1,81 @@
...
+// MessageType identifies the message type.
+type MessageType int
+
+const (
+	MessageData MessageType = iota
+	MessageError
+	MessageComplete
+	MessagePing
+	MessagePong
+)
```
Noroth: We have a message type ion the common package and in the protocol package. This could be a bit confusing. Should we adjust the naming a bit to be able to differentiate between them?

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/protocol.go
```diff
@@ -0,0 +1,81 @@
...
+	case MessageComplete:
+		return "complete"
+	case MessagePing:
+		return "ping"
+	case MessagePong:
+		return "pong"
+	default:
+		return "unknown"
+	}
+}
```
Noroth: You could use stringer. We already use it in other places as well

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/protocol.go
```diff
@@ -0,0 +1,81 @@
...
+		return &common.Message{Type: common.MessageTypeConnectionError, Err: m.Err}
+	case MessageComplete:
+		return &common.Message{Type: common.MessageTypeComplete}
+	default:
+		return &common.Message{Type: common.MessageTypeUnknown}
+	}
+}
+
+// MessageType identifies the message type.
+type MessageType int
```
Noroth: could also be uint8?

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/sse_transport.go
```diff
@@ -0,0 +1,266 @@
...
+	t.log.Debug("sseTransport.Subscribe",
+		abstractlogger.String("endpoint", opts.Endpoint),
+		abstractlogger.String("method", string(method)),
+	)
+
+	requestCtx, requestCancel := context.WithCancel(context.WithoutCancel(ctx))
+
+	// Attach cancel to transport context
+	context.AfterFunc(t.ctx, requestCancel)
+	context.AfterFunc(ctx, requestCancel)
```
Noroth: Someone will look at this in a couple of months and ask what is happening here :)

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/client.go
```diff
@@ -0,0 +1,97 @@
...
+// Subscribe creates a new upstream via the appropriate transport.
+func (c *Client) Subscribe(ctx context.Context, req *common.Request, opts common.Options, handler common.Handler) (func(), error) {
+	if c.ctx.Err() != nil {
+		return nil, ErrClientClosed
+	}
+
+	if opts.Transport == common.TransportSSE {
+		return c.sse.Subscribe(ctx, req, opts, handler)
+	}
+	return c.ws.Subscribe(ctx, req, opts, handler)
```
Noroth: Would it make sense to check if opts.Transport is an invalid string and use a switch? Or is it okay to always fallback to c.ws?

v2/pkg/engine/resolve/resolve.go
```diff
@@ -617,10 +584,13 @@ func (r *Resolver) executeSubscriptionUpdate(resolveCtx *Context, sub *sub, shar
...
 
 	if err := sub.writer.Flush(); err != nil {
+		sub.writeMu.Unlock()
 		// If flush fails (e.g. client disconnected), remove the subscription.
-		_ = r.AsyncUnsubscribeSubscription(sub.id)
+		_ = r.UnsubscribeSubscription(sub.id)
 		return
 	}
+	sub.lastWriteTime.Store(time.Now().UnixNano())
+	sub.writeMu.Unlock()
```
Noroth, line 682: The mutex locking logic here looks a bit error prone. I trust your judgement if you say there is no other way. But I personally would try to keep mutex logic in smaller sub functions because it might happen in the future that we make changes and end up deadlocking because the mutex logic might be a bit hard to understand

v2/pkg/engine/resolve/resolve.go
```diff
@@ -806,17 +720,16 @@ func (r *Resolver) handleAddSubscription(triggerID uint64, add *addSubscription)
...
-			},
-			final: false,
-		}
-		// After the startup hooks are executed, we can add the subscription to the subscriptions registry
-		// so that it can start receive events
-		trig.subscriptions[add.ctx] = s
+		// Register first so startup hooks can deliver initial data via UpdateSubscription.
+		trig.mu.Lock()
+		trig.subscriptions[add.id] = s
+		trig.mu.Unlock()
```
Noroth: maybe create a receiver function for that. You use it multiple times and then you don't need to always make sure you properly handle the mutex.

Also you can defer it in case we have a panic

v2/pkg/engine/resolve/resolve.go
```diff
@@ -827,175 +740,314 @@ func (r *Resolver) handleAddSubscription(triggerID uint64, add *addSubscription)
...
+		subscriptions: make(map[SubscriptionIdentifier]*subscriptionState),
+		updateBuf:     bytes.NewBuffer(make([]byte, 0, 1024)),
 		cancel:        cancel,
 		updater:       updater,
 	}
 	r.triggers[triggerID] = trig
-	trig.subscriptions[add.ctx] = s
+	trig.mu.Lock()
+	trig.subscriptions[add.id] = s
+	trig.mu.Unlock()
```
Noroth: same case as above

v2/pkg/engine/resolve/resolve.go
```diff
@@ -1395,30 +1415,32 @@ func (r *Resolver) AsyncResolveGraphQLSubscription(ctx *Context, subscription *G
...
+	if err := ctx.ctx.Err(); err != nil {
+		return err
+	}
+
 	headers, triggerID := r.prepareTrigger(ctx, subscription.Trigger.SourceName, input)
 
-	select {
-	case <-r.ctx.Done():
-		// Stop resolving if the resolver is shutting down
+	r.mu.Lock()
```
Noroth: I think this unlock can be `deferred`

.serena/project.yml
```diff
@@ -0,0 +1,152 @@
+# the name by which the project can be referenced within Serena
```
ysmolski: Is this file accidentally slipped into the PR?

Noroth (changes requested):

.serena/project.yml
endigma: was debating this with Jens, it's not technically user specific? just a tool not everyone uses. I'm open to ignoring it or committing it

execution/engine/config_factory_federation.go
endigma: This looks like the switch is probably unnecessary, as it was before my diff.

execution/graphql/result_writer.go
endigma: will do, i think they're documented on the interface this fulfills though

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/graphql_transport_ws.go
endigma: I can link them, sure

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/protocol.go
endigma: It could be confusing yeah, the protocol one still has things like ID in it because it's not fully de-protocoled yet for use with a handler

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/protocol.go
endigma: stringer?

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/sse_transport.go
endigma: ill clear it up, it's correct but not entirely clear

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/client.go
endigma: could be more defensive, sure

v2/pkg/engine/resolve/resolve.go
endigma: I'll see if I can move writeMu locking into the functions that need it

v2/pkg/engine/resolve/resolve.go
endigma: I'll look, could also use a syncmap, probably faster

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/common/options.go
```diff
@@ -0,0 +1,55 @@
...
+
+type SSEMethod string
+
+const (
+	SSEMethodAuto SSEMethod = ""     // Auto: POST for graphql-sse (default)
+	SSEMethodPOST SSEMethod = "POST" // POST with JSON body (graphql-sse spec)
+	SSEMethodGET  SSEMethod = "GET"  // GET with query parameters (traditional SSE)
+)
+
+type Options struct {
```
ysmolski, line 46: I would appreciate more doc comments

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/protocol.go
Noroth: https://pkg.go.dev/golang.org/x/tools/cmd/stringer

v2/pkg/engine/resolve/resolve.go
```diff
@@ -48,19 +49,19 @@ type AsyncErrorWriter interface {
...
 	maxConcurrency chan struct{}
 
-	triggers         map[uint64]*trigger
-	events           chan subscriptionEvent
-	triggerUpdateBuf *bytes.Buffer
+	mu                        sync.Mutex
+	shutdown                  bool
+	triggers                  map[uint64]*trigger
+	subscriptionsByID         map[SubscriptionIdentifier]*subscriptionState
+	subscriptionsByConnection map[int64]map[SubscriptionIdentifier]*subscriptionState
```
ysmolski, line 73: what are uint64 and int64 here? I can figure out `int64` from the field name though. Maybe it would make sense to introduce some type aliases for readability? 
`subscriptionsByConnection` is also a complex type. a comment explaining its nature would be good (why nested map)

v2/pkg/engine/resolve/resolve.go
```diff
@@ -437,131 +441,78 @@ func (r *Resolver) ArenaResolveGraphQLResponse(ctx *Context, response *GraphQLRe
 }
 
 type trigger struct {
+	mu            sync.RWMutex
```
ysmolski, line 458: what fields does this mutex protect?

v2/pkg/engine/resolve/resolve.go
```diff
@@ -437,131 +441,78 @@ func (r *Resolver) ArenaResolveGraphQLResponse(ctx *Context, response *GraphQLRe
...
-// workItem is used to encapsulate a function that needs to be
-// executed in the worker goroutine. fn will be executed, and if
-// final is true the worker will be stopped after fn is executed.
-type workItem struct {
-	fn    func()
-	final bool
-}
-
-type sub struct {
+type subscriptionState struct {
```
ysmolski, line 543: complex struct, some high level commen is wanted

v2/pkg/engine/resolve/resolve.go
```diff
@@ -437,131 +441,78 @@ func (r *Resolver) ArenaResolveGraphQLResponse(ctx *Context, response *GraphQLRe
...
 	resolve   *GraphQLSubscription
-	resolver  *Resolver
 	ctx       *Context
 	writer    SubscriptionResponseWriter
 	id        SubscriptionIdentifier
 	heartbeat bool
 	completed chan struct{}
-	// workChan is used to send work to the writer goroutine. All work is processed sequentially.
-	workChan chan workItem
+	writeMu   sync.Mutex
```
ysmolski: writes to what does it protect?

v2/pkg/engine/resolve/resolve.go
```diff
@@ -827,175 +740,314 @@ func (r *Resolver) handleAddSubscription(triggerID uint64, add *addSubscription)
...
-		ch:        r.events,
+		resolver:  r,
 		ctx:       ctx,
 	}
 	cloneCtx := add.ctx.clone(ctx)
 	trig = &trigger{
 		id:            triggerID,
-		subscriptions: make(map[*Context]*sub),
+		subscriptions: make(map[SubscriptionIdentifier]*subscriptionState),
+		updateBuf:     bytes.NewBuffer(make([]byte, 0, 1024)),
```
ysmolski: pls explain magic number

v2/pkg/engine/resolve/resolve.go
```diff
@@ -1445,128 +1467,109 @@ func (r *Resolver) subscriptionInput(ctx *Context, subscription *GraphQLSubscrip
 }
 
 type subscriptionUpdater struct {
+	mu        sync.Mutex
```
ysmolski, line 1466: does it protect everything in the struct?

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -2,930 +2,335 @@ package graphql_datasource
...
-	"github.com/cespare/xxhash/v2"
-	"github.com/gobwas/ws"
-	"github.com/gobwas/ws/wsutil"
+	"github.com/coder/websocket"
 	"github.com/jensneuse/abstractlogger"
-	"go.uber.org/atomic"
 
+	client "github.com/wundergraph/graphql-go-tools/v2/pkg/engine/datasource/graphql_datasource/subscriptionclient"
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/resolve"
-	"github.com/wundergraph/graphql-go-tools/v2/pkg/netpoll"
```
ysmolski, line 28: wooohooo. maybe we could remote netpoll as well? it had some quastionable implementation details. do we really need it?

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -2,930 +2,335 @@ package graphql_datasource
...
-		frameTimeout:    op.frameTimeout,
-		hashPool: sync.Pool{
-			New: func() interface{} {
-				return xxhash.New()
-			},
-		},
-		onWsConnectionInitCallback: op.onWsConnectionInitCallback,
-		netPollConfig:              op.netPollConfiguration,
+// WithWriteTimeout sets the timeout for WebSocket write operations (subscribe, unsubscribe, ping, pong).
+// Default: 5s.
```
ysmolski: Is this comment relevant? I don't see 5s anywhere

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -2,930 +2,335 @@ package graphql_datasource
...
-// Subscribe initiates a new GraphQL Subscription with the origin
-// If an existing WS connection with the same ID (Hash) exists, it is being re-used
-// If connection protocol is SSE, a new connection is always created
-// If no connection exists, the client initiates a new one
-func (c *subscriptionClient) Subscribe(ctx *resolve.Context, options GraphQLSubscriptionOptions, updater resolve.SubscriptionUpdater) error {
-	options.readTimeout = c.readTimeout
-	if options.UseSSE {
-		return c.subscribeSSE(ctx.Context(), c.engineCtx, options, updater)
+// WithReadLimit sets the maximum size in bytes for incoming WebSocket messages.
+// Default: 1MB.
```
ysmolski: Is this comment relevant? I don't see 1MB anywhere. maybe it is defined in the transport? in that case the default config on the top could mentioned it...

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/exports.go
```diff
@@ -0,0 +1,61 @@
...
+	ExecutionResult = common.ExecutionResult
+	Request         = common.Request
+	Options         = common.Options
+	Handler         = common.Handler
+	TransportType   = common.TransportType
+	WSSubprotocol   = common.WSSubprotocol
+	SSEMethod       = common.SSEMethod
+)
+
+// Re-export constants.
```
ysmolski, line 24: why are we doing re-export in this package?

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/common/options.go
```diff
@@ -0,0 +1,55 @@
+package common
+
+import (
+	"net/http"
+)
+
+type TransportType string
+
+const (
+	TransportWS  TransportType = "ws"
```
ysmolski: Excuse my ignorance. But do we have any docs on the differences between different transports and subprotocols from the client perspective? I see the names, but I would like to see more meaning attached to it, and especially, what to expect from those.

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -2,930 +2,335 @@ package graphql_datasource
...
-	data, err := readMessage(conn.netConn, c.frameTimeout, c.readTimeout)
-	if err != nil {
-		return handleConnectionError(err)
+// mapWSSubprotocol maps the string subprotocol to the client.WSSubprotocol type.
+func mapWSSubprotocol(proto string) client.WSSubprotocol {
+	switch proto {
+	case "graphql-ws":
+		return client.SubprotocolGraphQLWS
+	case "graphql-transport-ws":
+		return client.SubprotocolGraphQLTransportWS
```
ysmolski: these literals are equal to the constants? can we use constants everywhere?

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/protocol.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"context"
+	"errors"
+
+	"github.com/coder/websocket"
+
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/common"
+)
+
+type Protocol interface {
```
ysmolski, line 13: An exported interface, could you please add comments? Or maybe a high-level description. Up to you.

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/sse_conn.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+		c.handler(msg)
+
+		if msg.Type.IsTerminal() {
+			return
+		}
+	}
+}
+
+// parseEventBytes extracts the event type and data from raw SSE event bytes.
```
ysmolski, line 76: thank you, nice docs in this file

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_transport.go
```diff
@@ -0,0 +1,347 @@
...
+// WSTransportOptions configures a WSTransport.
+type WSTransportOptions struct {
+	UpgradeClient *http.Client
+	Logger        abstractlogger.Logger
+	PingInterval  time.Duration
+	PingTimeout   time.Duration
+	AckTimeout    time.Duration
+	WriteTimeout  time.Duration
+	ReadLimit     int64
+	// IdleTimeout is the duration a connection stays open after its last
```
ysmolski: I see this pattern in many files now. Do we have master documentation somewhere for those fields in one place?

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_conn.go
```diff
@@ -0,0 +1,281 @@
...
+	return c
+}
+
+func (c *wsConnection) subscribe(ctx context.Context, id string, req *common.Request, handler common.Handler) (func(), error) {
+	if c.closed.Load() {
+		return nil, common.ErrConnectionClosed
+	}
+
+	c.subsMu.Lock()
+
```
ysmolski: `closed` is being read without subsMu. What happens if `shutdown` is called after the `Load` but before the Lock? It might a TOCTU race. 

Maybe reading of `closed` should be protected as well?

And a corresponding test that fails:

	t.Run("subscribe rejects with ErrConnectionClosed under TOCTOU race", func(t *testing.T) {
		t.Parallel()

		// Regression test for TOCTOU between closed.Load() and subsMu.Lock()
		// in subscribe().
		//
		// Without a closed re-check under the lock, this sequence is possible:
		// 1. subscribe: closed.Load() → false  (check)
		// 2. shutdown:  closed.CAS(false,true)  (invalidates the check)
		// 3. shutdown:  swaps c.subs, dispatches errors to old handlers
		// 4. subscribe: subsMu.Lock(), adds handler to NEW empty map (use)
		// 5. subscribe: protocol.Subscribe → nil (mock doesn't check conn)
		// 6. subscribe returns (cancel, nil) — looks successful
		//
		// The handler is now orphaned: closed=true prevents any future
		// shutdown from running (CAS fails), so the handler will never
		// receive a terminal message. This test forces the exact interleaving
		// deterministically.

		conn, _ := newTestConn(t)
		proto := newMockProtocol()
		wsc := newWSConnection(conn, proto, wsConnectionOptions{})

		// Step 1: Hold subsMu so subscribe() blocks after its closed check.
		wsc.subsMu.Lock()

		subscribeResult := make(chan error, 1)
		go func() {
			// Passes closed.Load() (still false), then blocks on subsMu.Lock().
			_, err := wsc.subscribe(context.Background(), "sub-race", &common.Request{}, func(_ *common.Message) {})
			subscribeResult <- err
		}()

		// Let the goroutine reach the blocked Lock().
		runtime.Gosched()
		time.Sleep(5 * time.Millisecond)

		// Step 2: Simulate what shutdown does first — set closed=true.
		// We set it directly rather than calling shutdown() because shutdown
		// also needs subsMu (which we hold), and we want to control ordering.
		wsc.closed.Store(true)

		// Step 3: Release the lock. subscribe() now enters the critical
		// section with a stale view: it saw closed=false, but closed is now true.
		wsc.subsMu.Unlock()

		select {
		case err := <-subscribeResult:
			// With the fix: subscribe re-checks closed under the lock → ErrConnectionClosed.
			// Without the fix: subscribe succeeds (nil error) — the handler is orphaned
			// because closed=true means no future shutdown() can deliver a terminal message.
			require.ErrorIs(t, err, common.ErrConnectionClosed,
				"subscribe must detect closed state under the lock; without this check "+
					"the handler is orphaned (closed=true prevents future shutdown)")
		case <-time.After(time.Second):
			t.Fatal("subscribe did not return within timeout")
		}
	})

ysmolski (changes requested):
I gave it a first look. But I failed to comprehend the high level architecture of data flow in protocols/subs and especially from the angle of locks/mutexes. There are far too many locks going on and I cannot grasp the big picture when I read the code. Perhaps some design document could be devised to explain the architecture to mortals like me? 

I have some doubts regarding the mutexes design at the moment, but maybe if I understood what is going, I would have different opinion. I feel that current design is very fragile and prone to changes. I would afraid to touch the code I don't understant in full.

I left commments about missing docs and explanations too.

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/sse_transport.go
```diff
@@ -0,0 +1,266 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("create request: %w", err)
+	}
+
+	httpReq.Header.Set("Content-Type", "application/json")
+	httpReq.Header.Set("Accept", "text/event-stream")
+	httpReq.Header.Set("Cache-Control", "no-cache")
+
+	// Add custom headers
+	maps.Copy(httpReq.Header, opts.Headers)
```
ysmolski: Custom headers could overwrite SSE headers. In the old version SSE headers were overwriting the custom headers. Why did you change the order?

v2/pkg/engine/resolve/resolve.go
```diff
@@ -827,175 +740,314 @@ func (r *Resolver) handleAddSubscription(triggerID uint64, add *addSubscription)
...
+		}
+		if skip {
+			continue
+		}
+		pending = append(pending, pendingSubscriptionWrite{s})
+	}
+	trig.mu.Unlock()
+
+	for _, fe := range filterErrors {
+		r.asyncErrorWriter.WriteError(fe.ctx, fe.err, fe.response, fe.writer)
```
ysmolski: Should we protect this kind of writes with mutex?

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
endigma: Netpoll is still used in router for the time being

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/exports.go
endigma: some kind of happy medium between putting everything in the new client in one module and having a bunch of imports everywhere

v2/pkg/engine/resolve/resolve.go
```diff
@@ -806,17 +720,15 @@ func (r *Resolver) handleAddSubscription(triggerID uint64, add *addSubscription)
...
-		trig.subscriptions[add.ctx] = s
+		// Add the subscription to the registry so it can receive events
+		trig.mu.Lock()
+		trig.subscriptions[add.id] = s
+		trig.mu.Unlock()
+		r.addSubscriptionIndex(s)
+		// Execute the startup hooks in a goroutine to avoid holding the lock
+		go func() {
+			_ = r.executeStartupHooks(add, trig.updater)
+		}()
```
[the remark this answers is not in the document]
endigma: intentional

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/sse_transport_test.go
```diff
@@ -0,0 +1,880 @@
...
+			SSEMethod: common.SSEMethodGET,
+		}, handler)
+		require.NoError(t, err)
+		defer cancel()
+
+		// Verify GET method and query params
+		assert.Equal(t, http.MethodGet, receivedMethod)
+		assert.Equal(t, "subscription { test }", receivedQuery)
+		assert.Equal(t, `{"id":123}`, receivedVariables)
+		assert.Equal(t, "TestSub", receivedOperationName)
```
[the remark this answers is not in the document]
endigma: not true

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/graphql_transport_ws.go
```diff
@@ -0,0 +1,174 @@
...
+
+// Subscribe implements Protocol.
+func (p *GraphQLTransportWS) Subscribe(ctx context.Context, conn *websocket.Conn, id string, req *common.Request) error {
+	msg := outgoingMessage{
+		ID:      id,
+		Type:    gtwsTypeSubscribe,
+		Payload: req,
+	}
+	return wsjson.Write(ctx, conn, msg)
```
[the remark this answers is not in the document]
endigma: out of scope for protocol layer

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/graphql_ws.go
```diff
@@ -0,0 +1,175 @@
...
+		case gwsTypeConnectionKeepAlive:
+			// Keep-alive messages can arrive before ack, ignore them
+			continue
+		case gwsTypeConnectionError:
+			var errPayload map[string]any
+			if ackMessage.Payload != nil {
+				// If this fails, the error will have nil errors anyway, handling it does nothing unique
+				_ = json.Unmarshal(ackMessage.Payload, &errPayload)
+			}
+			return fmt.Errorf("%w: %v", ErrConnectionError, errPayload)
```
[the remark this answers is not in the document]
endigma: this is not an issue if errors are protocol correct

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/sse_transport.go
```diff
@@ -0,0 +1,268 @@
...
+	q := u.Query()
+	q.Set("query", req.Query)
+
+	if len(req.Variables) > 0 {
+		varsJSON, err := json.Marshal(req.Variables)
+		if err != nil {
+			return nil, fmt.Errorf("marshal variables: %w", err)
+		}
+		q.Set("variables", string(varsJSON))
+	}
```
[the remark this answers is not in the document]
endigma: not real

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/sse_transport.go
```diff
@@ -0,0 +1,268 @@
...
+		q.Set("operationName", req.OperationName)
+	}
+
+	if len(req.Extensions) > 0 {
+		extJSON, err := json.Marshal(req.Extensions)
+		if err != nil {
+			return nil, fmt.Errorf("marshal extensions: %w", err)
+		}
+		q.Set("extensions", string(extJSON))
+	}
```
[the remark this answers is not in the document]
endigma: nope, this is not what happens

.serena/project.yml
endigma: removed

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/protocol.go
endigma: not a good option here since it names the strings weirdly

v2/pkg/engine/resolve/resolve.go
endigma: it cannot

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_conn.go
endigma: fixed, good catch

v2/pkg/engine/resolve/resolve.go
endigma: yes!

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/sse_transport.go
endigma: fixed

v2/pkg/engine/resolve/resolve.go
ysmolski: awesome, thank you!

v2/pkg/engine/resolve/resolve.go
ysmolski: I see. It is complicated...

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
ysmolski: Is this default is being set somewhere else?

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
ysmolski: Is this default is being set somewhere else?

v2/pkg/engine/resolve/resolve.go
```diff
@@ -827,409 +815,458 @@ func (r *Resolver) handleAddSubscription(triggerID uint64, add *addSubscription)
...
-			err = add.resolve.Trigger.Source.Start(cloneCtx, add.headers, add.input, trig.updater)
-		}
+		err = add.resolve.Trigger.Source.Start(cloneCtx, add.headers, add.input, trig.updater)
 		if err != nil {
 			if r.options.Debug {
 				fmt.Printf("resolver:trigger:failed:%d\n", triggerID)
 			}
-			r.asyncErrorWriter.WriteError(add.ctx, err, add.resolve.Response, add.writer)
-			_ = r.emitTriggerClose(triggerID)
+			s.writeError(r.errorFormatter, add.ctx, err, add.resolve.Response)
```
ysmolski: here we send error only to the first subscribed. mayeb we could broadcast error to all the subscribers instead?
From LLM: 
When Source.Start() returns an error at resolve.go:847, only s (the first subscription) receives writeError at line 852. doneTriggerFromUpdater then closes all subscriptions' completed channels,
  but sub B never gets an error message — it just sees a clean termination.
  
Repro:
```go
// TestResolver_SourceStartFailure_SilentTermination verifies that when
// Source.Start() returns an error, ALL subscriptions on the trigger receive
// the error — not just the first one.
//
// Bug: only the first subscription (the one passed to the trigger-creation
// goroutine) gets writeError. doneTriggerFromUpdater closes every
// subscription's completed channel, but the others never receive an error
// message, so clients see a silent termination.
func TestResolver_SourceStartFailure_SilentTermination(t *testing.T) {
	resolverCtx, cancelResolver := context.WithCancel(context.Background())
	defer cancelResolver()

	// subBReady is closed once sub B's addSubscription has returned,
	// meaning B is registered on the trigger.
	subBReady := make(chan struct{})

	stream := &startFailStream{
		subBReady: subBReady,
	}

	plan := &GraphQLSubscription{
		Trigger: GraphQLSubscriptionTrigger{
			Source: stream,
			InputTemplate: InputTemplate{
				Segments: []TemplateSegment{
					{
						SegmentType: StaticSegmentType,
						Data:        []byte(`{"method":"POST","url":"http://localhost:4000","body":{"query":"subscription { counter }"}}`),
					},
				},
			},
			PostProcessing: PostProcessingConfiguration{
				SelectResponseDataPath:   []string{"data"},
				SelectResponseErrorsPath: []string{"errors"},
			},
		},
		Response: &GraphQLResponse{
			Data: &Object{
				Fields: []*Field{
					{
						Name: []byte("counter"),
						Value: &Integer{
							Path: []string{"counter"},
						},
					},
				},
			},
		},
	}

	resolver := New(resolverCtx, ResolverOptions{
		MaxConcurrency:                1024,
		AsyncErrorWriter:              &TestErrorWriter{},
		SubscriptionHeartbeatInterval: time.Hour,
	})

	writerA := &SubscriptionRecorder{buf: &bytes.Buffer{}}
	writerB := &SubscriptionRecorder{buf: &bytes.Buffer{}}

	idA := SubscriptionIdentifier{ConnectionID: NewConnectionID(), SubscriptionID: 1}
	idB := SubscriptionIdentifier{ConnectionID: NewConnectionID(), SubscriptionID: 2}

	ctxA := NewContext(context.Background())
	ctxB := NewContext(context.Background())

	// Subscribe A — creates the trigger.
	err := resolver.AsyncResolveGraphQLSubscription(ctxA, plan, writerA, idA)
	require.NoError(t, err)

	// Subscribe B — joins the existing trigger.
	err = resolver.AsyncResolveGraphQLSubscription(ctxB, plan, writerB, idB)
	require.NoError(t, err)

	// Unblock Source.Start() so it can fail.
	close(subBReady)

	// Sub A should receive an error.
	require.Eventually(t, func() bool {
		writerA.mux.Lock()
		defer writerA.mux.Unlock()
		return len(writerA.messages) > 0
	}, time.Second, 10*time.Millisecond, "sub A should receive an error from Source.Start failure")

	// Sub B should ALSO receive an error (not just a silent completed channel close).
	// If the bug is present, sub B gets no error message.
	require.Eventually(t, func() bool {
		writerB.mux.Lock()
		defer writerB.mux.Unlock()
		return len(writerB.messages) > 0
	}, 2*time.Second, 10*time.Millisecond,
		"sub B was silently terminated: completed channel closed but no error delivered — "+
			"Source.Start() failure only notified the first subscriber")
}

// startFailStream is a SubscriptionDataSource whose Start() waits for
// subBReady then returns an error.
type startFailStream struct {
	subBReady chan struct{}
}

func (s *startFailStream) Start(_ *Context, _ http.Header, _ []byte, _ SubscriptionUpdater) error {
	<-s.subBReady
	return errors.New("connection refused")
}
```

v2/pkg/engine/resolve/resolve.go
```diff
@@ -827,409 +815,458 @@ func (r *Resolver) handleAddSubscription(triggerID uint64, add *addSubscription)
...
 		if r.options.Debug {
 			fmt.Printf("resolver:trigger:start:%d\n", triggerID)
 		}
 
 		// This is blocking so the startup hook can decide if a subscription should be started or not by returning an error
-		err = r.executeStartupHooks(add, trig.updater)
+		err := r.executeStartupHooks(add, s, trig.updater)
 		if err != nil {
 			return
 		}
```
ysmolski: If startup hook fails then trigger could be orphaned.
In `executeStartupHooks` this line `_ = r.UnsubscribeSubscription(add.id)`  removes first sub only.
I am not sure how bad it is, but I was able LLM generate a test reproducing it:

```go
// TestResolver_StartupHookFailure_OrphansTrigger verifies that when the
// trigger-creation goroutine's startup hook fails, other subscriptions that
// have already joined the same trigger are properly cleaned up.
//
// Bug: the trigger-creation goroutine is the only path that calls
// Source.Start(). When its startup hook fails, it returns early without
// calling Start(). A second subscription that joined the trigger before
// the failure is left hanging on a trigger with no data source.
func TestResolver_StartupHookFailure_OrphansTrigger(t *testing.T) {
resolverCtx, cancelResolver := context.WithCancel(context.Background())
defer cancelResolver()

	// Create distinguishable contexts so the hook can tell A from B.
	ctxAInner, cancelA := context.WithCancel(context.Background())
	defer cancelA()
	ctxBInner, cancelB := context.WithCancel(context.Background())
	defer cancelB()

	// subBRegistered is closed once sub B has been registered on the
	// trigger, guaranteeing it joins before the startup hook for A fails.
	subBRegistered := make(chan struct{})

	stream := &hookFailStream{
		subBRegistered: subBRegistered,
		failCtx:        ctxAInner,
	}

	plan := &GraphQLSubscription{
		Trigger: GraphQLSubscriptionTrigger{
			Source: stream,
			InputTemplate: InputTemplate{
				Segments: []TemplateSegment{
					{
						SegmentType: StaticSegmentType,
						Data:        []byte(`{"method":"POST","url":"http://localhost:4000","body":{"query":"subscription { counter }"}}`),
					},
				},
			},
			PostProcessing: PostProcessingConfiguration{
				SelectResponseDataPath:   []string{"data"},
				SelectResponseErrorsPath: []string{"errors"},
			},
		},
		Response: &GraphQLResponse{
			Data: &Object{
				Fields: []*Field{
					{
						Name: []byte("counter"),
						Value: &Integer{
							Path: []string{"counter"},
						},
					},
				},
			},
		},
	}

	resolver := New(resolverCtx, ResolverOptions{
		MaxConcurrency:                1024,
		AsyncErrorWriter:              &TestErrorWriter{},
		SubscriptionHeartbeatInterval: time.Hour, // disable background heartbeats
	})

	writerA := &SubscriptionRecorder{buf: &bytes.Buffer{}}
	writerB := &SubscriptionRecorder{buf: &bytes.Buffer{}}

	idA := SubscriptionIdentifier{ConnectionID: NewConnectionID(), SubscriptionID: 1}
	idB := SubscriptionIdentifier{ConnectionID: NewConnectionID(), SubscriptionID: 2}

	ctxA := NewContext(ctxAInner)
	ctxB := NewContext(ctxBInner)

	// Subscribe A — this creates the trigger and spawns the goroutine
	// that runs startup hooks then Source.Start().
	err := resolver.AsyncResolveGraphQLSubscription(ctxA, plan, writerA, idA)
	require.NoError(t, err)

	// Subscribe B — this finds the existing trigger and joins it.
	err = resolver.AsyncResolveGraphQLSubscription(ctxB, plan, writerB, idB)
	require.NoError(t, err)

	// Signal that B is registered, unblocking A's startup hook
	// (which will then return an error).
	close(subBRegistered)

	// Sub A should receive an error (written as a message by TestErrorWriter)
	// and be cleaned up quickly.
	require.Eventually(t, func() bool {
		writerA.mux.Lock()
		defer writerA.mux.Unlock()
		return len(writerA.messages) > 0
	}, time.Second, 10*time.Millisecond, "sub A should receive an error from the failed startup hook")

	// Sub B is the one that would be orphaned by the bug.
	// It should also be cleaned up (either via error, complete, or messages)
	// within a reasonable timeout. If the bug is present, this times out.
	require.Eventually(t, func() bool {
		writerB.mux.Lock()
		hasMessages := len(writerB.messages) > 0
		writerB.mux.Unlock()
		isComplete := writerB.complete.Load()
		return hasMessages || isComplete
	}, 2*time.Second, 10*time.Millisecond,
		"sub B was orphaned: no error, no complete — "+
			"trigger left without a data source after first subscription's startup hook failed")

	// Source.Start should never have been called because the hook failed
	// before reaching it.
	require.False(t, stream.sourceStarted.Load(), "Source.Start() should not have been called")
}

// hookFailStream is a SubscriptionDataSource + HookableSubscriptionDataSource.
// The hook for the subscription whose context matches failCtx blocks until
// subBRegistered is closed, then returns an error. All other hooks succeed.
type hookFailStream struct {
subBRegistered chan struct{}
failCtx        context.Context
sourceStarted  atomic.Bool
}

func (s *hookFailStream) Start(_ *Context, _ http.Header, _ []byte, _ SubscriptionUpdater) error {
s.sourceStarted.Store(true)
// Block forever — in the bug scenario this is never reached.
select {}
}

func (s *hookFailStream) SubscriptionOnStart(ctx StartupHookContext, _ []byte) error {
if ctx.Context == s.failCtx {
// This is sub A (trigger-creation path): wait for B to join, then fail.
<-s.subBRegistered
return errors.New("startup hook failed")
}
// Sub B: succeed immediately.
	return nil
}
```

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/transport.go
```diff
@@ -0,0 +1,14 @@
...
+
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/common"
+)
+
+// Transport defines the interface for subscription transports.
+// A transport is responsible for managing the full connection to the upstream server.
+type Transport interface {
+	Subscribe(ctx context.Context, req *common.Request, opts common.Options) (results <-chan *common.Message, cancel func(), err error)
+	Close() error
+}
```
ysmolski: I dont think its being used anywhere

v2/pkg/engine/resolve/resolve.go
```diff
@@ -827,409 +815,458 @@ func (r *Resolver) handleAddSubscription(triggerID uint64, add *addSubscription)
...
+type pendingFilterError struct {
+	ctx      *Context
+	err      error
+	response *GraphQLResponse
+	sub      *subscriptionState
+}
+
+// handleTriggerUpdate sends data to all subscriptions of a trigger using snapshot-and-release.
+// The lock is released before performing I/O to avoid deadlocks when executeSubscriptionUpdate
+// calls AsyncUnsubscribeSubscription on flush failure.
```
ysmolski: stale comment

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_conn.go
```diff
@@ -0,0 +1,284 @@
...
+func (c *wsConnection) closeConn() {
+	c.shutdown(common.ErrConnectionClosed)
+}
+
+// writeTimeoutDuration returns the configured write timeout.
+func (c *wsConnection) writeTimeoutDuration() time.Duration {
+	return c.writeTimeout
+}
+
```
ysmolski: this could be just inlined in the tests

v2/pkg/engine/resolve/resolve.go
```diff
@@ -827,409 +815,458 @@ func (r *Resolver) handleAddSubscription(triggerID uint64, add *addSubscription)
...
+		r.reporter.SubscriptionCountDec(removedTotal)
+		if triggerDec > 0 {
+			r.reporter.TriggerCountDec(triggerDec)
+		}
 	}
+
+	r.triggers = make(map[uint64]*trigger)
+	r.subscriptionsByID = make(map[SubscriptionIdentifier]*subscriptionState)
+	r.subscriptionsByConnection = make(map[ConnectionID]map[SubscriptionIdentifier]*subscriptionState)
+	r.mu.Unlock()
```
ysmolski, line 1197: is this just a safety realloc if detachTriggerLocked somehow missed something?

ysmolski (approved):
This looks good to me, except for minor nits. Those two big tests need verifying - I did not have time to check them.

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/graphql_transport_ws.go
```diff
@@ -5,14 +5,17 @@ import (
...
 
 	"github.com/coder/websocket"
 	"github.com/coder/websocket/wsjson"
 
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/common"
 )
 
+// GraphQLTransportWS implements the graphql-transport-ws protocol.
+// See: https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md
+type GraphQLTransportWS struct{}
```
Noroth: Just a question. We are using the interface and `NewGraphQLTransportWS`. Does it make sense to export this type? Same for other protocols

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_conn.go
```diff
@@ -98,7 +98,10 @@ func (c *wsConnection) subscribe(ctx context.Context, id string, req *common.Req
 	c.subs[id] = handler
 	c.subsMu.Unlock()
```
Noroth, line 99: This should be fine, but if for any reason `c.subs` is nil you would never unlock in this case

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -67,7 +66,9 @@ func WithStreamingClient(c *http.Client) SubscriptionClientOption {
 // If not set, logging is disabled (silent operation).
 func WithLogger(log abstractlogger.Logger) SubscriptionClientOption {
 	return func(cfg *SubscriptionClientConfig) {
-		cfg.Logger = log
+		if log != nil {
+			cfg.Logger = log
+		}
```
Noroth, line 76: isn't `cfg.Logger` already nil? So does that `if` make sense or would you rather have some kind of noop logger?

v2/pkg/engine/resolve/resolve.go
```diff
@@ -27,10 +27,16 @@ const (
...
 
-// ConnectionIDs is used to create unique connection IDs for each subscription
-// Whenever a new connection is created, use this to generate a new ID
-// It is public because it can be used in more high level packages to instantiate a new connection
-var ConnectionIDs atomic.Int64
+// ConnectionID identifies a client connection for subscription routing.
+type ConnectionID int64
+
+// connectionIDs is the monotonic counter backing NewConnectionID.
+var connectionIDs atomic.Int64
```
Noroth: The name is a bit misleading in my opinion. connectionIDs sounds like a list or a set, not the counter.

v2/pkg/engine/resolve/resolve.go
```diff
@@ -497,21 +573,42 @@ func (s *subscriptionState) done() {
...
+func (s *subscriptionState) sendHeartbeat() error {
+	s.writeMu.Lock()
+	defer s.writeMu.Unlock()
+	if s.removed.Load() {
+		return nil
+	}
+	return s.writer.Heartbeat()
+}
+
 func (r *Resolver) executeSubscriptionUpdate(resolveCtx *Context, sub *subscriptionState, sharedInput []byte) {
```
Noroth, line 612: Well done! With the mutex refactoring and the new functions this is so much easier to comprehend!

v2/pkg/engine/resolve/resolve.go
```diff
@@ -789,52 +860,57 @@ func (r *Resolver) handleAddSubscription(triggerID uint64, add *addSubscription)
...
+	return nil
 }
 
-// markTriggerInitialized marks a trigger as initialized under the lock.
-func (r *Resolver) markTriggerInitialized(triggerID uint64) {
+func (r *Resolver) getTrigger(id uint64) (*trigger, bool) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	r.handleTriggerInitialized(triggerID)
+	trig, ok := r.triggers[id]
```
Noroth, line 866: Is there a chance that `r.triggers` is nil and this could cause a panic or why the `defer` here?
Also I don't think it matters much because Go should optimize the defer and inline it, just rather a micro optimization.

Noroth (commented):
Just some nits. Looking really good!

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_transport.go
```diff
@@ -0,0 +1,360 @@
...
+	IdleTimeout time.Duration
+}
+
+type WSTransport struct {
+	ctx  context.Context
+	opts WSTransportOptions
+
+	mu      sync.Mutex
+	dialing map[uint64]*dialResult
+	conns   map[uint64]*wsConnection
```
dkorittki, line 91: What was the design reason to choose two different maps for tracking dialing vs established connections? Why not have a single map `conns` here and put tracking connection state (dialing, established) in `wsConnection`? Is it to ensure `conns` only contains working, established connections and other parts of the code can rely on that fact? If for the latter it would be nice to have this in godoc / comment somewhere close.

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_transport.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+
+	if opts.PingInterval > 0 {
+		go t.pingLoop()
+	}
+
+	return t
+}
+
+func (t *WSTransport) Subscribe(ctx context.Context, req *common.Request, opts common.Options, handler common.Handler) (func(), error) {
```
dkorittki, line 138: Worth some godoc mentioning what it does under the hood

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_transport.go
```diff
@@ -0,0 +1,360 @@
...
+	wsConn.SetReadLimit(t.opts.ReadLimit)
+
+	proto, err := t.negotiateSubprotocol(opts.WSSubprotocol, wsConn.Subprotocol())
+	if err != nil {
+		t.opts.Logger.Error("wsTransport.dial",
+			abstractlogger.String("endpoint", opts.Endpoint),
+			abstractlogger.String("error", "subprotocol negotiation failed"),
+			abstractlogger.Error(err),
+		)
+		wsConn.Close(websocket.StatusProtocolError, err.Error())
```
dkorittki, line 279:
```suggestion
		_ = wsConn.Close(websocket.StatusProtocolError, err.Error())
```

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_transport.go
```diff
@@ -0,0 +1,360 @@
...
+	initCtx, initCancel := context.WithTimeout(ctx, t.opts.AckTimeout)
+	defer initCancel()
+
+	if err := proto.Init(initCtx, wsConn, opts.InitPayload); err != nil {
+		t.opts.Logger.Error("wsTransport.dial",
+			abstractlogger.String("endpoint", opts.Endpoint),
+			abstractlogger.String("error", "protocol init failed"),
+			abstractlogger.Error(err),
+		)
+		wsConn.Close(websocket.StatusProtocolError, "init failed")
```
dkorittki, line 292:
```suggestion
		_ = wsConn.Close(websocket.StatusProtocolError, "init failed")
```

v2/pkg/engine/resolve/resolve.go
endigma: renamed

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
endigma: it prevents the default cfg.Logger being overwritten if WithLogger is called with nil

endigma: since defaultSubscriptionClientConfig (which options act ontop of) has a noop abstractlogger so we don't panic

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_conn.go
endigma: not sure what you mean, can you elaborate? it's not possible for c.subs to be nil.

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/graphql_transport_ws.go
endigma: they can be unexported, yes

v2/pkg/engine/resolve/resolve.go
endigma: just a harmless double check, and maybe some minor benefit since an old used empty map technically takes up more room than a new empty map

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_conn.go
endigma: private getter for a private field, this is the peak of my engineering career

v2/pkg/engine/resolve/resolve.go
endigma: fixed, good catch

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/transport.go
endigma: correct, removed

v2/pkg/engine/resolve/resolve.go
endigma: moved to an explicit unlock, it cannot be nil

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_conn.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+}
+
+func (c *wsConnection) dispatch(msg *protocol.WireMessage) {
+	c.subsMu.RLock()
+	handler, exists := c.subs[msg.ID]
+	c.subsMu.RUnlock()
+
+	if !exists {
+		return
```
dkorittki, line 205: Tempted to say we should log a debug message here because it might be useful information for a user to know that their subgraph sent messages to with a subscription id unknown to us. On the other hand in rare occassions it could happen that we just deleted a subscriber from the map and received a message from upstream at the same time...

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/sse_conn.go
```diff
@@ -0,0 +1,175 @@
...
+// Based on r3labs/sse's processEvent but simplified for our needs.
+func (c *sseConnection) parseEventBytes(msg []byte) (eventType string, data []byte) {
+	if len(msg) == 0 {
+		return "", nil
+	}
+
+	// Split by newlines (normalize CR/LF)
+	for _, line := range bytes.FieldsFunc(msg, func(r rune) bool { return r == '\n' || r == '\r' }) {
```
dkorittki, line 84: Why not

```go
for line := range bytes.Lines(msg) {
	line = bytes.TrimRight(line, "\r\n")
	// ...
}
```

? Looks more intuitive to me

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_transport.go
```diff
@@ -0,0 +1,360 @@
...
+	"github.com/jensneuse/abstractlogger"
+	"github.com/rs/xid"
+
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/common"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/pool"
+)
+
+// ErrDialFailed indicates that the WebSocket dial (TCP + HTTP upgrade) failed.
+// The underlying cause is available via errors.Unwrap.
```
dkorittki, line 22: Move this inside the var block directly above `ErrDialFailed`. You can move `ErrInitFailed` inside this var block as well.

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/client.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/jensneuse/abstractlogger"
+
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/common"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport"
+)
+
+var ErrClientClosed = errors.New("client closed")
+
+type Client struct {
```
dkorittki, line 18: Since this is one of the most important exported types I think some godoc here makes sense

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -2,930 +2,350 @@ package graphql_datasource
...
-}
-
-// subscriptionClient allows running multiple subscriptions via the same WebSocket either SSE connection
-// It takes care of de-duplicating connections to the same origin under certain circumstances
-// If Hash(URL,Body,Headers) result in the same result, an existing connection is re-used
-type subscriptionClient struct {
-	streamingClient *http.Client
-	httpClient      *http.Client
+// SubscriptionClientConfig holds the subscription client configuration.
+type SubscriptionClientConfig struct {
```
dkorittki, line 19: Can be an unexported type

v2/pkg/engine/resolve/resolve.go
dkorittki: Maybe a way to make it clean is to put all state mutating logic into seperate functions covered by mutex handling and all other code uses these functions and does not need to deal with mutexes at all?

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/sse_transport.go
```diff
@@ -0,0 +1,259 @@
...
+		abstractlogger.String("method", string(method)),
+	)
+
+	// Use request context, but with transport requestCancel
+	requestCtx, requestCancel := context.WithCancel(context.WithoutCancel(ctx))
+
+	// Attach cancel to transport context
+	context.AfterFunc(t.ctx, requestCancel)
+	context.AfterFunc(ctx, requestCancel)
+
```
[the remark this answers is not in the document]
endigma: fixed

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
```diff
@@ -2,930 +2,320 @@ package graphql_datasource
...
 	}
-
-	if strings.HasPrefix(options.URL, "https") {
-		options.URL = "wss" + options.URL[5:]
-	} else if strings.HasPrefix(options.URL, "http") {
-		options.URL = "ws" + options.URL[4:]
-	}
-
-	return c.asyncSubscribeWS(ctx.Context(), c.engineCtx, id, options, updater)
 }
```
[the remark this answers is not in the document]
endigma: fixed

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_transport.go
```diff
@@ -0,0 +1,404 @@
...
+				if conn.isClosed() {
+					continue
+				}
+
+				if t.opts.pingTimeout > 0 && conn.pongOverdue(t.opts.pingTimeout) {
+					t.opts.logger.Debug("wsTransport.pingLoop",
+						abstractlogger.String("action", "pong_timeout"),
+					)
+					conn.closeConn()
+					continue
```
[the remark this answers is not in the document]
endigma: fixed

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/common/message.go
endigma: done

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/common/message.go
endigma: done

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/common/options.go
endigma: done

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/graphql_transport_ws.go
endigma: fixed

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/graphql_transport_ws.go
endigma: fixed

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/graphql_ws.go
endigma: fixed

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/protocol.go
endigma: renamed for clarity

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/protocol/protocol.go
endigma: done

v2/pkg/engine/resolve/resolve.go
endigma: @dkorittki this has been done already to the approximate limit, any more we're just fighting the architecture of resolver which fixing is out of scope, most of this locking behaviour is because of the writers provided by the router not being concurrency safe anyway

v2/pkg/engine/resolve/resolve.go
endigma: fixed mostly if not fully

v2/pkg/engine/resolve/resolve.go
endigma: fixed

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/client.go
endigma: made it a switch case with an error

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
endigma: moved to all one central spot in client.go or the wrapper

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
endigma: fixed

ysmolski (approved):
Gave it another go. LGTM from my side.

v2/pkg/engine/datasource/graphql_datasource/graphql_subscription_client.go
endigma: correct

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/client.go
endigma: added one

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_transport.go
endigma: Just style? I don't like var blocks with doc comments personally

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/sse_conn.go
endigma: SSE spec technically allows just `\r` as a line separator which is cursed. I'll go with bytes.Lines for now though

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_transport.go
endigma: done

v2/pkg/engine/datasource/graphql_datasource/subscriptionclient/transport/ws_transport.go
endigma: lifecycle of map items is very different, and it means every conns consumer needs to filter for ready connections before doing stuff on them (e.g. pingloop)

just very messy overall, partially valid wsConnection structs and checks everywhere

Noroth (approved):

dkorittki (approved):
lgtm

PR #1446, opened 2026-03-16
fix: avoid recursive input message stack overflows in gRPC planner

fengyuwusong:
## Summary
- reuse request message definitions while building recursive input object plans
- add a regression test for recursive input objects in the gRPC datasource planner

## Testing
- go test ./v2/pkg/engine/datasource/grpc_datasource

Refs wundergraph/cosmo#2649

## Summary by CodeRabbit

* **Performance Improvements**
  * Added caching to avoid rebuilding identical RPC messages during plan construction, reducing redundant work and improving plan generation speed.

* **Bug Fixes**
  * Introduced cycle detection when formatting RPC messages to prevent infinite recursion and correctly represent recursive message graphs.

* **Testing**
  * Added tests covering mutations with recursive input types, validating shared nested message reuse and correct execution plan structure.


v2/pkg/engine/datasource/grpc_datasource/execution_plan_recursive_input_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	andField := lookupField(conditionsField.Message.Fields, "and")
+	require.NotNil(t, andField)
+	require.True(t, andField.Repeated || andField.IsListType)
+	require.Same(t, conditionsField.Message, andField.Message)
+
+	orField := lookupField(conditionsField.Message.Fields, "or")
+	require.NotNil(t, orField)
+	require.True(t, orField.Repeated || orField.IsListType)
+	require.Same(t, conditionsField.Message, orField.Message)
```
Noroth, line 87: When looking at the test, you haven't checked if `key` and `value` are also part of message fields.
Also can you add more test cases that handle multiple arguments and nested arguments?

something like this:
```
input A {
  b: B  
}

input B {
  c: C
}

input C {
  a: A
  b: B
}

```

Noroth (changes requested):

fengyuwusong:
Closing this PR — the same stack-overflow issue has already been independently fixed in master with a functionally equivalent implementation:

- `formatRPCMessageVisited` (with a `visited map[*RPCMessage]struct{}`) provides cycle detection in plan stringification.
- `rpcPlanningContext.visitedInputTypes` caches recursive input messages in `buildMessageFromInputObjectType`.

These mirror this PR's `seen` map and `requestMessageCache` approach, so the core fix here is now redundant. Thanks @Noroth for the review feedback!

PR #1454, opened 2026-03-20
fix: check interface implementation in `potentiallySameObject` for nullability relaxation

davidomid:
Fixes https://github.com/wundergraph/graphql-go-tools/issues/1455

## Summary

When `RelaxSubgraphOperationFieldSelectionMergingNullability` is enabled, field selection merging should allow nullability differences between types that cannot overlap at runtime. The `potentiallySameObject` function previously treated any interface+object pair as potentially overlapping, which prevented nullability relaxation even when the object type does not implement the interface.

### Example

Using the existing `boxDefinition` test schema, consider these types:

- `NonNullStringBox1` — an **interface** with `scalar: String!`
- `StringBox` — an **object type** that implements `SomeBox` but does **not** implement `NonNullStringBox1`, with `scalar: String`

The following query:

```graphql
{
  someBox {
    ... on NonNullStringBox1 { scalar }
    ... on StringBox         { scalar }
  }
}
```

produces this validation error even with `RelaxSubgraphOperationFieldSelectionMergingNullability` enabled:

```
fields 'scalar' conflict because they return conflicting types 'String!' and 'String'
```

Since `StringBox` does not implement `NonNullStringBox1`, no runtime object can ever satisfy both type conditions simultaneously, so the nullability difference is safe to relax. After this fix, the query validates successfully when the relaxation flag is enabled.

### Fix

Updated `potentiallySameObject` to split the single interface case into three:

- **both interfaces** → conservatively assume overlap (`true`)
- **interface + object** → overlap only if the object implements the interface (check via `NodeImplementsInterfaceNode`)
- **two objects** → overlap only if same type name

## Testing

Two new test cases have been added alongside the existing `"disallows differing return type nullability when interface could overlap"` test, reusing the same `boxDefinition` schema:

1. **"allows differing return type nullability on interface vs non implementing type with relaxation"** — same query as the existing test (`NonNullStringBox1` vs `StringBox`) but with `FieldSelectionMerging(true)` → `Valid`. Directly demonstrates the fix: `StringBox` does not implement `NonNullStringBox1`, so relaxation applies.

2. **"rejects differing return type nullability on interface vs implementing type even with relaxation"** — `SomeBox` (interface, `scalar: String`) vs `NonNullStringBox1Impl` (implements `SomeBox`, `scalar: String!`), `FieldSelectionMerging(true)` → `Invalid`. Confirms that overlap detection still blocks relaxation when types actually overlap.

Additionally, some manual tests have been done by:
1. Pulling this branch into a branch of `wundergraph/cosmo`.
2. Pulling that `wundergraph/cosmo` branch into a federation router service.
3. Running some queries to verify that the conflict errors have been removed for non-overlapping types.

## Checklist

## Open Source AI Manifesto

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation of overlapping interface and object types during field selection merging.
  - Correctly allows compatible nullability differences when interface fragments overlap.
  - Continues to reject invalid nullability conflicts between unrelated object types.

- **Tests**
  - Added coverage for valid and invalid field selection merging scenarios involving nullable and non-null fields.


ysmolski:
Hi @davidomid. Is this PR ready for review again?

ysmolski (approved):
LGTM, but need another vote from @devsergiy too

tylerscoville:
@davidomid @ysmolski have there been any updates? I'm running into the same issue.

devsergiy (approved):

PR #1464, opened 2026-04-08
feat: defer additional fixes

devsergiy:
@coderabbitai summary

## Checklist

## Open Source AI Manifesto


docs/defer-design.md
```diff
@@ -0,0 +1,484 @@
...
+2. Removes `@defer` from the fragment node itself.
+3. Assigns a sequential integer ID to this defer group. IDs are assigned in AST walk order, so they reflect the order in which `@defer` fragments appear in the document.
+4. Records `parentDeferId` pointing to the enclosing defer group's ID, if any (for nested `@defer`).
+5. Stamps every field in the selection set with `@__defer_internal(id: "N", parentDeferId: "M", label: "...")`.
+
+After expansion a fragment like `... @defer { title }` becomes:
+
+```graphql
+... {
+  title @__defer_internal(id: "1")
```
jensneuse, line 123: when id is int, why string?

docs/defer-design.md
```diff
@@ -0,0 +1,490 @@
...
+
+### Step 1 — Collect nodes (`datasource_filter_collect_nodes_visitor.go`)
+
+Builds a `NodeSuggestion` tree that maps every field to one or more candidate datasources. For each field it reads `@__defer_internal` and attaches a `DeferInfo` struct (`id`, `parentDeferId`, `label`) to the suggestion, making the defer context of every field available to all subsequent steps.
+
+### Step 2 — Node selection and required fields (`node_selection_visitor.go`, `required_fields_visitor.go`)
+
+Resolves which datasource(s) handle each field. Also detects fields that require additional data to be fetched — `@key` fields for entity resolution and `@requires` fields for computed fields — and injects them directly into the operation AST in the correct defer scope.
+
+**`@requires` fields** are stamped with the same `@__defer_internal` as the field that needs them. They must be present in the same deferred fetch so the field resolver has the data it depends on.
```
jensneuse, line 192: what if multiple fields have overlapping requires?

docs/defer-design.md
```diff
@@ -0,0 +1,490 @@
...
+## High-Level Pipeline
+
+A query with `@defer` travels through four phases, each producing a richer representation for the next.
+
+### 1 · Normalization
+
+- `@defer` is removed from every fragment (inline or named spread)
+- Every field inside is stamped with `@__defer_internal(id, parentDeferId, label)`
+- defer IDs are assigned sequentially in AST walk order
+- A `___typename` placeholder is injected into any selection set where all children are deferred
```
ysmolski, line 68: Again, as I said `___` vs `__` might be not the best idea. Its very hard to distiungish them.

docs/defer-design.md
```diff
@@ -0,0 +1,490 @@
...
+---
+
+## Phase 1: Normalization
+
+**Files:**
+- `v2/pkg/astnormalization/inline_fragment_expand_defer.go`
+- `v2/pkg/astnormalization/defer_ensure_typename.go`
+- `v2/pkg/astnormalization/astnormalization.go` (opt-in via `WithInlineDefer()`)
+
+Normalization is enabled by passing `WithInlineDefer()` to the normalizer. Without it, `@defer` is left untouched and the rest of the pipeline treats the query as a normal synchronous request.
```
ysmolski, line 103: Is it because not all users will enable defer support?

docs/defer-design.md
```diff
@@ -0,0 +1,490 @@
...
+
+This visitor converts user-facing `@defer` on fragments into a per-field internal directive the planner can consume directly. It handles both inline fragments (`... on User @defer { ... }`, `... @defer { ... }`) and named fragment spreads (`...MyFragment @defer`).
+
+**What it does:**
+
+When it encounters `@defer` on a fragment:
+
+1. Checks `@defer(if: false)` — if disabled, removes `@defer` from the fragment but does not stamp any fields (they are treated as non-deferred).
+2. Removes `@defer` from the fragment node itself.
+3. Assigns a sequential integer ID to this defer group. IDs are assigned in AST walk order, so they reflect the order in which `@defer` fragments appear in the document.
```
ysmolski, line 115: So IDs on internal defers are, in fact, defer groups?

docs/defer-design.md
```diff
@@ -0,0 +1,490 @@
...
+... {
+  profile @__defer_internal(id: "1") {
+    ... {
+      bio @__defer_internal(id: "2", parentDeferId: "1")
+    }
+  }
+}
+```
+
+**Why stamp individual fields rather than keeping `@defer` on the fragment?**
```
ysmolski, line 139: FAQ sections are cool

docs/defer-design.md
```diff
@@ -0,0 +1,490 @@
...
+- `v2/pkg/engine/plan/visitor.go`
+
+Planning is the most involved phase for defer. Its job is to determine which datasource fetches each field, in which defer scope, and to build a set of planner instances — one per `(datasource, deferID)` pair — that will generate the downstream queries.
+
+### Why planners are scoped by deferID
+
+A planner is identified by its datasource hash **and** its deferID. Two planners can share the same datasource but serve different defer scopes. This separation is enforced during path assignment:
+
+- A planner whose `DeferID` is non-empty refuses to accept non-deferred fields.
+- A field with a `deferID` is only accepted by a planner whose `DeferID` matches exactly.
```
ysmolski, line 178: Are these two different kind of IDs?

devsergiy: sorry for the confusion it is the same kind of id
it comes from this line in code
```go
plannerConfig.DeferID() != field.deferID
```

docs/defer-design.md
devsergiy: kinda

It means that a single deferID could produce multiple fetches, which are grouped in GraphQLDeferResponse by deferID, e.g. it is a separate fetch tree

docs/defer-design.md
devsergiy: It is made pluggable, I'm not fully sure if there should be a global switch to turn it off

Also, it follows the approach in normalization where part of the rules are configured when you are creating the operation normalizer

One configuration option is missing currently - we should remove defer completely in case there was no Accept: multipart header

docs/defer-design.md
devsergiy: per discussion will change it to `internal__typename`

docs/defer-design.md
devsergiy: There is an issue to change it to the int

docs/defer-design.md
devsergiy: Currently, for such schemas

https://github.com/wundergraph/graphql-go-tools/blob/f59a327a6134e1a5298535b2895b1c23ef40f780/execution/engine/execution_engine_defer_test.go#L746-L799

Such a query with 2 fields requires selecting a different nested field in settings, but this fields are in the same defer scope
https://github.com/wundergraph/graphql-go-tools/blob/f59a327a6134e1a5298535b2895b1c23ef40f780/execution/engine/execution_engine_defer_test.go#L1366-L1376

Current logic always ads an alias to the required fields in defer scope, but reuses already existing alias

So operation with added required fields will look like this
```graphql
query DeferAllFields {
    user {
        name @__defer_internal(id: "1")
        billing @__defer_internal(id: "1") {
            plan @__defer_internal(id: "1")
        }
        settings @__defer_internal(id: "1") {
            region @__defer_internal(id: "1")
        }
        account @__defer_internal(id: "1") {
            type @__defer_internal(id: "1")
        }
        notifications @__defer_internal(id: "1")
        ___typename: __typename
        __internal_billing: billing @__defer_internal(id: "1") {
            plan @__defer_internal(id: "1")
        }
        __internal_settings: settings @__defer_internal(id: "1") {
            region @__defer_internal(id: "1")
            language @__defer_internal(id: "1")
        }
        __internal_name: name @__defer_internal(id: "1")
        __typename
        id
    }
}
```

In case of such query, when the fields with requirements in different defer scopes
https://github.com/wundergraph/graphql-go-tools/blob/f59a327a6134e1a5298535b2895b1c23ef40f780/execution/engine/execution_engine_defer_test.go#L1512-L1520

The query will look like this
```graphql
query DeferDerivedFieldsOnly {
    user {
        name
        billing {
            plan
        }
        settings {
            region
            language
        }
        account @__defer_internal(id: "1") {
            type @__defer_internal(id: "1")
        }
        notifications @__defer_internal(id: "2")
        __internal_billing: billing @__defer_internal(id: "1") {
            plan @__defer_internal(id: "1")
        }
        __internal_settings: settings @__defer_internal(id: "1") {
            region @__defer_internal(id: "1")
        }
        __internal_name: name @__defer_internal(id: "2")
        __internal_2_settings: settings @__defer_internal(id: "2") {
            language @__defer_internal(id: "2")
        }
        __typename
        id
    }
}
```

We always alias requirements because in situation like this
```graphql
query {
  user {
    settings {
      a
    }
    ... @defer {
      fieldRequiresSettingsB
      fieldRequiresSettingsC
    }
  }
}
```

We can't reuse settings, because it is not in the defer scope, and it will mean that we will fetch required fields with primary response, not via deffer group

devsergiy: hmm, after writing this comment I realized that there could be small improvement

```graphql
        settings @__defer_internal(id: "1") {
            region @__defer_internal(id: "1")
        }
        __internal_settings: settings @__defer_internal(id: "1") {
            region @__defer_internal(id: "1")
            language @__defer_internal(id: "1")
        }
```

In this case the existing field falls into the same defer scope, so we could actually reuse it

But in the case of such an operation

```graphql
        settings @__defer_internal(id: "1") {
            anyField @__defer_internal(id: "1")
            region @__defer_internal(id: "2")
        }
        __internal_settings: settings @__defer_internal(id: "1") {
            region @__defer_internal(id: "1")
            language @__defer_internal(id: "1")
        }
```

when needed `region` field is in the other defer scope, we will need to add an alias to this nested field, which currently is not implemented, as an alias is added only at the root of the requires selection set

Using aliases is a tradeoff to be able to properly distribute dependencies between fields, to assign them to a proper planner

In the last example `region` will be returned to the client only with defer group `2`
Potentially, I could mark a field that should be returned later, but fetched earlier via the other fetch

docs/defer-design.md
ysmolski: Sergiy wanted to be a dynamically typed Python/JS programmer, but we denied this to him!

v2/pkg/engine/resolve/resolve.go
```diff
@@ -475,40 +502,160 @@ func (r *Resolver) ResolveGraphQLDeferResponse(ctx *Context, response *GraphQLDe
...
+		groupLoader.appendSubgraphErrorsToContext()
+		dc.db.Unlock()
+		return err
+	}
+
+	// RENDER PHASE — serialised by the DataBuffer lock.
+	dc.db.Lock()
+	defer dc.db.Unlock()
+
+	isLast := syncatomic.AddInt64(remaining, -1) == 0
```
ysmolski: we dont decrement when error happens above? is this expected?

v2/pkg/engine/resolve/response.go
```diff
@@ -59,6 +59,26 @@ func (g *GraphQLResponse) SingleFlightAllowed() bool {
 type GraphQLDeferResponse struct {
 	Response *GraphQLResponse
 	Defers   []*DeferFetchGroup
+
+	// DeferDescriptors lists every @defer fragment in the operation, keyed by ID.
```
ysmolski, line 63: You should mention somewhere that valid IDs start with 1 and <1 means non defer. It might be non-obvious to new readers.

v2/pkg/engine/resolve/response.go
```diff
@@ -59,6 +59,26 @@ func (g *GraphQLResponse) SingleFlightAllowed() bool {
...
+	// DeferTree is the execution tree built from DeferDescriptors during post-processing.
+	// Nil until the buildDeferTree post-processor runs.
+	DeferTree *DeferTreeNode
+}
+
+// DeferDescriptor describes a single @defer fragment for the incremental-delivery
+// envelope. Path is the response path of the fragment (where it was mounted in
+// the operation); Label is the user-supplied label (empty when none); ParentID
+// is the id of the enclosing @defer (0 for top-level).
+type DeferDescriptor struct {
```
ysmolski, line 77: Nit, but I dontl ike how LLM does not care about newlines and sentences. Anyway, here is the proper commented block:

```go
// DeferDescriptor describes a single @defer fragment for the incremental-delivery envelope.
type DeferDescriptor struct {
// Valid IDs start with 1.
ID       int

// ParentID is the id of the enclosing @defer (0 for top-level).
ParentID int

// Label is the user-supplied label (empty when none);
Label    string

// Path is the response path of the fragment (where it was mounted in the 
Path     []string
}
```

v2/pkg/engine/resolve/defer_tree.go
```diff
@@ -0,0 +1,28 @@
+// v2/pkg/engine/resolve/defer_tree.go
```
ysmolski, line 1: redundant comment

v2/pkg/engine/resolve/resolve.go
```diff
@@ -494,43 +523,163 @@ func (r *Resolver) ResolveGraphQLDeferResponse(ctx *Context, response *GraphQLDe
 			return nil, err
 		}
 
-		if t.resolvable.hasErrors() {
-			// errors are already written into the initial payload;
-			// terminate the incremental stream
+		defer func() {
```
ysmolski, line 526: as we discussed, it should be declared before early returns from errors

v2/pkg/ast/ast_field.go
```diff
@@ -256,16 +254,42 @@ func (d *Document) AddDeferInternalDirectiveToField(fieldRef int, id int, label
 }
 
 func (d *Document) FieldInternalDeferID(fieldRef int) (id int, exists bool) {
-	directiveRef, exists := d.Fields[fieldRef].Directives.HasDirectiveByNameBytes(d, literal.DEFER_INTERNAL)
-	if !exists {
-		return 0, false
+	id, _, exists = d.FieldInternalDeferIDWithDirectiveRef(fieldRef)
```
ysmolski, line 257: what if we return `id==0` and `exists==false`?

v2/pkg/engine/resolve/resolve.go
```diff
@@ -494,43 +523,163 @@ func (r *Resolver) ResolveGraphQLDeferResponse(ctx *Context, response *GraphQLDe
...
+	// This is a pre-existing issue to be addressed separately.
+	dc.resolvable.skipValueCompletion = groupLoader.skipValueCompletion
+
+	descriptor := dc.resolvable.deferDescriptors[group.DeferID]
+	dc.resolvable.currentDefer = &descriptor
+	if err := dc.resolvable.ResolveDefer(dc.response.Response.Data, dc.writer, !isLast); err != nil {
+		return err
+	}
+	return dc.writer.Flush()
+}
```
ysmolski, line 631: Hmm. We do flush (network write) while holding the lock. Might be not good for performance. If the client is slow then groups fetch-merge become very slow. Maybe we could render it into the buffer while having the lock and then flush that buffer without the lock.

v2/pkg/astnormalization/defer_expand_into_internal.go
```diff
@@ -44,9 +44,17 @@ type deferInfo struct {
 
 func (f *deferExpandIntoInternalVisitor) EnterDocument(operation, _ *ast.Document) {
 	f.operation = operation
+	f.currentDeferId = 0
+	f.defers = f.defers[:0]
 }
 
 func (f *deferExpandIntoInternalVisitor) EnterInlineFragment(ref int) {
+	// expand defer only on operation fields
+	// this rule runs after fragments was inlined, but before they removed
```
ysmolski, line 53: This rule runs after fragments were inlined, but before they are removed.

v2/pkg/engine/resolve/loader.go
```diff
@@ -301,7 +353,9 @@ func (l *Loader) resolveSingle(item *FetchItem) error {
 		if err != nil {
 			return err
 		}
+		l.dataBuffer.Lock()
 		err = l.mergeResult(item, res, items)
+		l.dataBuffer.Unlock()
```
ysmolski: I know that you asked me not to look at it, but I could not help it. This locking looks weird to me.

ysmolski (commented):
1st pass is done

v2/pkg/engine/resolve/resolve_defer_parallel_test.go
```diff
@@ -0,0 +1,315 @@
...
+	ctx := NewContext(context.Background())
+
+	_, err := r.ResolveGraphQLDeferResponse(ctx, response, writer)
+	require.NoError(t, err)
+
+	joined := ctx.SubgraphErrors()
+	require.Error(t, joined, "defer-group subgraph errors must aggregate into the Context")
+	assert.Contains(t, joined.Error(), "subgraph-A")
+	assert.Contains(t, joined.Error(), "subgraph-B")
+}
```
ysmolski, line 310: What if we add following benchmarks to verify performance of parallel defer? I got following numbers on my box:
```
BenchmarkResolveDeferTree_ParallelSiblings/siblings=2-14     	 2570790	      9473 ns/op	    5049 B/op	      87 allocs/op
BenchmarkResolveDeferTree_ParallelSiblings/siblings=8-14     	  905434	     26857 ns/op	   14286 B/op	     249 allocs/op
BenchmarkResolveDeferTree_ParallelSiblings/siblings=32-14    	  182517	    127917 ns/op	   51885 B/op	     879 allocs/op
BenchmarkResolveDeferTree_ParallelSiblings/siblings=64-14    	   80995	    287135 ns/op	   99870 B/op	    1714 allocs/op
BenchmarkResolveDeferTree_ParallelSiblings_WithLatency/siblings=2-14         	   10000	   2338409 ns/op	    5244 B/op	      89 allocs/op
BenchmarkResolveDeferTree_ParallelSiblings_WithLatency/siblings=8-14         	   10000	   2393958 ns/op	   15056 B/op	     257 allocs/op
BenchmarkResolveDeferTree_ParallelSiblings_WithLatency/siblings=32-14        	    9036	   2668388 ns/op	   54975 B/op	     911 allocs/op
BenchmarkResolveDeferTree_ParallelSiblings_WithLatency/siblings=64-14        	    7702	   3215936 ns/op	  106040 B/op	    1778 allocs/op
```

```go
// BenchmarkResolveDeferTree_ParallelSiblings measures the per-request overhead of the parallel
// defer. For a fan of N root-level sibling defers resolved as a single DeferParallel node.
func BenchmarkResolveDeferTree_ParallelSiblings(b *testing.B) {
	for _, n := range []int{2, 8, 32, 64} {
		b.Run(fmt.Sprintf("siblings=%d", n), func(b *testing.B) {
			rCtx, cancel := context.WithCancel(context.Background())
			defer cancel()
			r := newResolver(rCtx)

			groups := make([]*DeferFetchGroup, n)
			descriptors := make(map[int]DeferDescriptor, n)
			children := make([]*DeferTreeNode, n)
			for i := range n {
				id := i + 1
				groups[i] = &DeferFetchGroup{
					DeferID: id,
					Fetches: Single(&SingleFetch{
						FetchConfiguration: FetchConfiguration{
							DataSource: FakeDataSource(fmt.Sprintf(`{"f%d":"v%d"}`, id, id)),
						},
					}),
				}
				descriptors[id] = DeferDescriptor{ID: id, ParentID: 0}
				children[i] = DeferSingle(groups[i])
			}

			// The response plan is read-only at resolve time and reused across
			// iterations, mirroring how a cached plan is served per request.
			response := minimalDeferResponse(groups, descriptors)
			response.DeferTree = DeferParallel(children...)

			b.ReportAllocs()
			b.ResetTimer()
			for i := 0; i < b.N; i++ {
				writer := &testDeferWriter{}
				ctx := NewContext(context.Background())
				if _, err := r.ResolveGraphQLDeferResponse(ctx, response, writer); err != nil {
					b.Fatal(err)
				}
			}
		})
	}
}

// BenchmarkResolveDeferTree_ParallelSiblings_WithLatency is the latency counterpart:
// every sibling's fetch sleeps deferFetchLatency.
// Because the N siblings resolve concurrently, the network waits overlap,
// so ns/op should stay ≈ deferFetchLatency regardless of N.
func BenchmarkResolveDeferTree_ParallelSiblings_WithLatency(b *testing.B) {
	const deferFetchLatency = 2 * time.Millisecond

	for _, n := range []int{2, 8, 32, 64} {
		b.Run(fmt.Sprintf("siblings=%d", n), func(b *testing.B) {
			rCtx, cancel := context.WithCancel(context.Background())
			defer cancel()
			r := newResolver(rCtx)

			groups := make([]*DeferFetchGroup, n)
			descriptors := make(map[int]DeferDescriptor, n)
			children := make([]*DeferTreeNode, n)
			for i := range n {
				id := i + 1
				ds := FakeDataSource(fmt.Sprintf(`{"f%d":"v%d"}`, id, id))
				ds.artificialLatency = deferFetchLatency
				groups[i] = &DeferFetchGroup{
					DeferID: id,
					Fetches: Single(&SingleFetch{
						FetchConfiguration: FetchConfiguration{DataSource: ds},
					}),
				}
				descriptors[id] = DeferDescriptor{ID: id, ParentID: 0}
				children[i] = DeferSingle(groups[i])
			}

			response := minimalDeferResponse(groups, descriptors)
			response.DeferTree = DeferParallel(children...)

			b.ReportAllocs()
			b.ResetTimer()
			for i := 0; i < b.N; i++ {
				writer := &testDeferWriter{}
				ctx := NewContext(context.Background())
				if _, err := r.ResolveGraphQLDeferResponse(ctx, response, writer); err != nil {
					b.Fatal(err)
				}
			}
		})
	}
}
```

v2/pkg/engine/resolve/response.go
devsergiy: fixed in the follow up

v2/pkg/engine/resolve/response.go
devsergiy: fixed in the follow up

v2/pkg/engine/resolve/defer_tree.go
devsergiy: fixed in the follow up

v2/pkg/engine/resolve/resolve.go
devsergiy: fixed in the follow up

v2/pkg/engine/resolve/resolve.go
devsergiy: here is overview

```
❯ but what is the answer? is it worth it? if we anyway should keep the order of flushes predictable                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                           
⏺ Short answer: no, I don't think it's worth it — and your instinct is the reason why.                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                           
  Here's the logic:                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                           
  The flushes are serial no matter what. Frame integrity (no interleaving) + terminal hasNext:false last force the network writes to be serialized. So the client-facing throughput — the sum of the flush times — is identical in both designs. The refactor cannot make the flushes      
  faster.                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                           
  So what would the refactor actually buy? Only this: group B's merge + render (CPU/astjson work) could run during group A's network flush, instead of after it. That's pipelining the cheap part behind the expensive part. The win is roughly (N-1) × (merge+render time) — and for      
  typical defer, merge+render is microseconds (small incremental payloads, a cheap astjson merge) while a slow client's flush is milliseconds. So the saving is on the order of microseconds while the wall-clock is dominated by the serial flushes you can't avoid. Negligible.          
                                                                                                                                                                                                                                                                                           
  Your point nails the cost/benefit: since we must keep flush order predictable anyway, the refactor doesn't remove serialization — it re-implements it with extra machinery (a dedicated flusher goroutine + an ordered queue + terminal-last logic + unbounded buffering that grows      
  memory under a slow client). You'd be adding complexity to re-achieve the ordering the single lock gives you for free, in exchange for a marginal pipelining gain.                                                                                                                       
                                                                                                                                                                                                                                                                                           
  And the blast radius is small. The DataBuffer lock is per-request — a slow client for request X stalls only X's own remaining defers, never other requests. And those defers couldn't reach that client any sooner anyway (serial flushes). So the "stall" is mostly cosmetic for        
  throughput.                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                           
  When it would matter: large fan-out (many parallel defers) and heavy per-defer payloads (expensive merge/render) and slow clients, all at once. That's a narrow corner.                                                                                                                  
                                                                                                                                                                                                                                                                                           
  My recommendation for the review reply: keep the current design; it's simpler and correct, and the serialization it relies on is unavoidable. The test now documents the behavior as a conscious choice. Revisit only if profiling a real workload shows flush-lock contention is a      
  measured bottleneck — at which point the buffered render + ordered off-lock flush is the known fix.                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                           
  One cheap half-measure if you ever want it without the full refactor: the fetch already overlaps (it's off-lock), which is the genuinely expensive part (network to subgraphs). It's only the merge that's serialized behind flush — and merge is the cheap part. So most of the         
  available parallelism is already captured.
```

v2/pkg/engine/resolve/resolve.go
ysmolski: i see it being removed

ysmolski (approved):

PR #1191, opened 2025-06-25
fix(graphiql): update graphiql playground to latest version

juliankoehn:
This PR migrates GraphiQL to v5 following the official migration guide:  
https://github.com/graphql/graphiql/blob/main/docs/migration/graphiql-5.0.0.md

relates https://github.com/graphql/graphiql/issues/4034

PR #1483, opened 2026-05-01
feat: switch from composition-go wrapper to cosmo wgc composition

### LINKED ISSUE #860: Popover does not disappear in Playground UI after latest Chrome update
Hello,

We have recently encountered an issue where the UI pop-up hints in the Playground never disappear while using the interface. From what we have gathered, this appears to be caused by a change in the latest version of Chrome.

Here is a screenshot of an example, pardon the blanked out field names :)

popup

Note the Date and the BigInt popups never disappear until refreshing the page.

We have noticed that the following error message is now appearing in the developer console:
Screenshot 2024-07-29 151005

We are running Chrome Version 127.0.6533.73 (Official Build) (64-bit) on Windows 11.

Currently our stack is on graphql-go-tools version 1.64.0, but I can confirm that upgrading to the latest version of V1 does not resolve the issue.

devsergiy:
Playground was fixed and moved from the engine to the execution. 

Closes #1191
Fixes #860

@coderabbitai summary


execution/federationtesting/gateway/gateway.go
```diff
@@ -3,94 +3,67 @@ package gateway
...
-	handler.ServeHTTP(w, r)
-}
-
-func (g *Gateway) Ready() {
-	<-g.readyCh
-}
+	var rc nodev1.RouterConfig
+	if err := protojson.Unmarshal(g.configFileContent, &rc); err != nil {
+		g.logger.Fatal("can't unmarshal composed config: %v", log.Error(err))
+	}
```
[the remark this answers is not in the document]
devsergiy: log.Fatal will panic

devsergiy:
@claude review

execution/federationtesting/util.go
```diff
@@ -58,10 +47,15 @@ func NewFederationSetup(addGateway ...func(s *FederationSetup) *httptest.Server)
 	}
 
 	if len(addGateway) > 0 {
-		setup.GatewayServer = addGateway[0](setup)
+		gw, err := addGateway[0](setup)
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
devsergiy: it is used only in tests, no runtime usage

execution/engine/testdata/config_factory_federation/compose.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
ysmolski, line 1: I don't see how testing triggers this. If this is gonna be called manually then `wgc`t should be pinned to a specific version so the config would not change in the CI.

devsergiy: only manually, never in ci, so we will control when it will be updated
latest should always work, otherwise something wrong

examples/engine/main.go
```diff
@@ -0,0 +1,666 @@
...
+		out := &bytes.Buffer{}
+		err, _ := resolver.ResolveGraphQLResponse(ctx, p.Response, nil, out)
+		if err != nil {
+			panic(err)
+		}
+		fmt.Println(out.String()) // Output: {"data":{"hello":"world"}}
+	case *plan.SubscriptionResponsePlan:
+		// this is a Query, so we ignore Subscriptions for now, but they are supported
+	}
+}
```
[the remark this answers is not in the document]
devsergiy: no real op is used, just an example of approach

execution/engine/testdata/config_factory_federation/compose.sh
ysmolski: Oh okay. What do you think about adding a check to the CI that the composed config in the repo is up to date with the latest wgc?

execution/playground/playground.go
```diff
@@ -3,26 +3,16 @@ package playground
...
-	contentTypeImagePNG       = "image/png"
-	contentTypeTextHTML       = "text/html"
-	contentTypeTextCSS        = "text/css"
-	contentTypeTextJavascript = "text/javascript"
-
-	cssFile     = "playground.css"
-	jsFile      = "playground.js"
-	faviconFile = "favicon.png"
-	logoFile    = "logo.png"
+	playgroundTemplate  = "playgroundTemplate"
```
ysmolski, line 13: Good call to move the playground from engine to the execution.

execution/playground/files/playground.html
```diff
@@ -0,0 +1,98 @@
...
+            "react-dom": "https://esm.sh/react-dom@19.1.0",
+            "react-dom/": "https://esm.sh/react-dom@19.1.0/",
+
+            "graphiql": "https://esm.sh/graphiql@5?standalone&external=react,react-dom,@graphiql/react,graphql",
+            "graphiql/": "https://esm.sh/graphiql@5/",
+            "@graphiql/plugin-explorer": "https://esm.sh/@graphiql/plugin-explorer@5?standalone&external=react,@graphiql/react,graphql",
+            "@graphiql/react": "https://esm.sh/@graphiql/react@0.37?standalone&external=react,react-dom,graphql,@graphiql/toolkit,@emotion/is-prop-valid",
+
+            "@graphiql/toolkit": "https://esm.sh/@graphiql/toolkit@0.11?standalone&external=graphql",
+            "graphql": "https://esm.sh/graphql@16.11.0",
```
ysmolski, line 56: I have tested `examples/federation` server. It works again.

ysmolski (approved):

execution/engine/testdata/config_factory_federation/compose.sh
devsergiy: don't think it is necessary

PR #1485, opened 2026-05-05
feat: support dot-path in slicingArguments

ysmolski:
This PR enables the use of dot-path slicing arguments in a listSize directives. If an operation uses deeply nested input objects, then user can specify the deeply nested int value to be used as an estimation of the returned list size:

    input PaginationInput {
      first: Int
      after: String
    }

    input SearchInput {
      pagination: PaginationInput
      query: String
    }

    type Query {
      search(input: SearchInput!): [Book]
           @listSize(slicingArguments: ["input.pagination.first"])
    }


v2/pkg/engine/plan/cost.go
```diff
@@ -82,18 +83,30 @@ type FieldListSize struct {
 // multiplier returns the multiplier based on arguments and variables.
 // It picks the maximum value among slicing arguments, otherwise it tries to use AssumedSize.
 // If neither is available, it falls back to defaultListSize.
-//
-// Does not take into account the SizedFields; TBD later.
 func (ls *FieldListSize) multiplier(arguments map[string]ArgumentInfo, vars *astjson.Value, defaultListSize int) int {
 	multiplier := -1
 	for _, slicingArg := range ls.SlicingArguments {
+		// First, try arg as the dot-path:
+		if vars != nil && strings.Contains(slicingArg, ".") {
```
devsergiy: Can it be extracted into a helper? used in 2 places

execution/engine/execution_engine_cost_test.go
```diff
@@ -1180,6 +1201,414 @@ func TestExecutionEngine_Cost(t *testing.T) {
...
+								},
+								Types: map[string]int{"Item": 3},
+							},
+						},
+						customConfig,
+					),
+				},
+				fields: fieldConfig,
+			},
+			"external: field 'Query.search' requires exactly one slicing argument, but none was provided, locations: [], path: [search]",
```
devsergiy, line 1608: Will default values support be implemented in a separate PR?

devsergiy (changes requested):

v2/pkg/engine/plan/cost.go
ysmolski: I have refactored it a bit. THere is still some potential to refactor it further, but I would prefer to avoid it now and do it in the fix of default values for slicingArguments.

v2/pkg/engine/plan/cost.go
```diff
@@ -82,42 +83,70 @@ type FieldListSize struct {
...
+func resolveSlicingArgIntValue(slicingArg string, arguments map[string]ArgumentInfo, vars *astjson.Value) (int, bool) {
+	path := strings.Split(slicingArg, ".")
+	inputArg := path[0]
+	arg, ok := arguments[inputArg]
+	if ok && arg.hasVariable && arg.isInputObject {
+		value := vars.Get(arg.varName)
+		for _, key := range path[1:] {
+			value = value.Get(key)
+			if value == nil {
+				return 0, false
```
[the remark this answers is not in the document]
ysmolski: why the need for the `value.Type() == astjson.TypeNull` check?

devsergiy (approved):

PR #1480, opened 2026-04-30
feat(grpc_datasource): add ConnectRPC transport for gRPC datasource

fengyuwusong:
@coderabbitai summary

## Summary

This PR introduces an `RPCTransport` abstraction inside the gRPC data source so that subgraph calls can be made over either native gRPC or [ConnectRPC](https://connectrpc.com/). It then adds the Connect implementation, regenerates the `grpctest` fixture with `buf` so the same backing `MockService` can be served over both transports, and locks in end-to-end coverage through the data source layer.

ConnectRPC is the right transport whenever end-to-end HTTP/2 to the subgraph is unavailable. The feature is consumed by the matching cosmo router work (see "Related").

## Related

- Required by: wundergraph/cosmo#2808 (router-side `grpc_protocol` configuration that picks the transport per subgraph).
- Replaces: #1453.

### Data source transport abstraction

- New `RPCTransport` interface (`transport.go`) wraps the previous direct `grpc.ClientConnInterface` use site so the data source can dispatch through any concrete transport.
- `NewDataSource(transport, config)` accepts the abstraction directly; `NewDataSourceGRPC(conn, config)` is a thin convenience wrapper that preserves the existing call shape for gRPC consumers.

### ConnectRPC transport

- `transport_connect.go` implements the Connect protocol over HTTP/1.1 with both Protobuf and JSON encodings.
- Follow-up fixes from earlier reviews are folded in:
  - base64-encode binary metadata headers,
  - nil-guard the response body and cap it at 10 MB,
  - detect oversized responses and truncate the surfaced error body.
- `transport_connect_test.go` covers Protobuf, JSON, Connect-protocol errors, non-JSON errors, and ASCII / binary header forwarding.

### grpctest fixture migration to `buf`

- New `buf.yaml` and `buf.gen.yaml` drive `protoc-gen-go`, `protoc-gen-go-grpc`, and `protoc-gen-connect-go` from a single generation step. The `Mproduct.proto=` plugin option remaps the proto's `go_package` onto the actual repository path so the generated Connect package can import `productv1` cleanly.
- `Makefile` `generate-proto` target now invokes `buf generate`.
- `productv1/productv1connect/product.connect.go` is checked in.

### Connect-compatible MockService

[...]

PR #1490, opened 2026-05-08
feat(grpc_datasource): introduce RPCTransport abstraction with gRPC implementation

fengyuwusong:
@coderabbitai summary

## Summary

Refactor the gRPC data source so the dial path goes through a small `RPCTransport` interface instead of a hard-coded `grpc.ClientConnInterface` dependency. This is a no-op for behaviour and a no-op for existing callers, but it lets follow-up work plug additional transports (Connect, gRPC-Web, in-memory mocks) into the same data source pipeline.

This is the first of two PRs that together replace #1480.

## Related

- Replaces: #1480 (split into this PR + a follow-up that adds the ConnectRPC implementation)
- Follow-up: a second PR will add a Connect-protocol implementation of `RPCTransport`, regenerate the `grpctest` fixture via `buf` to also emit Connect handlers, and exercise the data source end-to-end with both transports.
- Consumed by: wundergraph/cosmo#2808 (the router-side feature that motivated this work)

### `pkg/engine/datasource/grpc_datasource`

- New `transport.go`: defines the `RPCTransport` interface and a `GRPCTransport` implementation that wraps `grpc.ClientConnInterface`. The interface takes the same `(ctx, methodFullName, input, output)` shape the data source already passes to `grpc.ClientConnInterface.Invoke`, so swapping it in is mechanical.
- `grpc_datasource.go`: `DataSource` now keeps an `RPCTransport` rather than a raw `grpc.ClientConnInterface`. `NewDataSource` accepts the abstraction; a thin convenience wrapper `NewDataSourceGRPC` preserves the existing call shape for callers that already have a `grpc.ClientConnInterface`.
- New `transport_test.go`: shared helpers (`newTestCompiler`, `findMessageDesc`) plus a smoke test for `GRPCTransport.Invoke`. Helpers live here so the follow-up Connect transport tests can reuse them without a circular dependency.
- `grpc_datasource_test.go`, `grpc_datasource_federation_test.go`, `grpc_datasource_spy_test.go`: every existing `NewDataSource(conn, …)` call site is migrated onto `NewDataSourceGRPC(conn, …)`. Tests still take a `grpc.ClientConnInterface` directly.

### `pkg/engine/datasource/graphql_datasource`

- `graphql_datasource.go`: the planner constructed the gRPC data source inline via `grpcdatasource.NewDataSource(p.grpcClient, …)`. With the new signature this would no longer compile, so the call switches to `grpcdatasource.NewDataSourceGRPC(p.grpcClient, …)`. No behavioural change, no change to public Factory APIs.

## Backward compatibility

- All existing `NewDataSource` consumers should switch to `NewDataSourceGRPC`. The old function name is kept but takes the abstraction, not a `grpc.ClientConnInterface`.
- No behaviour change for the gRPC dial path: `GRPCTransport.Invoke` simply forwards to the underlying client connection.

## Testing

- `pkg/engine/datasource/grpc_datasource`: existing test suite (455+ cases) passes; new `TestGRPCTransport_Invoke` covers the wrapper directly.
- `pkg/engine/datasource/graphql_datasource`: existing tests pass against the migrated call site.

## Checklist

- [ ] Documentation has been updated on https://github.com/wundergraph/docs-website — N/A: no public-facing API change worth surfacing in docs at this stage; user-facing docs land with the consuming cosmo PR.
- [x] I have read the [Contributors Guide](https://github.com/wundergraph/cosmo/blob/main/CONTRIBUTING.md).

## Open Source AI Manifesto


v2/pkg/engine/datasource/grpc_datasource/grpc_datasource.go
```diff
@@ -91,6 +91,12 @@ func NewDataSource(client grpc.ClientConnInterface, config DataSourceConfig) (*D
 	}, nil
 }
 
+// NewDataSourceGRPC creates a new gRPC datasource using a gRPC ClientConnInterface.
+// This is a convenience function that wraps the connection in a grpcTransport.
+func NewDataSourceGRPC(client grpc.ClientConnInterface, config DataSourceConfig) (*DataSource, error) {
+	return NewDataSource(NewGRPCTransport(client), config)
+}
+
```
Noroth: It would be better to not have this function.
We can already call `NewDataSource(NewGRPCTransport()` and don't need `NewDataSourceGRPC`.

Otherwise we would have to define new wrapper functions for any client and maybe even client options. Let's not do that.

We should rather do something like this:
```go
dataSourceTransport := NewGRPCTransport(client)
grpcdatasource.NewDataSource(dataSourceTransport, config)
```

Noroth (changes requested):

v2/pkg/engine/datasource/grpc_datasource/grpc_datasource.go
fengyuwusong: Good point — agreed. Removed `NewDataSourceGRPC` in d16cca0d and updated all call sites
  (including the one in `graphql_datasource.go` and ~26 test sites) to use the explicit form:

  ```go
  grpcdatasource.NewDataSource(grpcdatasource.NewGRPCTransport(client), config)
  ```

  This keeps the transport composition explicit at the call site and avoids per-client wrapper
  proliferation.

fengyuwusong:
Hi @Noroth, I've addressed your review feedback in the latest commit (d16cca0):

Could you please take another look? Thanks!

Noroth (approved):

PR #1494, opened 2026-05-13
feat: add bytecode compilation and execution for GraphQL responses

TheApeMachine:
@coderabbitai summary

## Checklist

This commit introduces a new bytecode compilation and execution mechanism for GraphQL responses, enhancing performance and flexibility. It includes the implementation of various benchmarks and tests to ensure correctness and efficiency in processing GraphQL queries. New structures and functions for handling bytecode operations, including direct responses and nested object handling, have been added to improve the overall architecture of the engine.

> I just made this because of my answer on your form about what I would do about the code. I realized my answer was making a lot of assumptions, so I wanted to know if they were correct. Please ignore if this is not a direction you are interested in at this time :)

Using the final `-count=3` benchmark run on Apple M4 Max:

| Scenario                              |   Original    | Final Bytecode | Speedup |         B/op    |  allocs/op  |
|---------------------------|-----------:|---------------:|--------:|--------------:|-----------:|
| Loader only                         | 3845 ns/op |     3691 ns/op |   1.04x  | 3096 -> 2736 |   60 -> 51 |
| Flat query resolve               | 4661 ns/op |     4205 ns/op |   1.11x   | 4368 -> 4584 |   77 -> 44 |
| Nested static query            | 2945 ns/op |     1970 ns/op |   1.50x  | 4968 -> 4160  |   99 -> 33 |
| Batch entity query              | 3383 ns/op |     2885 ns/op |   1.17x  | 5811 -> 6228   |   98 -> 79 |
| Object entity query             | 2021 ns/op |     1799 ns/op  |   1.12x  | 3388 -> 4069  |   69 -> 56 |
| Nested batch entity query | 5946 ns/op |     4732 ns/op |   1.26x  | 7947 -> 8044  | 143 -> 121 |

Command used:

```sh
go test ./pkg/engine/resolve -run '^$' -bench 'BenchmarkResolveGraphQLResponse|BenchmarkLoaderLoadGraphQLResponseData' -benchmem -count=3
```

Interpretation: the best case is the static-shape nested query, where the bytecode direct renderer gets almost exactly the 1.5x target and cuts allocs by two thirds. Entity cases are more modest because they still pay ASTJSON merge/fetch-input costs, but they consistently reduce alloc count and still land around 1.12x-1.26x faster.

### DIFF
--- a/v2/pkg/engine/planbytecode/bench_test.go
+++ b/v2/pkg/engine/planbytecode/bench_test.go
@@ -1,0 +1,56 @@
+package planbytecode
+
+import (
+	"testing"
+
+	"github.com/wundergraph/astjson"
+	arena "github.com/wundergraph/go-arena"
+)
+
+var benchmarkSink int
+
+func BenchmarkJSONMergeASTJSON(b *testing.B) {
+	src := []byte(`{"id":"1","name":"Ada","reviews":[{"body":"ok","meta":{"score":10}},{"body":"fine","meta":{"score":8}}],"extra":true}`)
+	targetBytes := []byte(`{"id":"1"}`)
+	a := arena.NewMonotonicArena(arena.WithMinBufferSize(64 * 1024))
+	defer a.Release()
+
+	b.ReportAllocs()
+	b.SetBytes(int64(len(src)))
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		a.Reset()
+		target, err := astjson.ParseBytesWithArena(a, targetBytes)
+		if err != nil {
+			b.Fatal(err)
+		}
+		response, err := astjson.ParseBytesWithArena(a, src)
+		if err != nil {
+			b.Fatal(err)
+		}
+		merged, _, err := astjson.MergeValuesWithPath(a, target, response)
+		if err != nil {
+			b.Fatal(err)
+		}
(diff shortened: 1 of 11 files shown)


endigma:
Hi @TheApeMachine, wow, this is a pretty big change! Could you consider adding some more background to how your change works, your motivation, how you implemented it? 

Being completely honest my main concern with PRs like this is that it's 5000 new lines of code that our team has to become comfortable enough with what you've written to adopt it into our codebase and offer long term support for it. 

It's difficult to justify this based on purely the implementation and microbenchmarks (ns scale), so sharing some background would be great.

TheApeMachine:
@endigma Fully understand, and I will add some more background. However, just to make it clear, this was more in response to the question about what you would change, which was presented in the recruitment form. So please don't spend your time on this if this is not something you truly want/need to pursue. On the other hand, if performance is something on your roadmap, I believe there is a viable path for gains here. I will add the background in any case for completeness tomorrow.

endigma:
Hi @TheApeMachine, thanks for letting me know, I think this could be a good idea, but I don't think we'll be merging something like this at this time unless you've motivated your change with a real issue. 

Of course good luck with our hiring process!

devsergiy:
Not planned to be merged

Thanks,
Wundergraph Team

PR #1495, opened 2026-05-13
feat: make walker arena buffer size configurable and pool the walkers in collectPath

tylerscoville:
We have an absurdly large query due to legacy reasons (4,264 lines, 290 fragment definitions, 721 fragment spreads 😬). This on cold starts is causing the query planner to take ~60s to complete it on the first go-around. While we explore reducing query size and adding cache warming, I thought I would also explore making the planning itself faster. With these changes, I was able to get the planning time from ~60s down to ~15s.

The walker's arena was initialized with a 64-byte minimum buffer size, resulting in O(n²) allocation scans for large queries. Additionally, collectPath created and discarded a fresh walker (with a new arena) for every union/interface field rewrite.

- Add configurable ArenaMinBufferSize to Walker and plan. A 64-byte buffer is fine for small queries but causes O(n²) scanning for large queries (4000+ lines, hundreds of fragments). A 32KB buffer is better for large queries but wastes memory for small ones. Configurability lets operators match buffer size to their actual query patterns
- Reuse walkers via WalkerFromPool in collectPath instead of allocating new ones. CollectPath is called hundreds of times per query plan (once per union/interface field rewrite). Each call previously allocated a fresh walker with a new arena, then discarded it immediately.

Feel free to let me know how I can make this PR better!

@coderabbitai summary

## Checklist

## Open Source AI Manifesto


ysmolski:
Hello @tylerscoville. Thanks for making contribution to our project. Some things reduce the chances of this PR being accepted. For example, this PR has two fixes going on at the same time. Also, there are no issues raised in this repo where we could discuss if those are issues or non-issues. Thanks again.

tylerscoville:
@ysmolski, thanks for the reply. The go-arena bump fixed the issues I was experiencing!

ysmolski:
@tylerscoville I am pleased to hear this! Awesome!

PR #1500, opened 2026-05-18
fix(astprinter): preserve description indentation and break prose-bearing field args

endigma:


v2/pkg/ast/ast_description_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/ast"
+)
+
+func TestPrintDescription(t *testing.T) {
+	t.Run("preserves inner indentation beyond common indent", func(t *testing.T) {
+		var doc ast.Document
+		desc := doc.ImportDescription("Outer.\n    indented\n        deeper\n    back")
```
devsergiy: Maybe use a multiline string? It will be more visible what we are testing

v2/pkg/astprinter/astprinter.go
```diff
@@ -587,22 +587,45 @@ func (p *printVisitor) LeaveFieldDefinition(ref int) {
...
-	switch p.Ancestors[len(p.Ancestors)-1].Kind {
+	switch parent.Kind {
 	case ast.NodeKindDirectiveDefinition, ast.NodeKindInputObjectTypeDefinition, ast.NodeKindInputObjectTypeExtension:
 		p.writeIndented(p.document.InputValueDefinitionNameBytes(ref))
+	case ast.NodeKindFieldDefinition:
+		if multilineField {
+			for i := 0; i < argDepth; i++ {
+				p.write(p.indent)
+			}
+			p.write(p.document.InputValueDefinitionNameBytes(ref))
```
devsergiy: should we add an additional helper `writeIndentedWithDepth` in addition to `writeIndented` and `writeIndented` could use this helper internally?

devsergiy (changes requested):

devsergiy (approved):

PR #1505, opened 2026-05-20
fix: use remapped variables in cost calculation

ysmolski:
I have no idea how it slipped but I implemented Cost Control 
without any regard to remapped variables.

This PR fixes the situation. It adds an abstraction to deal with remapped
variables. Additionally, I have applied it in the resolver.


v2/pkg/engine/variables/variables.go
```diff
@@ -0,0 +1,47 @@
+// Package variables provides a request-scoped view over GraphQL operation
+// variables that transparently honors variable remapping introduced during
+// operation normalization.
+package variables
```
devsergiy: Is it really beneficial to have an additional package for this helper?
We most likely won't expand it

v2/pkg/engine/variables/variables.go
```diff
@@ -0,0 +1,47 @@
...
+
+import "github.com/wundergraph/astjson"
+
+// Set is a read-side view of request variables.
+//
+// The zero value is valid and behaves as an empty set: Get returns nil for any
+// name. Set is intended to be passed by value; it carries only two
+// pointers and is safe to copy.
+type Set struct {
```
devsergiy: Is it really a set? Is it the correct name for such a thing?

v2/pkg/engine/variables/variables.go
```diff
@@ -0,0 +1,47 @@
...
+
+// NewSet returns a Set that reads variable values from vars,
+// using remap (new -> old name) to translate post-normalization variable names
+// back to the original keys present in vars.
+// Either argument may be nil; a nil remap means no translation is performed.
+func NewSet(vars *astjson.Value, remap map[string]string) Set {
+	return Set{variables: vars, remap: remap}
+}
+
+// Get walks the variable tree along path. path[0] is the variable name and
```
devsergiy: We probably could say that the path is a keyed path in the variables json

v2/pkg/engine/variables/variables.go
```diff
@@ -0,0 +1,47 @@
...
+// is translated through the remap if an entry exists.
+// Subsequent elements are walked as nested keys on the resulting JSON value.
+// Returns nil if the set is empty, the path is empty, or any segment is missing.
+func (v Set) Get(path ...string) *astjson.Value {
```
devsergiy: q: What about the array path?

v2/pkg/engine/variables/variables.go
ysmolski: It is being used both in the planner and resolver subpackages of the engine package. I don't know the better place for it.

v2/pkg/engine/variables/variables.go
ysmolski: Top level of keys is a set, but for variables.

v2/pkg/engine/variables/variables.go
ysmolski: ok fixed

v2/pkg/engine/variables/variables.go
ysmolski: technically nothing prevent to use arrays in there, but there are not use cases for that ATM.

v2/pkg/engine/variables/variables.go
ysmolski: i moved it to resolve

v2/pkg/engine/variables/variables.go
ysmolski: And renamed to VariablesView

devsergiy (approved):

devsergiy (approved):

PR #1506, opened 2026-05-21
fix(websocket): prevent SIGSEGV in heartbeat goroutine when InitFunc returns (nil, err)

argoyle:
## Problem

In `ProtocolGraphQLTransportWSHandler.Handle`, the `connection_init` branch unconditionally calls `p.startHeartbeat(ctx)` even when `p.handleInit` returned an error. A common implementation pattern is for an `InitFunc` to return `(nil, err)` when authentication fails — at which point `Handle` proceeds to spawn the heartbeat goroutine on a **nil** context, and the next `<-ctx.Done()` panics with `invalid memory address or nil pointer dereference`. Because the panic happens in a goroutine started by the library, the process dies — there is no `recover` on the path.

Repro:

```go
protocol.initFunc = func(_ context.Context, _ InitPayload) (context.Context, error) {
    return nil, errors.New("auth failed")
}
err := protocol.Handle(ctx, engine, []byte(`{"type":"connection_init","payload":{"bad":"token"}}`))
// runtime panic before this line returns
```

We hit this in production: a frontend Apollo `GraphQLWsLink` with `lazy: true` opens a subscription before the user logs in, sends `connection_init` with an empty `authToken`, our `InitFunc` calls the auth provider on the empty token, the provider returns an error, we return `(nil, err)` — and the gateway pod crashes. Restart loop blocked all subscription traffic.

## Fix

`execution/subscription/websocket/protocol_graphql_transport_ws.go`:

1. After `closeConnectionWithReason`, **return the init error** instead of falling through to `startHeartbeat`. The connection has already been closed; there is no reason to start a heartbeat.
2. Close with `NewCloseReason(4401, "Unauthorized")` instead of `CompiledCloseReasonInternalServerError` (1011). The graphql-transport-ws spec uses 4401 for auth failure, and Apollo's `graphql-ws` client treats 4401 as terminal (no retry) while 1011 triggers exponential-backoff reconnects that pile up on a busy frontend.

Defensive `(nil, err)` from `InitFunc` is also arguably an implementation bug on the caller's side — but tolerating it costs nothing and removes the foot-gun.

## Test

`execution/subscription/websocket/protocol_graphql_transport_ws_test.go`: added a sub-test under `TestProtocolGraphQLTransportWSHandler_Handle` ("`should not panic and should close with 4401 when InitFunc returns (nil, err)`") that exercises the bug and asserts:
- `Handle` returns the init error
- `p.heartbeatStarted` is false (goroutine never started)
- the test client is disconnected

Before the fix the test panics. After, it passes.

## Affected versions

The unconditional `startHeartbeat` call has been in place at least since `execution/v1.12.0` and through `execution/v1.15.0`. No `replace` directive / fork required after this merges.

## Checklist

- [x] Tests added (regression test covers the panic path)
- [x] Follows project coding standards (gofumpt, govet clean on touched lines)
- [x] Behavior change documented in inline comments


endigma (approved):

endigma:
Hi @argoyle, thanks for your contribution!

PR #1513, opened 2026-05-22
fix(jsonschema): emit $ref/$defs for recursive input types

asoorm:
## Problem

`BuildJsonSchema` inlines input types and truncates recursion at a fixed depth, dropping self-referential fields (e.g. `left`/`right`). With `additionalProperties: false`, valid nested payloads are then rejected: `additional properties 'left', 'right' not allowed`. The same input passes normal GraphQL; only the generated JSON Schema (MCP path) rejects it.

Also fixes a tracker leak: the early `return nil` at the depth limit skipped cleanup, leaking the recursion counter across sibling uses of a type.

## Fix

Emit `$ref`/`$defs` for recursive input types: a first pass finds self-/mutually-recursive types; each is emitted once under root `$defs` and referenced via `$ref`. Recursion terminates by construction and nesting works to any depth. Non-recursive types are unchanged (still inlined). `maxRecursionDepth` is kept on the exported signatures for compatibility but is now a no-op.

## Design note

Revisits #1124, which deliberately dropped `$def` for LLM compatibility. That holds for non-recursive inputs (kept inline); for recursive inputs the only alternative to `$ref` is a schema that rejects valid data, so `$defs` is reintroduced for recursive types only. The validator (`santhosh-tekuri/jsonschema`) supports it.

## Tests

- New `recursive_input_test.go`: validates a nested recursive payload against the generated schema using `santhosh-tekuri/jsonschema/v5`.
- Updated the two tests that encoded the old truncation behaviour; non-recursive golden tests unchanged.

Closes ENG-9631


v2/pkg/engine/jsonschema/variables_schema.go
```diff
@@ -30,15 +34,20 @@ func NewVariablesSchemaBuilder(operationDocument, definitionDocument *ast.Docume
 	return NewVariablesSchemaBuilderWithOptions(operationDocument, definitionDocument, 3)
 }
 
-// NewVariablesSchemaBuilderWithOptions creates a new VariablesSchemaBuilder with custom options
+// NewVariablesSchemaBuilderWithOptions creates a new VariablesSchemaBuilder.
+//
+// Deprecated: maxRecursionDepth is accepted for backwards compatibility but is no
+// longer used. Recursive input types are represented via "$ref"/"$defs", which
+// terminates by construction and supports arbitrary nesting depth.
```
ysmolski: Let's remove maxRecursionDepth since it is used mostly likely internally only.

v2/pkg/engine/jsonschema/schema.go
```diff
@@ -27,6 +27,13 @@ type JsonSchema struct {
 	Description          string                 `json:"description,omitempty"`
 	Nullable             bool                   `json:"nullable,omitempty"`
 
+	// Ref references a schema defined under the root "$defs" (e.g. "#/$defs/MyInput").
+	// Used to represent recursive input types, which cannot be inlined.
+	Ref string `json:"$ref,omitempty"`
+	// Defs holds reusable schema definitions, referenced via Ref. Only populated
+	// on the root schema.
+	Defs map[string]*JsonSchema `json:"$defs,omitempty"`
+
```
ysmolski, line 36: Who is the user of those fields?

ysmolski (changes requested):

v2/pkg/engine/jsonschema/variables_schema.go
asoorm: Done

v2/pkg/engine/jsonschema/variables_schema.go
ysmolski: stale comment

ysmolski: I guess also stale.

ysmolski (approved):

PR #1521, opened 2026-05-29
fix: calculate costs for abstract fields without double counting

ysmolski:
This PR fixes how costs for abstract types are calculated. Especially for fragments used on abstract types.

For example, for actual costs, instead of summing each type of the union, engine sums only costs for types actually seen in the returned results.

For estimated costs, instead of summing all the implementing nodes of abstract types, we just pick the implementing type with the maximum cost and add that to the costs of fields selected on the abstract type itself.

I had to recalculate and verify numbers in tests.

Unfortunately, this fix is not final. There is unfinished work with how the cost of the field itself is selected. Right now we pick the maximum weight across implementing types and dataSources independently from children costs selection. This concerns the estimation the most. The better approach would be to be tie this process with the selection of the type with maximum cost.


ysmolski:
@coderabbitai review

ysmolski:
@claude review once

devsergiy (approved):

PR #1535, opened 2026-06-11
feat(engine): plan-time schedule-tree scheduler + runtime dataflow executor

jensneuse:
# feat(engine): plan-time schedule-tree scheduler + runtime dataflow executor

Two independent, default-off engine execution improvements that collapse
federated query latency to the dependency-graph critical path under skewed
subgraph latency, plus hot-path allocation reductions.
Measured on the gateways-benchmark workload (one slow subgraph):
**−37% median latency** for either feature vs the shipping wave executor.

## What's in here (4 commits)

1. **perf(resolve): pre-size buffers and slab-allocate parallel fetch results**
   Single-slab result allocation in `resolveParallel`, HTTP request buffer
   pre-sizing, singleflight key buffer reuse, resolvable field-path buffer
   pre-sizing. Verified at parity on a flags-off A/B (+1.2%, within noise).

2. **feat(postprocess): dominance-based schedule-tree scheduler (`WithBuildScheduleTree`)**
   Ports the `improve-query-order` scheduler with two load-bearing fixes found
   by adversarial review against real planner output:
   - `validateParallelLeaves` trusts FetchID dependency edges ONLY. The
     path-containment heuristics false-positive on the canonical federation
     plan (a fetch attaching under an ancestor-provided subtree is independent
     of a sibling merging into a prefix path — the planner emits explicit
     edges when a dependency is real, e.g. `deps [1,2]` on the fetch that
     actually reads sibling data) and panicked every affected request.
   - `buildScheduleTreeProcessor` never panics: any scheduler/validator error
     falls back to the legacy `orderSequenceByDependencies` +
     `createParallelNodes` pipeline, planning exactly as with the option off.
   Includes the property/scenario suites, a 13-fetch real-plan schedule pin
   (`TestBuildScheduleTreeRealGatewaysBenchmarkPlan`), and the ADR.

3. **feat(resolve): 3-phase prepare/load/merge protocol for nested schedule trees**
   The executor support for nested plans, with three deviations, each pinned
   by a regression test that fails if the deviation is reverted:
   - flat Parallels inside a nested tree route through the locked nested path
     (`useMergeMu ||` gate) — fixes a verified data race;
   - skip-on-errored-dependency gated on `useMergeMu` — flat-plan error-path
     bytes stay identical to the legacy executor;
   - pooled batch tools registered for `Put` BEFORE the prepare-error return —
     fixes a pooled-arena leak on failed batch prepares.

4. **feat(resolve): dataflow executor (`ResolverOptions.EnableDataflowExecution`)**
   Per-FetchID dependency gating for flat query plans: each fetch's network
   load starts when its OWN dependencies have merged. Nine falsifiable
   invariants documented in the file header; the load-bearing ones:
   - every arena access runs on the coordinator goroutine (workers execute
     only the arena-free source load) — no mutex, race-free by construction;
   - the three audited error sinks are swap-captured per leaf and replayed in
     leaf order — error/extension output is byte-identical to the wave
     executor;
   - dispatch pops the global leaf-index minimum, so pre-fetch hook call order
     (authorizer/rate-limiter) matches wave spawn order whenever the DAG
     permits;
   - nested (schedule-tree) plans are structurally rejected and fall back —
     the two features compose safely.

## Default-off guarantee

With both options off, behavior is byte-identical and at performance parity:
- byte-identity gate: one sha256 across all four flag modes, 50 requests each;
- flags-off A/B vs the unmodified base: within noise;
- the full `-race` suites for resolve/postprocess/graphql_datasource pass with
  zero exclusions, including `-count=50` stress on the byte-identity and race
  stress tests.

## Known properties (documented in code)

- Nested executor appends errors/extensions in completion order across
  parallel branches (error-path only; tree-order sink staging is the named
  follow-up, and the dataflow executor already implements that staging).
- Dataflow merge-CONFLICT error attribution on type-conflicting sibling data
  (invalid federation data) follows completion order.

## Review provenance

The implementation survived three adversarial codex review cycles to
GATE-CLEAN during development plus a final whole-branch review (GATE-CLEAN);
every P1 found along the way is fixed and pinned by an empirically falsified
regression test.

### DIFF
--- a/docs/adr/0001-query-plan-scheduler.md
+++ b/docs/adr/0001-query-plan-scheduler.md
@@ -1,0 +1,517 @@
+# ADR 0001: Query Plan Scheduler
+
+Status: Accepted.
+Date: 2026-05-09.
+
+## Context
+
+The query planner emits fetches as a flat sequence and then runs postprocessors that sort by dependencies and group eligible siblings into parallel nodes.
+That design is simple, but it leaves latency on the table when independent dependency chains coexist in the same operation.
+The motivating shape is a query that fetches `me` from one subgraph, fetches `organisations` from another subgraph, and then fetches `currentPractice` through an entity hop that depends only on `me`.
+The old postprocessor emits `Sequence(Parallel(user, organisation), practice)`.
+That forces `practice` to wait for `organisation`, even though `organisation` does not provide any dependency required by `practice`.
+The better tree is `Parallel(Sequence(user, practice), organisation)`.
+That tree has the critical path `max(user + practice, organisation)` instead of `max(user, organisation) + practice`.
+
+The old executor could not safely run that better tree.
+Its parallel implementation assumed every child of a `Parallel` node was a leaf fetch.
+It preselected merge targets, ran network loads concurrently, and then merged results serially after all loads completed.
+If a `Parallel` child became a nested `Sequence`, the loader would have to recurse inside a goroutine.
+That would make selection, input rendering, merge, JSON arena writes, and tainted-object map writes happen concurrently.
+Those operations are not thread safe.
+
+The scheduler therefore had two coupled requirements.
+First, the executor needed to support nested sequence and parallel fetch trees without racing shared response state.
+Second, the postprocessor needed a dependency-aware tree builder that could emit nested plans while preserving existing default behavior until the new path is explicitly enabled.
+
+## Decision
+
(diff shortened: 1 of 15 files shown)


ysmolski:
First part was resolved by https://github.com/wundergraph/graphql-go-tools/pull/1594.

As for the 2nd part -- we are not ready for this complexity yet.

PR #1536, opened 2026-06-12
fix(resolve): fix GC-liveness bug in the loadBatchEntityFetch

ysmolski:
The arena memory is invisible to GC. The method was allocating 1 element slice on the Go heap and referenced from the arena bytes. At some point GC collected that slice and pointer was poisoned.

-race was indicating the problem very well.

The fix is to allocate the slice in the arena memory, so its lifetime is tied to the lifetime of pooled objects.


v2/pkg/engine/resolve/resolve_test.go
```diff
@@ -8021,6 +7763,41 @@ func Benchmark_NestedBatchingArena(b *testing.B) {
 	})
 }
 
+// TestNestedBatching_GCPressure is a regression test for ENG-9717.
+// The batch entity fetch used to store heap-allocated target buckets into arena memory,
+// hiding them from the GC.
+func TestNestedBatching_GCPressure(t *testing.T) {
+	prevGC := debug.SetGCPercent(1)
+	defer debug.SetGCPercent(prevGC)
```
Noroth: You could use `t.Cleanup`

ysmolski: fixed

devsergiy (approved):

PR #1540, opened 2026-06-16
feat: Apollo-compatible null variable handling and abstract type cost defaults

damian-rakus:
## Summary

This PR makes two changes to align our GraphQL gateway behavior with Apollo Router:

1. **Null variable coercion with defaults**: When a nullable variable is explicitly set to `null` and used in a non-null argument position that has a schema default value, the variable reference is "split" so the downstream subgraph treats it as "not provided" and applies the schema default. This matches Apollo Router's behavior where schema defaults take precedence over explicit null for non-null arguments.

2. **Abstract type cost calculation**: Uses the field's return-type default weight (scalar=0, object=1) for abstract-type selections instead of finding the maximum `@cost` weight across all implementing types. This matches Apollo Router's cost calculation behavior.

Both changes are gated behind compile-time feature flags in `mondaytweaks` packages (one per dependency layer) so they can be toggled off without removing code.

## Changes

**Null variable coercion** (`v2/pkg/astnormalization/`):
- Added `variables_null_coercion_with_defaults.go` — new AST visitor that detects null variables in non-null+default argument positions and splits them into new "not provided" variables
- Added `variables_null_coercion_with_defaults_test.go` — 11 test cases covering happy paths, no-ops, edge cases, and full normalizer integration
- Wired the visitor into both `OperationNormalizer` and `VariablesNormalizer` pipelines, gated by `mondaytweaks.CoerceNullVariablesWithDefaults`
- Added `astnormalization/mondaytweaks/mondaytweaks.go` — feature flag package for normalization layer (avoids circular dep with engine)

**Abstract type cost defaults** (`v2/pkg/engine/`):
- Modified `plan/cost.go` to skip max-weight-across-implementors logic when `mondaytweaks.UseInterfaceDefaultCostForAbstractTypes` is true
- Added `engine/mondaytweaks/mondaytweaks.go` — feature flag package for engine layer
- Updated cost test expectations in `execution/engine/execution_engine_cost_test.go`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ pkg/parser/testdata/lotto.graphql
 *node_modules*
 *vendor*
 .serena
+.omc/
--- a/execution/engine/execution_engine_cost_test.go
+++ b/execution/engine/execution_engine_cost_test.go
@@ -243,8 +243,8 @@ func TestExecutionEngine_Cost(t *testing.T) {
 					),
 				},
 				expectedResponse:      `{"data":{"hero":{"name":"Luke Skywalker","height":"12"}}}`,
-				expectedEstimatedCost: intPtr(22), // Query.hero (2) + Human.height (3) + Droid.name (17=max(7, 17))
-				expectedActualCost:    intPtr(22),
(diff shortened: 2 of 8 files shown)


damian-rakus:
Apologies, opened PR by mistake

PR #1549, opened 2026-06-25
fix(postprocess): memoize transitive fetch dependencies in orderSequenceByDependencies

arutkowski00:
@coderabbitai summary

## Summary

`orderSequenceByDependencies.ProcessFetchTree` sorts the fetch-tree child nodes with `slices.SortFunc`. For every pairwise comparison it called `nodeDependsOn` **twice**, and `nodeDependsOn` re-derived a node's full transitive dependency set by recursing with **no memoization**, using `nodeByFetchID` — an **O(N) linear scan** of `root.ChildNodes` — for each dependency lookup.

For densely-connected fetch trees this becomes **exponential**. The pathological case is a mutation with many aliased root fields where fetch `i` depends on `[0..i-1]`: the comparator re-walks an ever-growing dependency chain on every comparison.

## Symptom

Planning CPU blows up on operations that produce densely-connected fetch trees. Measured against a real federated config:

- **28 aliased mutation root fields: 27.7s** of planning time (planning was ~100% of request latency in the affected region).

## Root cause

Unmemoized recursive transitive-dependency resolution + linear node lookup in `nodeDependsOn`, invoked twice per sort comparison → **O(2^N)**.

## Fix

Precompute, **once per `ProcessFetchTree` call**:

1. a `fetchID -> *FetchTreeNode` index map (replaces the O(N) `nodeByFetchID` scan), and
2. a memoized `map[int][]int` transitive-dependency map via memoized DFS, with an in-progress cycle guard.

The `SortFunc` comparator then reads the precomputed sets. The comparator logic is **byte-identical** to before and **output ordering is preserved**. The now-unused `nodeDependsOn` / `nodeByFetchID` methods are removed.

This removes the **O(2^N)** recomputation; the per-comparison cost now reduces to cache reads. (Building the cache still materializes and sorts transitive closures, so dense graphs cost more than O(N+E) — but the exponential blow-up is gone.)

## Measured impact

- 28 aliased mutation root fields: **27.7s -> 1.6ms**
- N=200 nodes: stays at **73ms**

## Regression risk: LOW

Pure internal refactor. Ordering is preserved, all existing `postprocess` tests pass, and a new dense-chain regression test (N=31, node `i` depends on `[0..i-1]`, asserting ascending fetch-ID ordering) is added — it does not terminate against the old code but completes instantly with the fix.

## Tests

`cd v2 && go test ./pkg/engine/postprocess/...` — all pass (`go vet` clean, package builds).

## Checklist

---

_Contributed on behalf of monday.com. If a CLA is required for this contribution, please let us know and we will complete it._

### DIFF
--- a/v2/pkg/engine/postprocess/order_sequence_by_dependencies.go
+++ b/v2/pkg/engine/postprocess/order_sequence_by_dependencies.go
@@ -15,12 +15,60 @@ func (o *orderSequenceByDependencies) ProcessFetchTree(root *resolve.FetchTreeNo
 	if o.disable {
 		return
 	}
+	// Precompute, once per call, a fetchID->node index so dependency resolution
+	// no longer needs an O(N) linear scan of root.ChildNodes per lookup.
+	nodeByID := make(map[int]*resolve.FetchTreeNode, len(root.ChildNodes))
+	for _, node := range root.ChildNodes {
+		nodeByID[o.nodeFetchID(node)] = node
+	}
+	// deps caches, per fetch ID, the recursively-resolved transitive dependency
+	// set (sorted, deduped, excluding the node's own ID). Computing this once with
+	// a memoized DFS turns what used to be an O(2^N) recursion (re-derived twice
+	// per comparison inside SortFunc) into O(N+E).
+	deps := make(map[int][]int, len(root.ChildNodes))
+	inProgress := make(map[int]bool, len(root.ChildNodes))
+	var resolveDeps func(id int) []int
+	resolveDeps = func(id int) []int {
+		if cached, ok := deps[id]; ok {
+			return cached
+		}
+		if inProgress[id] {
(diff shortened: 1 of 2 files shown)


arutkowski00:
Addressed the complexity nitpick: reworded the dependency-resolution comments (and the PR description) to state current behaviour instead of claiming an O(N+E) bound — the memoized DFS removes the O(2^N) per-comparison recomputation, but materializing/sorting the closures still costs more than O(N+E) on dense graphs. Also fixed the CI failures: shortened the PR title under the 100-char commitlint limit and modernized the regression-test loop (`for i := range n`) for golangci-lint.

ysmolski:
@arutkowski00 I have created https://github.com/wundergraph/graphql-go-tools/pull/1657 that will supersede this PR.

PR #1559, opened 2026-07-01
fix(resolve): gate SUBREQUEST_HTTP_ERROR on mondaytweaks flag to avoid exposing subgraph names

damian-rakus:
## Summary
Prevents subgraph architecture details from leaking to clients by gating the `SUBREQUEST_HTTP_ERROR` Apollo Router compatibility path behind a new `mondaytweaks.ApolloRouterCompatibilitySubrequestHTTPError` feature flag. Without this guard, the error code was always emitted when the loader-level flag was set, potentially revealing internal service names via structured error extensions.

## Changes
- **`v2/pkg/mondaytweaks/mondaytweaks.go`**: Added `ApolloRouterCompatibilitySubrequestHTTPError = true` constant to control the feature globally via monday's tweak flags
- **`v2/pkg/engine/resolve/loader.go`**: Added `mondaytweaks.ApolloRouterCompatibilitySubrequestHTTPError &&` guard in `renderErrorsFailedToFetch` (new path) and `renderErrorsStatusFallback` (existing path) so the SUBREQUEST_HTTP_ERROR code is only attached when both the tweak flag and the loader-level flag are enabled
- **`v2/pkg/engine/resolve/loader_arena_gc_test.go`**: Added `_errorWithStatusCodeDataSource` test helper that simulates transport-level failures with a status code but no response body
- **`v2/pkg/engine/resolve/resolve_arena_gc_test.go`**: Added `TestArenaGCSafety_ApolloRouterCompatError_FailedToFetch` covering the new `renderErrorsFailedToFetch` path under GC pressure

## Monday.com Item
https://monday.monday.com/boards/8222767873/views/182723743/pulses/12250997381


damian-rakus:
Apologies, agent created PR to the original repo instead of fork

PR #1509, opened 2026-05-21
feat(grpc_datasource): add ConnectRPC transport implementation

fengyuwusong:
## Summary

Add a ConnectRPC implementation of the `RPCTransport` interface introduced in #1490, so the gRPC data source can dial Connect subgraphs without changing the compiler or JSON builder. The same `MockService` fixture is regenerated via `buf` to emit Connect handlers alongside the existing gRPC code, and a new end-to-end test exercises the data source over Connect against that fixture.

This is the second of the two PRs that together replace #1480.

## Related

- Builds on: #1490 (`RPCTransport` interface + `GRPCTransport` implementation)
- Replaces: #1480 (the original combined PR, now split into #1490 + this one)
- Consumed by: wundergraph/cosmo#2808 (the router-side feature that motivated this work)

### `pkg/grpctest` — buf-driven fixture

- New `buf.yaml` / `buf.gen.yaml`: drive `protoc-gen-go`, `protoc-gen-go-grpc`, and `protoc-gen-connect-go` from a single `buf generate` invocation, keeping the existing import path (`graphql-go-tools/v2/pkg/grpctest/productv1`) via `M=` overrides.
- `Makefile`: `generate-proto` now calls `buf generate`; `install-protoc` also installs `protoc-gen-connect-go`.
- New `productv1/productv1connect/product.connect.go`: buf-generated Connect handler/client. Sibling of the existing `productv1/product_grpc.pb.go`, so the package serves Connect, gRPC, and gRPC-Web from one HTTP endpoint.
- New `mockservice_connect.go`: hand-maintained adapter (`MockServiceConnect`) wrapping the existing `MockService` onto `productv1connect.ProductServiceHandler`. Same backing implementation, two transports.

### `pkg/engine/datasource/grpc_datasource` — ConnectRPC transport

- New `transport_connect.go`: `NewConnectTransport(ConnectTransportConfig{BaseURL, Encoding, Interceptors, HTTPClient})` implements `RPCTransport.Invoke` on top of `connectrpc.com/connect`. Supports `ConnectEncodingProtobuf` ("proto") and `ConnectEncodingJSON` ("json") wire formats. Translates `grpc/metadata` to and from Connect headers using the same conventions as `GRPCTransport`, with binary request headers (`-bin` suffix) base64-encoded per the Connect spec.
- The `Interceptors` field exposes connect-go's interceptor chain so callers can plug in tracing, logging, retries, or auth header injection without wrapping the HTTPClient.
[...]

PR #1577, opened 2026-07-06
feat: add new visitor for inline arguments validation

SkArchon:
This PR adds a visitor which can be registered by the router, and is used to validate inline arguments. And on top of that allow setting of inline argument values for printing with extensions (this is in case of non-enforced).

@coderabbitai summary

## Checklist

## Open Source AI Manifesto


v2/pkg/astnormalization/inline_arguments.go
```diff
@@ -0,0 +1,125 @@
...
+type InlineArgumentsValidationOptions struct {
+	Enforce      bool
+	ErrorMessage string
+	ErrorCode    string
+	StatusCode   int
+}
+
+type InlineArgumentsValidator struct {
+	Options  InlineArgumentsValidationOptions
+	Findings []InlineArgument
```
Noroth: Just call it `InlineArguments` instead of Findings maybe?

v2/pkg/astnormalization/inline_arguments.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func (v *InlineArgumentsValidator) ClearFindings() {
+	if v == nil {
+		return
+	}
+	v.Findings = v.Findings[:0]
+}
+
+func (v *InlineArgumentsValidator) HadInlineArguments() bool {
```
Noroth: Past tense sounds a bit off, would recommend to just call it `HasInlineArguments` because it's not over yet 😀

SkArchon: This is done in unc (o *OperationKit) collectInlineArguments(result *astnormalization.NormalizationResult) { now

v2/pkg/astnormalization/inline_arguments.go
SkArchon: Done

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -366,7 +366,7 @@ func (p *Planner[T]) ConfigureFetch() resolve.FetchConfiguration {
 			return resolve.FetchConfiguration{}
 		}
 
-		if p.rpcTransport == nil {
+		if p.rpcTransport == nil && !p.config.grpc.Disabled {
```
SkArchon: This was breaking the test "Should not be able to call plugin if it is not enabled", since the error message and status code was different. Unsure if this is the correct fix or changing the error message and status. @Noroth 

https://github.com/wundergraph/cosmo/blob/c441f55485731c4005553b92af84ec2d83dbd2af/router-tests/protocol/router_plugin_test.go#L53-L72

This was from this PR, but there is no router equivalent PR to catch it: https://github.com/wundergraph/graphql-go-tools/pull/1509

v2/pkg/engine/datasource/grpc_datasource/grpc_datasource.go
```diff
@@ -113,7 +113,12 @@ func (d *DataSource) Load(ctx context.Context, headers http.Header, input []byte
...
 
 	if d.disabled {
-		return builder.writeErrorBytes(fmt.Errorf("gRPC datasource needs to be enabled to be used")), nil
+		return builder.writeErrorBytes(fmt.Errorf("gRPC / connect datasource needs to be enabled to be used")), nil
+	}
+
+	// If the transport is nil we will return the following error message instead
+	if d.transport == nil {
+		return builder.writeErrorBytes(fmt.Errorf("gRPC / connect configuration requires an rpc transport")), nil
 	}
```
Noroth, line 122: Instead of writing error bytes return an error here, if we are at this stage, something is really wrong

v2/pkg/astnormalization/inline_arguments.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/astvisitor"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/lexer/position"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/operationreport"
+)
+
+// InlineArgument describes a single argument in an operation whose value was
+// supplied inline (a literal) instead of as a variable.
+type InlineArgument struct {
+	ArgumentName  string
+	EnclosingName string
```
devsergiy: maybe `AncestorName` and `AncestorKind`?

v2/pkg/astnormalization/inline_arguments.go
```diff
@@ -0,0 +1,116 @@
...
+	EnclosingKind ast.NodeKind
+	ValueKind     ast.ValueKind
+	Position      position.Position
+}
+
+func (a InlineArgument) QualifiedName() string {
+	switch a.EnclosingKind {
+	case ast.NodeKindField:
+		return a.EnclosingName + "." + a.ArgumentName
```
devsergiy: for a field should we also show Typename?

like `Query.users#id`

field name itself not so informative

maybe even more useful will be a query path
`query.users.friends#filter`

v2/pkg/astnormalization/astnormalization.go
```diff
@@ -384,12 +401,25 @@ func (o *OperationNormalizer) NormalizeOperation(operation, definition *ast.Docu
...
-// NormalizeNamedOperation applies all registered rules to one specific named operation in the AST
 func (o *OperationNormalizer) NormalizeNamedOperation(operation, definition *ast.Document, operationName []byte, report *operationreport.Report) {
+	o.NormalizeNamedOperationWithResult(operation, definition, operationName, report, RunOptions{})
+}
+
+func (o *OperationNormalizer) NormalizeNamedOperationWithResult(
+	operation, definition *ast.Document,
+	operationName []byte,
+	report *operationreport.Report,
+	runOpts RunOptions,
```
devsergiy, line 412: should options be shaped in the same way as regular normalizer options? e.g. `With`

v2/pkg/astnormalization/inline_arguments.go
```diff
@@ -0,0 +1,116 @@
...
+		return
+	}
+
+	finding := InlineArgument{
+		ArgumentName: v.operation.ArgumentNameString(ref),
+		ValueKind:    valueKind,
+		Position:     v.operation.Arguments[ref].Position,
+	}
+
+	if len(v.Ancestors) > 0 {
```
devsergiy: you could use `(w *Walker) Ancestor()` here

v2/pkg/engine/resolve/resolvable.go
```diff
@@ -1038,6 +1076,9 @@ func (r *Resolvable) hasExtensions() bool {
 	if r.ctx.ExecutionOptions.IncludeQueryPlanInResponse {
 		return true
 	}
+	if len(r.ctx.InlineArguments) > 0 {
```
devsergiy, line 1079: question: do we have to render this in the engine itself?

we print here any generic extensions - we could pass them in the router

devsergiy (changes requested):

v2/pkg/engine/resolve/resolvable.go
SkArchon: We have two cases, one where
* We need to print extensions along with data (in case we are using the warning level), which is printed as part of the engine (from what I understood)
* We need to print them as errors directly, for this we use the router

devsergiy: yes you are right, currently there is no way to do that in a generic way

v2/pkg/astnormalization/astnormalization.go
SkArchon: I was thinking we could keep it as it is as I dont anticipate more options here, wdyt?

devsergiy (approved):

PR #1588, opened 2026-07-14
feat: improve abstract field validation

endigma:
Also incorporates a fix to one of the execution test runners, we cannot run `t.Parallel()` inside because the whole thing is slightly flawed where the wrong `t` is captured for `must…(t, …)` type invocations on fields of the TC struct.

This is normally not an issue as the `must…` helpers are usually just ceremony for dropping the error value, but when we might actually expect them to fail they will panic by calling `require.XXX(t, …)` with the parent `t` instead of the subtest `t`. 

This can be resolved by in these cases evaluating it and all of its arguments within the subtest closure `func(t) { … }` but then the `tparallel` linter complains about missing parallel, which cannot be fixed because the inner function also tried to call it.

The proper way to fix this would be to have `dataSources` etc fields on the TC be evaluated with subtest `t` via like `dataSources: func (t testing.T) []*DataSource` instead of directly providing it, but this would be a very intrusive change.


execution/engine/abstract_type_validation_test.go
```diff
@@ -0,0 +1,298 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/wundergraph/graphql-go-tools/execution/graphql"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/datasource/graphql_datasource"
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/plan"
+)
+
+func TestAbstractTypeValidation(t *testing.T) {
+	t.Parallel()
```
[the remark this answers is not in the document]
endigma: Could it just be moved out of the closure instead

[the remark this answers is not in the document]
endigma: 💎 great job coderabbit

execution/engine/execution_engine_cost_test.go
```diff
@@ -296,7 +296,7 @@ func TestExecutionEngine_Cost(t *testing.T) {
 		// When the subgraph resolves a single (non-list) abstract field and does NOT return __typename,
 		// we must still record one occurrence for that field's path, falling back to the declared
 		// abstract type name in actual costs.
-		t.Run("single abstract field without __typename takes into account implementing types", runWithoutError(
```
ysmolski, line 299: the comment above is outdated now

execution/engine/execution_engine_cost_test.go
```diff
@@ -1313,7 +1316,8 @@ func TestExecutionEngine_Cost(t *testing.T) {
 							customConfig,
 						),
 					},
-					expectedResponse:      `{"data":{"hero":{"name":"Luke","friends":[{"name":"Leia"}]}}}`,
+					// friends items carry no __typename, so they are rejected and nulled
+					expectedResponse:      `{"errors":[{"message":"Subgraph 'id' returned an invalid value for __typename field.","path":["hero","friends",0],"extensions":{"code":"INVALID_GRAPHQL"}}],"data":{"hero":{"name":"Luke","friends":[null]}}}`,
 					expectedEstimatedCost: intPtr(20), // 2 + 1*(0 + 6*(3 + 1*0))
 					expectedActualCost:    intPtr(5),  // 2 + 1*(0 + 1*(3 + 1*0))
```
ysmolski: `// 2 + 1*(0 + 1*3)`

execution/engine/execution_engine_helpers_test.go
```diff
@@ -110,7 +110,7 @@ func createTestRoundTripper(t *testing.T, testCase roundTripperTestCase) testRou
 				receivedBodyBytes, err = io.ReadAll(req.Body)
 				require.NoError(t, err)
 			}
-			require.Equal(t, testCase.expectedBody, string(receivedBodyBytes), "roundTripperTestCase received unexpected body")
+			assert.Equal(t, testCase.expectedBody, string(receivedBodyBytes), "roundTripperTestCase received unexpected body")
```
ysmolski, line 113: I am not sure about this change. Can you explain why it happened?

execution/engine/execution_engine_test.go
```diff
@@ -63,7 +63,6 @@ func mustFactory(t testing.TB, httpClient *http.Client) plan.PlannerFactory[grap
 
 func runExecutionTest(testCase ExecutionEngineTestCase, withError bool, expectedErrorMessage string, options ...executionTestOptions) func(t *testing.T) {
 	return func(t *testing.T) {
-		t.Parallel()
```
ysmolski, line 66: Why it was removed?

execution/engine/execution_engine_cost_test.go
endigma: I don't understand

execution/engine/execution_engine_helpers_test.go
endigma: You cannot use `require` in goroutines, only in the main test function, it calls `runtime.Goexit()` which will be the wrong one if not the top level test.

execution/engine/execution_engine_test.go
endigma: This whole runner function is very buggy and likes to panic when you use it because it calls `t.Parallel` then uses the wrong `t` in callsites for `must...` functions. Fixing it requires wrapping it in a closure, but then `tparallel` linter complains, so it is one or the other.

execution/engine/execution_engine_cost_test.go
ysmolski: removing should work

ysmolski (approved):

ysmolski:
LGTM

PR #1594, opened 2026-07-16
feat: implement multi fetch to the same subgraph

devsergiy:
Adds an opt-in MultiFetch optimization: entity fetches that 
target the same subgraph and execute in the same parallel wave 
are merged into a single upstream request, 
cutting the number of subgraph round-trips.

How it works

- Planner: entity fetches record their normalized upstream operation,
variables, and request envelope as merge material.

- Postprocess: the new createMultiFetch stage groups same-datasource entity fetches 
within each parallel wave and merges them into one MultiEntityFetch. 
Member operations are combined via AST merging with collision-safe variable renaming; 
each member becomes an aliased _entities field (f1, f2, …) guarded by @include(if: $includeFN). 

- Loader: a merged fetch renders per-entry representations, sends one request, 
then demuxes the response by alias; per-entry data merging, error partitioning and path rewriting.
Entries excluded at prepare time (denied, no items) ship includeFN:false and 
an empty representations array, so variable coercion still passes upstream.

Defaults

Off by default at both gates: the postprocess stage is the pipeline's only opt-in stage,
and without the planner flag no merge material is recorded. 

---------
Co-authored-by: Claude Fable 5 <noreply@anthropic.com>
Co-authored-by: Alberto Garcia Hierro <alberto@wundergraph.com>
Co-authored-by: Yury Smolski <140245+ysmolski@users.noreply.github.com>


v2/pkg/engine/plan/configuration.go
```diff
@@ -20,7 +20,14 @@ type Configuration struct {
 	DisableResolveFieldPositions bool
 	// EnableOperationNamePropagation appends the operation name from nested operations
 	EnableOperationNamePropagation bool
-	CustomResolveMap               map[string]resolve.CustomResolve
+
+	// EnableMultiFetch records planner artifacts on entity fetches so the
```
ysmolski, line 24: What are the planner artifacts? :-D

docs/multi-fetch/plan.md
```diff
@@ -0,0 +1,752 @@
+# MultiFetch Implementation Plan
```
ysmolski: We should not include this.

docs/multi-fetch/review-plan.md
```diff
@@ -0,0 +1,216 @@
+# MultiFetch Review Follow-ups — Implementation Plan
```
ysmolski: We should not include this.

docs/multi-fetch/design-v2.md
```diff
@@ -0,0 +1,173 @@
+# MultiFetch v2 design — deferred input rendering and merge-stage relocation
```
ysmolski: These are not for humans right? And they duplicate the code changes in PR itself, but probably not equal, so they are outdated already. And after a while they will be outdated even more. And future LLMs will read these files and get confused. I think we should not include these temporary "artifacts" into our code base.

v2/pkg/astimport/astimport_test.go
```diff
@@ -375,3 +376,76 @@ func TestImportVariableDefinitionPreservesBlockStringFlag(t *testing.T) {
...
+	})
+
+	printed, err := astprinter.PrintString(to)
+	require.NoError(t, err)
+
+	assert.Contains(t, printed, "_entities(representations: $representations_f1)")
+	assert.Contains(t, printed, "p: products(first: $first_f1)")
+	assert.Contains(t, printed, "@custom(arg: $first_f1)")
+	assert.Contains(t, printed, "... on Employee")
+	assert.Contains(t, printed, "nested {id}")
```
ysmolski: Please confirm that these checks are sufficient.

v2/pkg/engine/postprocess/create_multi_fetch.go
```diff
@@ -0,0 +1,444 @@
...
+			isToken = false
+			continue
+		}
+		isToken = true
+	}
+	return indices
+}
+
+// minID returns the lowest member fetch ID, used as the merged fetch's survivor ID.
+func minID(ids []int) int {
```
ysmolski: [https://pkg.go.dev/slices#Min](<https://pkg.go.dev/slices#Min>)

and btw this is the example of comment that does not mean much at all.

ysmolski (commented):
partial review

v2/pkg/astimport/astimport_test.go
devsergiy: actually better to assert printed query

v2/pkg/engine/resolve/fetch.go
```diff
@@ -308,6 +329,44 @@ func (fc *FetchConfiguration) Equals(other *FetchConfiguration) bool {
...
+	for i := range a.Variables {
+		if a.Variables[i].Name != b.Variables[i].Name {
+			return false
+		}
+		if !bytes.Equal(a.Variables[i].Value, b.Variables[i].Value) {
+			return false
+		}
+	}
+	aQuery, err := a.PrintedQuery()
+	if err != nil {
```
fiam, line 360: question: should we panic or be louder here? it seems there's potential to swallow an error

ysmolski: I dont think it will ever be error; only if `a.document` is nil. So the current error handling is alright.

v2/pkg/asttransform/merge_operations.go
```diff
@@ -0,0 +1,103 @@
...
+	VariableRename map[string]string
+}
+
+// MergeOperationDocuments builds a new document with a single query operation
+// (named operationName, anonymous if empty) containing every member's root
+// selection aliased and guarded by @include(if: $IncludeVariable). Variable
+// definitions are emitted per member in document order followed by that
+// member's synthetic include definition. Each member's root selection must be a
+// single field; fragment spreads in member documents are rejected.
+func MergeOperationDocuments(operationName string, members []OperationMergeMember) (*ast.Document, error) {
```
ysmolski, line 35: what happens to directives of merged documents? should we care about them?

v2/pkg/engine/datasource/httpclient/input_assembly.go
```diff
@@ -0,0 +1,66 @@
...
+// AssembleGraphQLRequestInput builds the upstream GraphQL request input
+// envelope in the exact order the planner assembles it today: body.variables,
+// body.query, header (only when non-empty and not JSON null), url, method.
+//
+// It uses the same SetInputBodyWithPath / SetInputHeader / SetInputURL /
+// SetInputMethod setters in the same sequence, so its output is byte-identical
+// to the inline assembly in createInputForQuery/ConfigureFetch. variables and
+// query are the raw bytes written at body.variables and body.query
+// respectively; header is the marshalled header JSON (nil or empty to omit);
+// url and method are the fetch endpoint values ("" to omit).
```
ysmolski: Redundant. Pls remove this part. Also error is always nil

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -828,6 +913,56 @@ func (p *Planner[T]) LeaveDocument(_, _ *ast.Document) {
...
+// name; only the synthetic moves. The result is computed once per document and
+// cached (reset in EnterDocument).
+//
+// Detection uses the downstream operation (p.visitor.Operation) because the
+// synthetic definition is added while scaffolding the upstream operation,
+// before client variables are imported into it — at write time the upstream
+// operation cannot yet reveal the collision. This makes the check conservative:
+// a client that declares $representations but never passes it to this subgraph
+// still triggers the rename, which is harmless because the renamed synthetic is
+// internal to the subgraph request.
```
ysmolski: I do not think that this is a readable English. @Aenimus what do you think about these comments? I am utterly confused to have to parse through such LLM stuff.

v2/pkg/engine/postprocess/postprocess.go
```diff
@@ -171,6 +208,15 @@ func NewProcessor(options ...ProcessorOption) *Processor {
 			addMissingNestedDependencies: &addMissingNestedDependencies{
 				disable: opts.disableAddMissingNestedDependencies,
 			},
+			createMultiFetch: &createMultiFetch{
+				disable: !enableMultiFetch,
+			},
+			// renderSubgraphInputs runs unconditionally (no disable flag): it must
+			// (only its op-name suffix rewrite follows fetchIDAppender's flag)
+			// render/clear artifacts even under DisableResolveInputTemplates.
```
ysmolski: Man, I dunno…

v2/pkg/engine/postprocess/create_multi_fetch.go
```diff
@@ -0,0 +1,433 @@
...
+			continue
+		}
+		isToken = true
+	}
+	return indices
+}
+
+// unionDependencies returns the members' DependsOnFetchIDs minus member IDs;
+// duplicates are tolerated, mirroring deduplicateSingleFetches.
+func unionDependencies(members []*resolve.SingleFetch, ids []int) []int {
```
ysmolski, line 274: Imagine a query that touches two different root fields, where both need extra fields from the same third subgraph:

```
query {
 a { ... entityFieldsFromC }   # root from subgraph A
 b { ... entityFieldsFromC }   # root from subgraph B
}
```

The planner produces four fetches:

* fetch 1 → subgraph A, resolves a (wave 1)
* fetch 2 → subgraph B, resolves b (wave 1)
* fetch 3 → subgraph C, entities under a, depends on fetch 1
* fetch 4 → subgraph C, entities under b, depends on fetch 2

Fetches 3 and 4 run in the same parallel wave and go to the same subgraph C. They get merged into one MultiEntityFetch, and unionDependencies gives the merged fetch the combined dependency list: DependsOnFetchIDs = \[1, 2\]. The individual "entry 3 only needs fetch 1, entry 4 only needs fetch 2" information is discarded.

The failure. Now suppose fetch 1 fails at the transport level (connection refused, timeout). The loader records fetch ID 1 in erroredFetchIDs. When the merged fetch is about to run, shouldSkipErroredDependencyLocked checks its dependencies \[1, 2\] against the errored set, finds 1, and skips the entire merged fetch; no request to subgraph C is ever sent. It also records the merged fetch's own ID as errored, so anything downstream of either member gets skipped too.

Are we okay, that such merging of fetches couples the failures of unrelated fetches, like 3 and 4?

ysmolski (commented):
2nd batch

execution/engine/execution_engine_multi_fetch_test.go
```diff
@@ -0,0 +1,321 @@
...
+// exercise the whole execution engine end-to-end (planning + postprocessing +
+// runtime), unlike the plan-level goldens in the v2 datasource and postprocess
+// packages.
+//
+// Configuration.EnableMultiFetch() is the single switch: it sets
+// plan.Configuration.EnableMultiFetch = true, and NewExecutionEngine derives the
+// postprocess.EnableMultiFetch() processor option from that same flag. One
+// switch drives BOTH required flags (planner artifact recording + postprocess
+// merge stage), so they can never drift apart. With the switch off there is zero
+// behavior change.
```
ysmolski: All this block is redundant. Maybe just leave:

```
// Full federation integration tests for the MultiFetch optimization.
```

v2/pkg/engine/postprocess/render_subgraph_inputs.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	// fetchIDAppender compensation: fetch_id_appender.go mutates
+	// only printed strings and no-ops on the empty deferred Input, so we apply
+	// the identical whole-input first-occurrence replace after assembly. This
+	// preserves today's semantics exactly, including the pre-existing quirk that
+	// the replace could hit a URL/header substring.
+	if !r.disableRewriteOpNames && fetch.OperationName != "" {
+		expandedName := fmt.Sprintf("%s__%d", fetch.OperationName, fetch.FetchID)
+		input = []byte(strings.Replace(string(input), fetch.OperationName, expandedName, 1))
```
ysmolski: maybe `bytes.Replace`?

v2/pkg/engine/postprocess/create_multi_fetch.go
```diff
@@ -0,0 +1,433 @@
...
+// createParallelNodes, run just before this stage) and merges same-datasource
+// entity fetches inside every real FetchTreeNodeKindParallel group. There is no
+// scratch tree and no DeferID partitioning: defer groups are extracted into
+// their own trees and organized separately before this stage runs, so the
+// parallel groups we see already encode the correct execution waves.
+func (c *createMultiFetch) walk(root, node *resolve.FetchTreeNode) {
+	if node == nil {
+		return
+	}
+	for i := 0; i < len(node.ChildNodes); i++ {
```
ysmolski: `for i := range node.ChildNode`

v2/pkg/engine/resolve/loader.go
```diff
@@ -1627,6 +1731,15 @@ func (b *batchEntityTools) reset() {
 	}
 }
 
+// clearDedupState resets the per-entry dedup scope without touching the
+// arena, whose buffers must survive until final input assembly.
+func (b *batchEntityTools) clearDedupState() {
+	b.keyGen.Reset()
+	for k := range b.batchHashToIndex {
+		delete(b.batchHashToIndex, k)
+	}
```
ysmolski: `clear(b.batchHashToIndex)`

v2/pkg/engine/postprocess/create_multi_fetch.go
```diff
@@ -0,0 +1,433 @@
...
+		info.QueryPlan = &resolve.QueryPlan{Query: pretty, DependsOnFields: dependsOnFields}
+	}
+	return info
+}
+
+// collectGroups returns every candidate set to merge across the ORGANIZED tree:
+// it descends into each FetchTreeNodeKindParallel group and, within it, buckets
+// candidates by DataSourceID (see groupCandidatesByDataSource). It is a
+// read-only query used by ProcessFetchTree's walk and by unit tests; it does not
+// mutate the tree.
```
ysmolski: It is not used by `ProcessFetchTree's walk`

v2/pkg/engine/resolve/fetch_multi.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+// MultiEntityFetch merges several same-subgraph entity fetches into one
+// request with aliased _entities fields guarded by @include variables.
+type MultiEntityFetch struct {
+	FetchDependencies
+
+	Input                MultiEntityInput
+	DataSource           DataSource
+	DataSourceIdentifier []byte
```
ysmolski: This field is not used anywhere

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
```diff
@@ -383,6 +439,31 @@ func (p *Planner[T]) ConfigureFetch() resolve.FetchConfiguration {
 		}
 	}
 
+	var subgraphOperation *resolve.SubgraphOperation
+	if p.recordUpstreamVariables && (requiresEntityFetch || requiresEntityBatchFetch) {
```
ysmolski: this condition was already computed at line 391

v2/pkg/astimport/astimport.go
```diff
@@ -104,13 +123,19 @@ func (i *Importer) ImportValue(fromValue ast.Value, from, to *ast.Document) (val
 		value.Ref = to.ImportEnumValue(from.EnumValueNameBytes(fromValue.Ref))
 
 	case ast.ValueKindVariable:
-		value.Ref = to.ImportVariableValue(from.VariableValueNameBytes(fromValue.Ref))
+		name := from.VariableValueNameString(fromValue.Ref)
+		if rename != nil {
```
ysmolski: this check is not needed

v2/pkg/asttransform/merge_operations.go
```diff
@@ -0,0 +1,103 @@
...
+// definition, matching the emit order the printed output depends on.
+func mergeVariableDefinitions(merged *ast.Document, opRef int, importer *astimport.Importer, member OperationMergeMember) {
+	doc := member.Document
+	for _, defRef := range doc.OperationDefinitions[0].VariableDefinitions.Refs {
+		origName := doc.VariableDefinitionNameString(defRef)
+		importedRef := importer.ImportVariableDefinitionWithVariableNameRename(defRef, doc, merged, member.VariableRename[origName])
+		merged.AddImportedVariableDefinitionToOperationDefinition(opRef, importedRef)
+	}
+	boolType := merged.AddNamedType([]byte("Boolean"))
+	nonNullBool := merged.AddNonNullType(boolType)
```
ysmolski: Do we have to add N `Boolean!` types here? Maybe we could add just one in the `MergeOperationDocuments`and pass reference heret for all N synthetic variables?

ysmolski (commented):
another batch

v2/pkg/engine/postprocess/create_multi_fetch.go
```diff
@@ -0,0 +1,433 @@
...
+	// the same fetch variable across members — the structured equivalent of the
+	// old envelope-token check. In practice the planner never emits $$K$$ tokens
+	// into the envelope (url/method are static strings and the header is static
+	// config marshalled to JSON), so for real plans this check finds no tokens
+	// and reduces to nothing; it is retained to preserve the exact merge-abort
+	// semantics for envelopes that do carry template tokens.
+	envelopeTokenSource := base.Envelope.Method + base.Envelope.URL + string(base.Envelope.Header)
+	for _, k := range envelopeTokenIndices(envelopeTokenSource) {
+		for i := 1; i < len(members); i++ {
+			if k >= len(members[i].Variables) || k >= len(members[0].Variables) ||
```
ysmolski: `k >= len(members[0].Variables)` could be moved outside the inner loop

v2/pkg/engine/resolve/loader.go
fiam: Great catch, we should try to find a linter for this!

v2/pkg/engine/postprocess/create_multi_fetch.go
fiam: changed in 7f6180d3 - thank you!

v2/pkg/engine/datasource/httpclient/input_assembly.go
fiam: updated in 8602c005, thanks for catching that!

v2/pkg/engine/resolve/fetch_multi.go
fiam: removed in 462c588a — thank you!

v2/pkg/engine/postprocess/create_multi_fetch.go
fiam: fixed in 33be7c0e, thank you!

execution/engine/execution_engine_multi_fetch_test.go
fiam: updated in 3df9ae4e, thank you!

v2/pkg/engine/postprocess/render_subgraph_inputs.go
fiam: changed in e1b36935, thanks for the suggestion!

fiam: I did a quick-and-dirty benchmark on this, and using `bytes.Replace` directly came in 2x faster

v2/pkg/astimport/astimport.go
fiam: Thanks, we should find a linter for this one too.

fiam: fixed in 728c9520, thank you!

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
fiam: updated in 3e2b263a, thank you!

v2/pkg/engine/postprocess/create_multi_fetch.go
fiam: You are right, the production walk calls groupCandidatesByDataSource directly, so collectGroups was only used by tests. I moved it out of the production file and made it a free helper in create_multi_fetch_test.go in a5f4e7ac. Thank you!

v2/pkg/astimport/astimport_test.go
fiam: Changed in c983daa0. The test now asserts the complete printed operation.

v2/pkg/engine/postprocess/postprocess.go
fiam: Changed in b220150f. I removed this duplicated constructor comment; the invariant is already documented on renderSubgraphInputs.

v2/pkg/engine/datasource/graphql_datasource/graphql_datasource.go
fiam: Changed in b220150f. The planner helper comments now state only their purpose and omit the implementation narrative.

v2/pkg/asttransform/merge_operations.go
fiam: @devsergiy could you PTAL at this one? Thanks!

v2/pkg/asttransform/merge_operations.go
fiam: Changed in 9e188e14. MergeOperationDocuments now creates one immutable Boolean! type node and passes its reference to mergeVariableDefinitions, so every synthetic include variable reuses the same AST type. The compact and pretty printed output remains unchanged.

fiam: @devsergiy PTAL at this one

v2/pkg/engine/postprocess/create_multi_fetch.go
fiam: deferring to @devsergiy for this

v2/pkg/astimport/astimport.go
```diff
@@ -201,6 +240,31 @@ func (i *Importer) ImportVariableDefinitionWithRename(ref int, from, to *ast.Doc
...
+		},
+	}
+
+	if from.VariableDefinitions[ref].DefaultValue.IsDefined {
+		variableDefinition.DefaultValue.Value = i.ImportValue(from.VariableDefinitions[ref].DefaultValue.Value, from, to)
+	}
+
+	to.VariableDefinitions = append(to.VariableDefinitions, variableDefinition)
+	return len(to.VariableDefinitions) - 1
+}
```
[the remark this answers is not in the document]
ysmolski: I have added TODO for later

execution/engine/execution_engine_multi_fetch_test.go
```diff
@@ -0,0 +1,311 @@
...
+		if rec != nil {
+			rec.record(string(body))
+		}
+		resp, ok := responses[string(body)]
+		if !ok {
+			t.Logf("[%s] unexpected body: %s", host, string(body))
+			return &http.Response{StatusCode: 400, Body: io.NopCloser(bytes.NewBufferString("unexpected body"))}
+		}
+		return &http.Response{StatusCode: resp.statusCode, Body: io.NopCloser(bytes.NewBufferString(resp.body))}
+	})
```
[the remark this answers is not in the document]
ysmolski: fixed

v2/pkg/asttransform/merge_operations.go
ysmolski: I am leaving TODO in place.

v2/pkg/asttransform/merge_operations.go
ysmolski: This fix looks good to me

v2/pkg/engine/postprocess/render_subgraph_inputs.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// Reuse the print-once cached minified operation (seeded by the planner);
+	// PrintedQuery falls back to printing Document for artifacts that were not
+	// printed eagerly (e.g. hand-built test fetches).
+	query, err := op.PrintedQuery()
+	if err != nil {
+		return
+	}
+
+	input := httpclient.AssembleGraphQLRequestInput(variables, query, op.Envelope.Header, op.Envelope.URL, op.Envelope.Method)
```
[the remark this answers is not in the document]
ysmolski: No need to verify these errors. They cannot happen

ysmolski (approved):

v2/pkg/asttransform/merge_operations.go
devsergiy: we haven't implemented support of importing execution directives for the fields
technically helpers are there, but they should be wired
I would consider it a small follow-up task, because they are rarely used

v2/pkg/engine/postprocess/create_multi_fetch.go
devsergiy: hmm, actually no, we should distinguish such case

There is a functionality of skipping one of the nested fetches with skip or include directive, do not remember which exact directive I have used, it is currently used to skip fetches when there was no entities found, e.g. empty items

but merged fetch still carries information from the original fetch, so we could detect that part of the multifetch is executable and skip other parts

v2/pkg/asttransform/merge_operations.go
devsergiy: In this case, it doesn't matter so much, as we create AST on the fly
though the parser handles it exactly like it was before the comment, because it is just a token

It is not a reference to a boolean type; it is just the name of the named type
https://spec.graphql.org/October2021/#sec-Type-References

v2/pkg/engine/postprocess/create_multi_fetch.go
ysmolski: @devsergiy I wonder if we should fix this in this PR or a follow up. The fix is not trivial and touches many points.

ysmolski: Postponed in a separate ticket

PR #1595, opened 2026-07-17
fix(resolve): resolve per-subscriber subscription updates concurrently

mwisner:


v2/pkg/engine/resolve/resolve.go
```diff
@@ -918,6 +918,17 @@ type subscriptionState struct {
...
+	// consecutive events for the same SubscriptionIdentifier resolve and write in
+	// order. Different subscriptions have different resolveMu instances, so updates
+	// to distinct subscribers on a trigger still run concurrently.
+	//
+	// This preserves per-subscription event ordering even when the pubsub layer
+	// abandons an event's in-flight goroutines on handler_timeout and starts the
+	// next event: the next event's update for this subscription queues behind the
+	// previous one here rather than racing it (writeMu only orders the final write,
+	// not by event).
+	resolveMu sync.Mutex
```
endigma, line 931: `UpdateSubscription` now works by:

1. Lock `s.mu`, check `done`/`ctx`, release `s.mu`
2. Lookup up the trigger (`r.mu`, run the filter (`trig.mu`)
3. Lock `sub.resolveMu` in `executeSubscriptionUpdate`, resolve, write

Between 1 and 3 the `UpdateSubscription` goroutine holds no lock, so event order depends on which goroutine reached `resolveMu` first.

So some bad things can happen:
- Data frame written after terminal frame
- Same-subscription events out of order
- Filter errors write under `writeMu` only, so errors for event B could land before event A
- A goroutine past step 1 could survive `done` and deliver a stale event to a new trigger
- Probably more

I've have some other ideas about how to reduce locking overall in this area but haven't had the time to develop them further as well, ideally I'd like to widen the pipe from the new `/v2/pkg/engine/datasource/graphql_datasource/subscriptionclient` all the way into the router and vice versa to remove as much locking and synchronisation as possible so different subscriptions are mostly decoupled.

---

Claude's suggestion for fixing it (take with a grain of salt, mutexes in here are tricky business!):

> Suggestion: record order at step 1, while s.mu is held and order is still knowable, instead of deriving it from lock acquisition:
> 
> 1. Stamp each admitted event with a per-subscription sequence number under s.mu.
> 2. Resolve concurrently, in any order.
> 3. At write time, publish strictly by sequence number.
> 4. Have complete()/error() set a terminal flag checked at write time next to removed; late resolves drop their output, and the terminal path stops waiting on in-flight resolves.

Sounds vaguely correct, but I haven't thought about it much yet.

endigma (changes requested):

PR #1605, opened 2026-07-20
feat(resolve): resolve-once-broadcast for identical subscribers on a trigger

mwisner:
## Problem

On a subscription trigger update, `handleTriggerUpdate` spawns one goroutine per subscriber, and each independently runs the full resolve — `InitSubscription` + the subgraph fetch + serialize. For a high-fanout EDFS subscription where N subscribers select the **same fields** and produce **byte-identical** output (same plan, same variables, same subgraph headers, no per-viewer auth/tracing), this is N× redundant work: N subgraph fetches and N serializes for one event. Subgraph load and router CPU per event scale linearly with subscriber count.

Subgraph single-flight doesn't help here: it only collapses concurrently in-flight `Query` fetches, not the per-subscriber subscription resolve path.

## Change

Bucket the (already filtered) subscribers on a trigger by a key that captures everything that can change the serialized output, **resolve once per group**, and fan the identical bytes out to each member concurrently. Subscribers whose output can vary per viewer resolve individually, exactly as before.

Grouping key (`subGroupKey`):
- **plan** — `*GraphQLSubscription` pointer identity (the plan cache hands identical operations the same plan)
- **variables** — `ctx.VariablesHash`
- **headers** — `ctx.SubgraphHeadersBuilder.HashAll()` (all subgraph request headers)

A subscriber is excluded from grouping (resolves solo, fail-closed) when output could be viewer-specific:
- the plan has authorization coordinates (`@authenticated`/`@requiresScopes` can gate fields on the viewer's scopes), or
- tracing is enabled for the request (a sampled request renders a per-request trace extension).

## What's preserved

- **Subscription filters** — `filterSubscriptions` still runs per subscriber *before* grouping, so `@openfed__subscriptionFilter` routing is unchanged; only subscribers that already passed the filter for this event are grouped.
- **Event ordering** — unchanged; the per-trigger updater mutex still serializes event A vs B.
- **Per-subscriber writes** — each member keeps its own `writeMu`/writer and re-checks `removed` before writing; only the immutable resolved bytes are shared.
- **Error semantics** — a shared resolve error is delivered to every member of that group (they'd all have hit the same error), and only that group.

## Benefits

- Collapses N resolves/fetches/serializes to **one per distinct output**. For same-event high-fanout subscriptions, subgraph fetch count and serialization CPU stop scaling with subscriber count — the last-subscriber path no longer pays for N fetches.

## Tradeoffs

- Grouping only engages when output is provably identical. Per-viewer authorization (scopes) or tracing opt out, so those subscriptions behave exactly as today (no regression, no benefit).
- Correctness lives in the grouping key: it must include *every* determinant of the output. `HashAll()` is intentionally over-inclusive (hashes all known subgraph headers) so it's safe rather than minimal; the auth/tracing gate is conservative (any authorization coordinates → don't group).
- A shared resolve error now fans out to a whole group rather than a single subscription — this is intended, but it does couple the group's error delivery.
- Small per-event bucketing cost (a map keyed by `subGroupKey`), negligible against the fetches/serializes saved.

Happy to gate this behind a resolver option if you'd prefer it opt-in rather than automatic.

## Testing

Full `pkg/engine/resolve` suite passes, including with `-race`.

### DIFF
--- a/v2/pkg/engine/resolve/resolve.go
+++ b/v2/pkg/engine/resolve/resolve.go
@@ -979,11 +979,11 @@ func (s *subscriptionState) sendHeartbeat() error {
 	return s.writer.Heartbeat()
 }
 
-func (r *Resolver) executeSubscriptionUpdate(resolveCtx *Context, sub *subscriptionState, sharedInput []byte) {
-	if r.options.Debug {
-		fmt.Printf("resolver:trigger:subscription:update:%d\n", sub.id.SubscriptionID)
-	}
-
+// resolveSubscriptionUpdateToBytes runs a single subscription resolve (init + pre-fetch auth +
+// load + serialize) for the given representative context/subscription and returns the serialized
+// response body. It does not touch any subscriber's writer, so the result can be fanned out to a
+// group of subscribers whose resolved output is identical (see executeSubscriptionUpdateGroup).
+func (r *Resolver) resolveSubscriptionUpdateToBytes(resolveCtx *Context, sub *subscriptionState, sharedInput []byte) ([]byte, error) {
 	ctx, cancel := context.WithTimeout(resolveCtx.ctx, r.maxSubscriptionFetchTimeout)
 	defer cancel()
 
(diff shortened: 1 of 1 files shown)


dkorittki:
Discussed with @mwisner that I continue this work on a different pull request https://github.com/wundergraph/graphql-go-tools/pull/1611 --> closing this one

PR #1612, opened 2026-07-23
feat: schedule fetch trees optimally

ysmolski:
scheduleFetches is a dependency-aware scheduler that emits nested Sequence/Parallel trees, collapsing independent chains onto their own branches instead of synchronizing them at wave barriers.

This scheduler tries to break the top level DAG into weakly connected components. For each component it picks between two strategies of scheduling:
  - the waves tree, that barriers all ready roots per step,
  - the inlined tree, that pulls each root's exclusively-reachable descendants into that root's branch.

This scheduler implements a strategy that is guaranteed to be strictly the best condering all possible makespans of fetches.

According to production data, recursive wave scheduling provides the most value. It creates wave tree that is deep and where fetches already ordered in the order of dependencies. Inlining gives about 3-5% improvements on top of that. 

## Summary by CodeRabbit

* **New Features**
  * Added optional dependency-aware scheduling for fetch execution.
  * Added configuration support for enabling scheduled fetch processing.
  * Scheduling supports parallel execution while preserving dependency order and subscription behavior.
  * Retains existing processing behavior through fallback when scheduling is unavailable.

* **Bug Fixes**
  * Improved duplicate dependency handling and deterministic fetch ordering.
  * Added validation for invalid, cyclic, or unsupported schedules.

* **Tests**
  * Expanded coverage for dependency graphs, cycles, large plans, subscriptions, multi-fetch scenarios, and scheduler fallback.


v2/pkg/engine/postprocess/util_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+// deps sets the fetch IDs the fetch depends on.
+func deps(ids ...int) sfOpt {
+	return func(node *resolve.FetchTreeNode) {
+		node.Item.Fetch.(*resolve.SingleFetch).FetchDependencies.DependsOnFetchIDs = ids
+	}
+}
+
+// at nests the fetch at responsePath, e.g. "a.b" becomes ObjectPath("a"), ObjectPath("b").
+func at(responsePath string) sfOpt {
```
devsergiy:
```suggestion
func respPath(responsePath string) singleFetchOption {
```

v2/pkg/engine/postprocess/util_test.go
```diff
@@ -0,0 +1,169 @@
...
+			path[i] = resolve.ObjectPath(segment)
+		}
+		node.Item.FetchPath = path
+		node.Item.ResponsePath = responsePath
+		node.Item.ResponsePathElements = segments
+	}
+}
+
+// provides sets the merge path the fetch provides in the response.
+func provides(mergePath ...string) sfOpt {
```
devsergiy:
```suggestion
func mergePath(mergePath ...string) singleFetchOption {
```

v2/pkg/engine/postprocess/util_test.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+// Various test helpers to reduce clutter in tests.
+
+func nodes(items ...*resolve.FetchTreeNode) []*resolve.FetchTreeNode {
+	return items
+}
+
+// sfOpt configures optional fields of a single-fetch node built by sf.
+type sfOpt func(node *resolve.FetchTreeNode)
```
devsergiy:
```suggestion
type singleFetchOption func(node *resolve.FetchTreeNode)
```

v2/pkg/engine/postprocess/util_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+// deps sets the fetch IDs the fetch depends on.
+func deps(ids ...int) sfOpt {
```
devsergiy:
```suggestion
func dependsOn(ids ...int) singleFetchOption {
```

v2/pkg/engine/postprocess/util_test.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// provides sets the merge path the fetch provides in the response.
+func provides(mergePath ...string) sfOpt {
+	return func(node *resolve.FetchTreeNode) {
+		node.Item.Fetch.(*resolve.SingleFetch).PostProcessing.MergePath = mergePath
+	}
+}
+
+func sf(id int, opts ...sfOpt) *resolve.FetchTreeNode {
```
devsergiy:
```suggestion
func singleFetch(id int, opts ...singleFetchOption) *resolve.FetchTreeNode {
```

v2/pkg/engine/postprocess/util_test.go
```diff
@@ -0,0 +1,169 @@
...
+	node := resolve.Single(&resolve.SingleFetch{FetchDependencies: resolve.FetchDependencies{FetchID: id}})
+	for _, opt := range opts {
+		opt(node)
+	}
+	return node
+}
+
+func ef(id int, deps ...int) *resolve.FetchTreeNode {
```
devsergiy:
```suggestion
func entityFetch(id int, deps ...int) *resolve.FetchTreeNode {
```

v2/pkg/engine/postprocess/util_test.go
```diff
@@ -0,0 +1,169 @@
...
+	return resolve.Single(&resolve.EntityFetch{FetchDependencies: resolve.FetchDependencies{FetchID: id, DependsOnFetchIDs: deps}})
+}
+
+func bf(id int, deps ...int) *resolve.FetchTreeNode {
```
devsergiy:
```suggestion
func batchEntityFetch(id int, deps ...int) *resolve.FetchTreeNode {
```

v2/pkg/engine/postprocess/util_test.go
```diff
@@ -0,0 +1,169 @@
...
+	return resolve.Single(&resolve.BatchEntityFetch{FetchDependencies: resolve.FetchDependencies{FetchID: id, DependsOnFetchIDs: deps}})
+}
+
+func seq(children ...*resolve.FetchTreeNode) *resolve.FetchTreeNode {
```
devsergiy, line 58:
```suggestion
func sequenceNode(children ...*resolve.FetchTreeNode) *resolve.FetchTreeNode {
```

v2/pkg/engine/postprocess/util_test.go
```diff
@@ -0,0 +1,169 @@
...
+	return resolve.Sequence(children...)
+}
+
+func par(children ...*resolve.FetchTreeNode) *resolve.FetchTreeNode {
```
devsergiy, line 62:
```suggestion
func parallelNode(children ...*resolve.FetchTreeNode) *resolve.FetchTreeNode {
```

v2/pkg/engine/postprocess/util_test.go
```diff
@@ -0,0 +1,169 @@
...
+func fetchesByID(input []*resolve.FetchTreeNode) map[int]*resolve.FetchTreeNode {
+	byID := make(map[int]*resolve.FetchTreeNode, len(input))
+	for _, node := range input {
+		byID[node.Item.Fetch.Dependencies().FetchID] = node
+	}
+	return byID
+}
+
+// materialize returns shape with every leaf replaced by the input node carrying the same fetch ID.
+func materialize(t *testing.T, shape *resolve.FetchTreeNode, byID map[int]*resolve.FetchTreeNode) *resolve.FetchTreeNode {
```
devsergiy: It was rather unclear what this function is doing or actually why it is doing it

The root cause is that "want" assertions actually do not have fetch dependencies added

Maybe we could just add them and get rid of this function?

v2/pkg/engine/postprocess/util_test.go
```diff
@@ -0,0 +1,169 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/wundergraph/graphql-go-tools/v2/pkg/engine/resolve"
+)
+
+// Various test helpers to reduce clutter in tests.
+
+func nodes(items ...*resolve.FetchTreeNode) []*resolve.FetchTreeNode {
```
devsergiy, line 15: overall for these helpers, they are short, but tests are very hard to read in this form

please consider more wordy names, smth like I have put in the suggested changes

v2/pkg/engine/postprocess/postprocess.go
```diff
@@ -53,6 +54,10 @@ func (p *FetchTreeProcessors) processFlatFetchTree(response *resolve.GraphQLResp
 // organizeFetchTree organizes the fetch tree by ordering sequence nodes by dependencies and creating parallel nodes.
 // after this step fetches have tree structure of serial and parallel nodes.
 func (p *FetchTreeProcessors) organizeFetchTree(fetches *resolve.FetchTreeNode) {
```
devsergiy, line 62: schedule fetches, in case of error fallbacks to waves, should it be explicit here?

smth like this?

```go
func (p *FetchTreeProcessors) organizeFetchTree(fetches *resolve.FetchTreeNode) {
	if !p.scheduleFetches.disable {
		if err := p.scheduleFetches.ProcessFetchTree(fetches); err == nil {
			return
		}
	}
	p.organizeFetchTreeInWaves(fetches)
}

func (p *FetchTreeProcessors) organizeFetchTreeInWaves(fetches *resolve.FetchTreeNode) {
	p.orderSequenceByDependencies.ProcessFetchTree(fetches)
	p.createParallelNodes.ProcessFetchTree(fetches)
}
```

v2/pkg/engine/postprocess/schedule_fetches_test.go
```diff
@@ -0,0 +1,1278 @@
...
+				par(sf(1), sf(3)),
+				par(sf(2), sf(4)),
+				sf(5),
+				par(sf(6), sf(8)),
+				par(sf(7), sf(9)),
+				sf(10),
+			),
+		},
+		{
+			name:  "requires chain",
```
devsergiy: this and next 3-4 tests doesn't make a lot of sense, their descriptions doesn't matter in context of what we are testing and they are looks covered by other cases

v2/pkg/engine/postprocess/schedule_fetches_test.go
```diff
@@ -0,0 +1,1278 @@
...
+				sf(22, deps(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21)),
+				sf(23, deps(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22)),
+			),
+			want: seq(sf(0), sf(1), sf(2), sf(3), sf(4), sf(5), sf(6), sf(7), sf(8), sf(9), sf(10), sf(11), sf(12), sf(13), sf(14), sf(15), sf(16), sf(17), sf(18), sf(19), sf(20), sf(21), sf(22), sf(23)),
+			// The legacy wave pipeline:
+			// seq(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23)
+		},
+		{
+			// 3becc2e0f56d3d513c374e8035d1533cc6795740193a45c0c07def3f1f11eb0b
+			name: "deeply nested operation with wide dependencies",
```
devsergiy: Could we add a benchmark for this complex use case? Is there any significant overhead with a scheduler compared to Waves?

v2/pkg/engine/postprocess/schedule_fetches.go
```diff
@@ -0,0 +1,516 @@
...
+func buildScheduleTree(roots []*resolve.FetchTreeNode, dag *fetchDAG) (*resolve.FetchTreeNode, error) {
+	ids := make([]int, 0, len(roots))
+	for _, root := range roots {
+		if root == nil || root.Item == nil || root.Item.Fetch == nil {
+			continue
+		}
+		ids = append(ids, root.Item.Fetch.Dependencies().FetchID)
+	}
+	// Pick the best strategy on the top level for weakly connected trees.
+	components := weaklyConnectedComponents(sortedUnique(ids), dag)
```
devsergiy: q: why we do dedup in a `sortedUnique(ids)`? shouldn't we have here only unique ids?

devsergiy (changes requested):

v2/pkg/engine/postprocess/schedule_fetches_test.go
```diff
@@ -0,0 +1,1278 @@
...
+	ids := make([]int, 0, len(d.nodes))
+	for id := range d.nodes {
+		ids = append(ids, id)
+	}
+	slices.Sort(ids)
+	return ids
+}
+
+// hasCycle runs Kahn's algorithm over the full DAG.
+func (d *fetchDAG) hasCycle() bool {
```
devsergiy: `sortedIDs` and `hasCycle` are used only in this test file; should it be just a function receiving the `fetchDAG`?

v2/pkg/engine/postprocess/util_test.go
ysmolski: okay, I have pushed on the side of shortness too much. Renaming now...

v2/pkg/engine/postprocess/util_test.go
ysmolski: done, except for one test where it was aliases into deps for shortness

v2/pkg/engine/postprocess/util_test.go
ysmolski: done

v2/pkg/engine/postprocess/util_test.go
ysmolski: Here and par, seq -- I am not sure. Test cases would start to look verbose. And we have some tests where high density of code makes them more readably.

v2/pkg/engine/postprocess/util_test.go
ysmolski: The function was added to remove dependencies from the expected part of the test as not important. It is to reduce the verbiage of tests. I have fixed the description of this function

v2/pkg/engine/postprocess/postprocess.go
ysmolski: Reordered and made a comment that includes a future a merging step

v2/pkg/engine/postprocess/schedule_fetches.go
ysmolski: huh. Not sure how I missed it, but dedupe is totally redundant

v2/pkg/engine/postprocess/schedule_fetches_test.go
ysmolski: Thanks for spotting it. Removed redundant

v2/pkg/engine/postprocess/util_test.go
ysmolski: removed

v2/pkg/engine/postprocess/util_test.go
ysmolski: removed

v2/pkg/engine/postprocess/schedule_fetches_test.go
ysmolski: Added a comparison. The new algo is faster but consumes more memory:

```
$ go test -bench=BenchmarkOrganizeFetchTree -benchmem -run='^$' ./... -benchtime=10s
goos: darwin
goarch: arm64
pkg: github.com/wundergraph/graphql-go-tools/v2/pkg/engine/postprocess
cpu: Apple M4 Max
BenchmarkOrganizeFetchTree/schedule-14         	   36297	    330734 ns/op	  406910 B/op	    3691 allocs/op
BenchmarkOrganizeFetchTree/waves-14            	   26619	    455530 ns/op	  244089 B/op	    8321 allocs/op
```

v2/pkg/engine/postprocess/schedule_fetches_test.go
ysmolski: made them functions

v2/pkg/engine/postprocess/util_test.go
ysmolski: fixed

v2/pkg/engine/postprocess/util_test.go
ysmolski: fixed

v2/pkg/engine/postprocess/schedule_fetches_test.go
```diff
@@ -965,21 +838,24 @@ func TestScheduleFetches_SubscriptionRootStaysSequence(t *testing.T) {
...
+	require.Nil(t, err)
 	require.Equal(t, resolve.FetchTreeNodeKindSequence, parallel.Kind)
 	require.Equal(t, trigger, parallel.Trigger)
 	require.Equal(t, nodes(par(sf(0), sf(1))), parallel.ChildNodes)
 
 	// Without a Trigger the root may collapse into the scheduled tree.
 	sync := seq(sf(0))
-	(&scheduleFetches{}).ProcessFetchTree(sync)
+	err = (&scheduleFetches{}).ProcessFetchTree(sync)
+	require.Nil(t, err)
```
devsergiy: nit: you could use require.NoError

devsergiy (approved):

PR #1618, opened 2026-08-04
feat: entitycache cache interface

SkArchon:
This PR creates the interface of the cache for the adapters used by entity caching. The router will create implementations based on this interface.

Used here: https://github.com/wundergraph/cosmo/pull/3137

## Summary by CodeRabbit

* **New Features**
  * Added a standardized interface for batch retrieving and storing cached values.
  * Added support for per-item expiration times, including validation for missing or invalid lifetimes.
  * Cache reads now return found entries by key, omitting missing values while preserving empty stored values.
  * Defined consistent handling for duplicate keys, last-write-wins updates, partial storage failures, and reporting successfully stored entries.

## Checklist

## Open Source AI Manifesto

### DIFF
--- a/v2/pkg/entitycaching/cache.go
+++ b/v2/pkg/entitycaching/cache.go
@@ -21,14 +21,20 @@ type Result struct {
 	Found bool
 }
 
+// Cache is a batch oriented key/value cache.
 type Cache interface {
 	// GetMany looks up every key and returns one Result per key, in the same
 	// order as keys, so results[i] belongs to keys[i]. Duplicate keys are
 	// looked up independently.
+	// A miss is reported as a Result with Found false, never as an error. An
+	// error means the lookup itself failed, and the returned slice is then nil:
+	// there is no partial read to salvage, so callers treat the whole batch as
+	// a miss and recompute. Only a nil error guarantees a slice of len(keys).
 	GetMany(ctx context.Context, keys []string) ([]Result, error)
 
 	// SetMany stores every item. When items contains the same key twice, the
 	// last one wins. A nil error is not a guarantee that a subsequent GetMany
-	// will hit.
+	// will hit. An error means an unspecified subset of the items may already
+	// have been stored.
 	SetMany(ctx context.Context, items []Item) error
 }
--- a/v2/pkg/entitycaching/cache.go
+++ b/v2/pkg/entitycaching/cache.go
@@ -2,38 +2,57 @@ package entitycaching
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"time"
 )
 
+// Item is a single cache entry, both on the way in through SetMany and on the
+// way back out of GetMany.
 type Item struct {
-	Key   string
+	// Key is the key the entry is stored under, exactly as GetMany was asked
+	// for it, so an Item carries enough to be acted on once it has been taken
+	// out of the map it came in.
+	Key string
+	// Value is the cached bytes. An entry with an empty Value is still an
+	// entry: GetMany omitting the key is the only thing that means a miss.
 	Value []byte
-	TTL   time.Duration
-}
-
-// Result is a single lookup outcome returned by Cache.GetMany.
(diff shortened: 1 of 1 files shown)


v2/pkg/entitycaching/cache.go
```diff
@@ -0,0 +1,32 @@
+package entitycaching
+
+import (
+	"context"
+	"time"
+)
+
+type Item struct {
+	Key       string
+	Value     []byte
```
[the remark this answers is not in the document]
SkArchon: check now

v2/pkg/entitycaching/cache.go
```diff
@@ -0,0 +1,39 @@
...
+	// error means the lookup itself failed, and the returned slice is then nil:
+	// there is no partial read to salvage, so callers treat the whole batch as
+	// a miss and recompute. Only a nil error guarantees a slice of len(keys).
+	GetMany(ctx context.Context, keys []string) ([]Result, error)
+
+	// SetMany stores every item. When items contains the same key twice, the
+	// last one wins. A nil error is not a guarantee that a subsequent GetMany
+	// will hit. An error means an unspecified subset of the items may already
+	// have been stored.
+	SetMany(ctx context.Context, items []Item) error
```
dkorittki, line 41: Is `SetMany` transactional? If item 3 out of 5 has a problem while storing, are the other 4 discarded as well? would be nice to mention this on the godoc. Spoiler: I would like it to be transactional, everything else is cumbersome to deal with on the call site.

> A nil error is not a guarantee that a subsequent GetMany will hit

Why is that?

> An error means an unspecified subset of the items may already have been stored

Nothing else? No networking problems etc.?

Is an error returned if one of the entries has a bad or missing TTL?

v2/pkg/entitycaching/cache.go
```diff
@@ -0,0 +1,39 @@
...
+// Cache is a batch oriented key/value cache.
+type Cache interface {
+	// GetMany looks up every key and returns one Result per key, in the same
+	// order as keys, so results[i] belongs to keys[i]. Duplicate keys are
+	// looked up independently.
+	// A miss is reported as a Result with Found false, never as an error. An
+	// error means the lookup itself failed, and the returned slice is then nil:
+	// there is no partial read to salvage, so callers treat the whole batch as
+	// a miss and recompute. Only a nil error guarantees a slice of len(keys).
+	GetMany(ctx context.Context, keys []string) ([]Result, error)
```
dkorittki: Is ordering important?

If not I would prefer `map[string][]byte` as the return type. You can use `m["my-key"]nil` for empty values ( instead of `Result.Found = false`) or return `nil, err` in case of errors.

In general I would like to avoid constant type convertions between slices and maps in the engine or router.

dkorittki: What kind of role does the TTL play here? You have to specify one when using `SetMany` but it's not returned on `GetMany` for the same item. Are outdated items simply not part of the result set here?

In general I think it's best to use `Item` for both `SetMany` and `GetMany`. Lets the interface feel consistent. `Item` in, `Item` out.

SkArchon: I would say ordering is important for now, but it depends on how the engine would be using the interface which we dont have visibility on for now.

v2/pkg/entitycaching/cache.go
SkArchon: `A nil error is not a guarantee that a subsequent GetMany will hit`: will remove
`An error means an unspecified subset of the items may already have been stored`: will change to `Depending on the error, an unspecified subset of the items may already have been stored`

>Is an error returned if one of the entries has a bad or missing TTL?
If any entries have invalid ttls we will not even attempt to save.

dkorittki (approved):
I think we need to work on the interface eventually but it really depends how it's going to be used. Since this change is not useable from any path right now I think we can integrate it and adjust when its time to.

PR #1622, opened 2026-08-06
chore: centralize Go tooling

fiam:
## Summary

- add a dedicated `tools` workspace module as the single source of truth for Go developer tools
- pin Buf, golangci-lint, gotestsum, gqlgen, go-enum, mockgen, stringer, and the protobuf generators
- invoke tooling through `go tool` from Makefiles, `go:generate` directives, and Buf plugin configuration
- remove `GOPATH`/`go install` generator setup and tool-only dependencies from product modules
- make both CI workflows read the golangci-lint version from the tools module
- refresh all generated code, configure the accounts fixture for Federation 2, and align standalone modules with v2.18.0
- delete the download wrapper and keep the workspace graph compatible with the split `google.golang.org/genproto` modules

## Validation

- `go generate ./...` in `v2`, `execution`, `examples/federation`, and `examples/engine` (three idempotent full passes)
- `make -C v2/pkg/grpctest generate-proto`
- `make -C v2 lint`
- `make -C v2 test-quick`
- `make -C execution test-quick`
- `go test ./...` in both example modules
- `GOWORK=off go mod verify` from `tools`
- pinned-version resolution for all ten tools
- `git diff --check`

All generators converge without further changes, both gated modules report zero lint issues, all tests pass, and protobuf generation is byte-for-byte stable.

## Summary by CodeRabbit

- **Improvements**
  - Standardized code generation, linting, formatting, testing, and protocol generation through the Go toolchain.
  - Tool versions are now managed centrally, improving consistency across local development and automated checks.
  - Updated federation examples and testing configuration to support the latest dependency versions and federation features.
  - Simplified setup by removing obsolete tool installation steps and refreshing supporting dependencies.
  - CI checks now respond to relevant workspace and tooling changes and select compatible lint versions automatically.


execution/Makefile
```diff
@@ -22,14 +22,14 @@ update-fixtures:
...
+	../scripts/golangci-lint.sh fmt
 
 lint:
-	golangci-lint run
+	../scripts/golangci-lint.sh run
 
 .PHONY: lint-fix
 lint-fix:
-	golangci-lint run --fix
+	../scripts/golangci-lint.sh run --fix
```
Noroth: I'm not 100% convinced with this approach. Sure you can always download and delete golangci-lint, but at the same time, why are we not using tools?

We can create a global go.tools.mod file or something like that and move everything there. And then use `go tool xxxx` as command in the makefile.

Noroth (changes requested):

execution/Makefile
fiam: Thanks for taking a look! I'm happy to iterate on it until we find consensus.

I'm happy to use go to manage golang-ci-lint and the other tools we already have, as long as we use a separate module for all of them. Does that sound good?

Noroth: Sounds good to me

tools/go.mod
```diff
@@ -0,0 +1,337 @@
...
+	pluginrpc.com/pluginrpc v0.5.0 // indirect
+)
+
+tool (
+	connectrpc.com/connect/cmd/protoc-gen-connect-go
+	github.com/99designs/gqlgen
+	github.com/abice/go-enum
+	github.com/bufbuild/buf/cmd/buf
+	github.com/golang/mock/mockgen
+	github.com/golangci/golangci-lint/v2/cmd/golangci-lint
```
endigma, line 332: installing from source is not recommended by [golangci-lint docs](<https://golangci-lint.run/docs/welcome/install/local/#install-from-sources>)

I would personally recommend we use [mise](https://mise.en.dev) for all of this, including [golangci-lint](https://golangci-lint.run/docs/welcome/install/local/#mise), as a more complete solution for tool management and eventually task management (replacing Makefiles and scripts via mise-tasks)

endigma (changes requested):

tools/go.mod
fiam: Thanks! Just to clarify, CI is not building `golangci-lint` from source. It reads the version from `tools/go.mod` and passes it to the official action, which downloads the prebuilt binary.
  `go tool` is only used locally.

  I believe this is still a step towards better reproducibility. Before this change, `make lint` runs whatever `golangci-lint` happens to be in `$PATH`, while CI has its version hardcoded
  separately in two workflows. Local and CI can therefore use different versions, and we need to keep those definitions in sync manually. With this change, there is a single pinned version used by everything. And we're also extending that to all other Go tools we use in this repo.

  It is not perfect reproducibility because the local version is built with the local Go version and our `tools` module dependency graph, which is the reason golangci-lint does not recommend
  this approach. However, I think in practical terms this is more than reproducible enough, and if it was not CI (which runs the canonical release binary) would catch it.
  
 I'm not against using mise, but I think adopting it for all tools and replacing Makefiles/scripts is a larger change that we should discuss separately. For example, using a `tools` module allows us to manage tools like `stringer` or `gqlgen` that AFAIK mise does not support.

endigma:
```toml
#mise.toml
[tools]
go = "1.25"
golangci-lint = xyz
"go:golang.org/x/tools/cmd/stringer" = "latest"
"go:github.com/99designs/gqlgen" = "latest"
```

works, and we don't have to replace make/scripts now, mise is a perfectly competent environment/tools manager without any of that.

even a mixture of go tool and mise should be fine, and in any case this PR is an improvement to reproducibility

fiam: Thanks for pointing that out. I wasn’t aware mise could manage arbitrary Go tools.

That said, I still think adopting mise is a discussion for another day. `go tool` is built into Go, uses concepts Go developers are already familiar with, and doesn’t introduce another tool as part of this change.

As you said, this PR already improves reproducibility, and it doesn’t prevent us from moving to mise later.

endigma (approved):

Noroth (approved):

PR #1634, opened 2026-08-12
chore: enable perfsprint linter

fiam:
## Summary

- enable the `perfsprint` golangci-lint check
- exclude `_test.go` files, where formatting readability matters more than micro-optimizations
- replace avoidable `fmt.Sprintf` and static `fmt.Errorf` calls
- use `strconv.Itoa` for integer formatting
- replace repeated string concatenation in loops with `strings.Builder`

This clears 155 existing `perfsprint` findings across `v2` and `execution` while leaving 14 test-only findings unchanged.

References ROUTER-616

## Validation

- golangci-lint v2.10.1 with `GOOS=linux` in `v2`
- golangci-lint v2.10.1 with `GOOS=linux` in `execution`
- `make test-quick` in `v2` (5,700 tests, 45 skipped, plus benchmark compile/run)
- `make test-quick` in `execution` (910 tests, 4 skipped, plus benchmark compile/run)
- `git diff --check`

## Summary by CodeRabbit

* **Refactor**
  * Improved efficiency and consistency in string construction, error handling, JSON rendering, request generation, and reporting.
  * Preserved existing outputs, validation behavior, error messages, and generated responses.
  * Maintained consistent behavior across service responses and error reporting.

* **Chores**
  * Enabled additional performance-focused linting to support ongoing code quality improvements.
  * Simplified internal processing without changing externally visible functionality.


ysmolski:
I will review it

execution/subscription/engine.go
```diff
@@ -25,7 +24,7 @@ func (e *errOnBeforeStartHookFailure) Unwrap() error {
 }
 
 func (e *errOnBeforeStartHookFailure) Error() string {
-	return fmt.Sprintf("on before start hook failed: %s", e.wrappedErr.Error())
+	return "on before start hook failed: " + e.wrappedErr.Error()
```
ysmolski, line 27: I would never expect that this is faster version of making strings

endigma (changes requested):
I believe the swapping errors.New is not necessary anymore as of [1.26](https://go.dev/doc/go1.26) due to a compiler optimization, possibly other cases are now also equivalent

The linter and associated benchmarking was done in 2019, and I do not think we should accept this change without our own benchmarks proving a noticeable improvement to make up for the readability issues

Happy to keep the obvious optimization bits though e.g swapping Sprintf for Fprintf where it makes sense

ysmolski:
I tend to agree with @endigma on this. We should not enable all of the checks of this linter, but rather pick only those that make sense both from the performance and readability POVs.

execution/subscription/engine.go
fiam: This makes a big difference in relative terms because `Sprintf` needs to scan the string, find the `%s` and copy parts of the source string plus the dynamic string into the new string. Also because `Sprintf` takes variadic arguments, it requires allocating a slice of any to call it and boxing the arguments in `any`. 


```
                               sec/op
  fmt.Sprintf                  41.45 ns ± 3%
  concatenation                14.70 ns ± 2%

                               B/op    allocs/op
  fmt.Sprintf                  80      2
  concatenation                64      1
```

For these small functions it also has a compounding effect because it typically takes make them "inlinable".

This is also visible in the generated assembly (I use `go build -gcflags -S 2>&1|less` and then look for the function name, but there are other ways that you might like better)

With `Sprintf` in my computer it ends up like this (omitting everything before the call to `wrappedErr.Error()`, which is the same):

```
  CALL    (R1)                        

  STP     (ZR, ZR), autotmp_3-16(SP)
  CALL    runtime.convTstring(SB) 
  MOVD    $type:string(SB), R1
  STP     (R1, R0), autotmp_3-16(SP)
  MOVD    $"on before start hook failed: %s"(SB), R0
  MOVD    $31, R1
  MOVD    $autotmp_3-16(SP), R2
  MOVD    $1, R3
  MOVD    R3, R4 
  CALL    fmt.Sprintf(SB)
  RET
```

While just string concat is way sorter.

```
  CALL    (R1)
  MOVD    $29, R2
  MOVD    $"on before start hook failed: "(SB), R1
  CALL    runtime.concatstring2(SB)
  RET
```

fiam:
> I believe the swapping errors.New is not necessary anymore as of [1.26](https://go.dev/doc/go1.26) due to a compiler optimization, possibly other cases are now also equivalent
> 
> The linter and associated benchmarking was done in 2019, and I do not think we should accept this change without our own benchmarks proving a noticeable improvement to make up for the readability issues
> 
> Happy to keep the obvious optimization bits though e.g swapping Sprintf for Fprintf where it makes sense

While now it doesn't allocate more than `errors.New`, `fmt.Errorf` still needs to scan the string for `%`, see the [commit](https://github.com/golang/go/commit/c1e6e49d5d3f3fb927f1bfd1b453d8e7c906c6ac#diff-5c6f09263bdb45193c8e145d16bda6a8ee881c566c5978fd704e74c9f3e093b6R36).

I do agree that after Go 1.26 this is a bit opinionated for the advantage that it brings, so I've disabled that check specifically. The rest are still enabled for all non-test files. PTAL!

endigma (approved):

PR #1636, opened 2026-08-14
chore: add cache control parser

Noroth:
## Checklist

## Open Source AI Manifesto

## Summary by CodeRabbit

- **New Features**
  - Added support for parsing HTTP `Cache-Control` response headers.
  - Recognizes `max-age`, `s-maxage`, `no-store`, `no-cache`, `public`, and `private` directives.
  - Supports multiple header values, quoted field lists, case-insensitive directives, duplicate directives, and unknown directives.
  - Validates malformed directives and safely handles excessively large cache durations.

- **Tests**
  - Added comprehensive coverage for valid, invalid, duplicate, malformed, and edge-case header inputs, including fuzz testing.


v2/pkg/engine/cache/cache_control.go
```diff
@@ -0,0 +1,351 @@
...
+	"strconv"
+	"strings"
+	"time"
+)
+
+// FieldNames is a map of field names.
+type FieldNames struct {
+	// lockFields is a flag to prevent adding any fields.
+	// This is set when an empty field list is parsed.
+	lockFields bool
```
fiam: question: do we need this functionality? It doesn't seem particularly well done either, since `Add` doesn't give any feedback to the caller.

v2/pkg/engine/cache/cache_control.go
```diff
@@ -0,0 +1,351 @@
...
+
+	// Present indicates whether the directive is present in the response.
+	Present bool
```
fiam: question: could this be done via returning a `nil` when field names are not present?

v2/pkg/engine/cache/cache_control.go
```diff
@@ -0,0 +1,351 @@
...
+	// Fields is a list of field names.
+	// Do not add any fields to this map directly. Use the add method instead
+	Fields map[string]struct{}
```
fiam: nit: this might be better expressed as an API contract (i.e. not export `Fields`)

v2/pkg/engine/cache/cache_control.go
```diff
@@ -0,0 +1,351 @@
...
+
+func NewFieldNames() *FieldNames {
+	return &FieldNames{
+		// By default, when a FieldNames is created, it is present.
+		Present: true,
+	}
+}
+
+func (f *FieldNames) Add(name string) {
+	if f.Fields == nil {
```
fiam: nit: assuming most of the time a Cache-Control header is present it will have non-zero fields, nitializing `.Fields` on `NewFieldNames` might be more efficient.

v2/pkg/engine/cache/cache_control.go
```diff
@@ -0,0 +1,351 @@
...
+		if field == "" {
+			continue
+		}
+
+		fields.Add(field)
+		added = true
+	}
+
+	if !added {
+		fields.ClearAndLock()
```
fiam: question: isn't this redundant? if we haven't added any fields, what does clearing the inner map do?

v2/pkg/engine/cache/cache_control.go
```diff
@@ -0,0 +1,351 @@
...
+	// whole response, not just a few fields. So an empty argument wins, and we
+	// lock the set so a later occurrence can't weaken it.
+	if !arg.present {
+		fields.ClearAndLock()
+		return
+	}
+
+	var added bool
+	for field := range strings.SplitSeq(arg.text, ",") {
+		field = strings.TrimSpace(field)
```
fiam, line 286: question: could the lexer/parser take care of this and absorb non-significant whitespace?

v2/pkg/engine/cache/cache_control.go
```diff
@@ -0,0 +1,351 @@
...
+			return nil
+		}
+		seen.sMaxAge = true
+
+		if value, ok := deltaSecondsArgument(arg); ok {
+			cc.SMaxAge = &value
+		}
+
+	case "no-store":
+		// A stray argument means nothing to a boolean directive; discard it.
```
fiam: question: isn't this comment a bit confusing? it says `discard it` but then we set a variable to `true` just after it.

fiam (commented):
not an exhaustive review since this is still in draft, just some comments. Thanks!

v2/pkg/engine/cache/cache_control.go
Noroth: updated to `noFields`

v2/pkg/engine/cache/cache_control.go
```diff
@@ -0,0 +1,351 @@
...
+
+	if num > math.MaxInt32 {
+		return DeltaSeconds(math.MaxInt32), nil
+	}
+
+	return DeltaSeconds(int32(num)), nil
+}
+
+// ToGoSeconds converts a DeltaSeconds to a time.Duration.
+func (d DeltaSeconds) ToGoSeconds() time.Duration {
```
fiam: nit: maybe just `Duration()` is more idiomatic?

v2/pkg/engine/cache/cache_control.go
Noroth: Yeah we can do that

v2/pkg/engine/cache/cache_control.go
Noroth: Made fields private and added an iterator function

v2/pkg/engine/cache/cache_control.go
Noroth: done

Noroth:
@coderabbitai review

SkArchon (commented):
I think some verbs are missing like :"stale-while-revalidate / stale-if-error". Might be worth adding for correctness even though I don't think we will consider these in the first versions.

v2/pkg/engine/cache/cache_control.go
```diff
@@ -0,0 +1,358 @@
...
+func ParseCacheControlResponse(headers http.Header) (*CacheControlResponse, error) {
+	// Nothing but whitespace and delimiters carries no directives, which is
+	// indistinguishable from the header being absent.
+	value := strings.Trim(strings.Join(headers.Values("Cache-Control"), ","), "\t\r\n")
+	if value == "" {
+		return &CacheControlResponse{}, nil
+	}
+
+	return parse(value)
+}
```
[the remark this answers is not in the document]
Noroth: I think it is okay in this scenario. We don't have to be too strict with the overall input string. We still check the individual elements within the input string for invalid characters. Trimming the input string is fine.

v2/pkg/engine/cache/cache_control.go
Noroth: We could but I think this is okay for now. Moving the logic to the lexer would make it more complicated to read and we would have to emit more tokens and incorporate that in the parser logic.

SkArchon (approved):
lgtm

v2/pkg/engine/cache/cache_control.go
```diff
@@ -0,0 +1,358 @@
...
+	if num > math.MaxInt32 {
+		return DeltaSeconds(math.MaxInt32), nil
+	}
+
+	return DeltaSeconds(int32(num)), nil
+}
+
+// AsDuration converts a DeltaSeconds to a time.Duration.
+func (d DeltaSeconds) AsDuration() time.Duration {
+	return time.Duration(d) * time.Second
```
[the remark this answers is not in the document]
Noroth: This PR adds the function, there are no consumers

PR #1637, opened 2026-08-15
fix: one stuck subscriber write no longer blocks heartbeats for other triggers

ebachle:
## Summary

`subscriptionState.sendHeartbeat` acquires `writeMu` with a blocking `Lock()`. That same mutex
is held for the full duration of a downstream data push in `executeSubscriptionUpdate`, which
guards `Write`/`Flush`/`Complete`/`Heartbeat`/`Error` against interleaving. Because the SSE
transport sets no write deadline (`WriteTimeout: 0`, deliberately, to support long-lived
streaming), a single subscriber whose connection stops draining — without closing — can hold
`writeMu` indefinitely.

`heartbeatTriggerSubscriptions` / `sendTriggerHeartbeats` process every trigger on the resolver
sequentially, in one goroutine. So one subscriber's stuck write doesn't just delay its own
heartbeat — it blocks the entire heartbeat sweep from ever reaching any other trigger on the
process, for as long as the write stays stuck.

## Impact, as observed in production

We run a Federation deployment (Cosmo router) with a mix of per-entity subscription topics and
one large shared topic (`graphql-events`) carrying updates for many different entities,
deduplicated by the engine onto a single underlying Kafka consumer. A production `pprof` capture
during a real incident caught this directly, not inferred after the fact:

- One goroutine blocked for a confirmed 3 minutes inside `executeSubscriptionUpdate`, at
  `sub.writer.Flush()` — a genuine OS-level socket write, not merely slow (Go's goroutine dump
  resets `waitsince` on every unpark, so a sustained multi-minute annotation on a single wait means
  it truly never made progress in that window).
- A second goroutine — the heartbeat sweep — blocked the same 3 minutes trying to acquire the
  exact same `writeMu`, confirmed via an identical pointer value across both stack traces (the
  `sub *subscriptionState` argument), not inferred from timing alone.
- Because the affected subscriber happened to be on the shared `graphql-events` topic, that
  topic's own Kafka consumer (which calls into the update path synchronously per record) was
  blocked for the same window — cross-checked against our internal throughput metrics for that
  topic, which show a hard stop in messages received for ~3.5 minutes, immediately followed by a
  backlog-catch-up burst once the connection was resolved. Every other entity's real-time updates
  on that shared topic went dark for the duration, not just the stuck subscriber's own.
- We confirmed this recurs: the same paired signature (a connection-count metric crashing,
  together with the shared topic's throughput hard-stopping) showed up 3 times in a single night
  on one deployment, not once.

Net effect: one client that goes silent without a clean disconnect — an ordinary real-world
failure mode (backgrounded app, network drop, client crash) — can silently starve heartbeats *and*
real event delivery for every other subscriber sharing that heartbeat sweep or that topic, not
just its own connection.

## Why this specifically hit SSE, not WebSocket

This incident happened over SSE. WebSocket subscriptions aren't exposed to the same *duration* of
risk, because the router already sets a real write deadline for that transport
(`ENGINE_WEBSOCKET_SERVER_WRITE_TIMEOUT` / `ENGINE_WEBSOCKET_CLIENT_WRITE_TIMEOUT`, both defaulting
to 10s) — a stuck WebSocket write fails within a bounded window, releases `writeMu`, and the
subscription gets cleaned up normally through the existing flush-error path. SSE has no equivalent:
its HTTP server is configured with `WriteTimeout: 0`, deliberately, to support long-lived streaming,
which also means nothing bounds how long a write can hang once a client goes silent. The fix for
this class of problem already exists in this codebase's design for one transport; it was just never
extended to the other.

That write-timeout lives in the router-level transport wiring, not in this package, so it's outside
the scope of this PR — `graphql-go-tools`'s `resolve` package is transport-agnostic and has no
access to the concrete writer's underlying connection to set a deadline on. We plan to add an
equivalent deadline in our own router integration separately, but wanted to flag it here since it's
directly relevant: it's a **complementary** fix, not a substitute for this one. `TryLock` stops one
stuck write from taking every other subscriber down with it; a write deadline would additionally
stop that stuck write itself from lingering indefinitely, cleaning up the dead connection instead of
holding a goroutine and socket open forever. Both matter; neither replaces the other. If it's useful,
`resolve` could expose an optional interface (e.g. a writer that implements a `SetWriteDeadline`-style
method) that `executeSubscriptionUpdate` checks for and calls before each `Flush()`, so any router
integration could opt in without `resolve` needing transport-specific knowledge — happy to help with
that as a follow-up if it's a direction you'd want.

## Fix

`subscriptionState.sendHeartbeat` now uses `writeMu.TryLock()` instead of `Lock()`. On contention
— meaning a write is genuinely in progress for that subscriber right now — it skips sending a
heartbeat this cycle and returns `nil`, rather than blocking or treating it as a heartbeat
failure. This is safe: contention only tells us a write is in flight, not that the subscriber is
gone, so this deliberately does not call `UnsubscribeSubscription`. The next heartbeat tick
retries normally once the write completes (or the connection is eventually torn down some other
way).

This is intentionally the smallest change that closes the hole. It does not restructure
`sendTriggerHeartbeats` / `heartbeatTriggerSubscriptions` into a concurrent sweep — that's a
larger change this specific bug doesn't require. Once a stuck subscriber's own heartbeat attempt
no longer blocks, the sequential sweep simply moves past it in microseconds instead of stalling
for however long the write stays stuck.

No behavior changes for any subscriber whose write isn't currently in flight. The only observable
difference for the stuck subscriber itself is that it stops receiving heartbeats while its write
is stuck (it was never receiving them reliably before either, since its own heartbeat call was
the one blocking) — the actual fix is that this no longer has any effect on other subscribers.

## Tests

Two tests reproduce the bug from different angles, and both are inverted in this PR to assert
the fixed behavior instead:

- `TestResolver_StuckWriteDoesNotBlockHeartbeatForUnrelatedTrigger` — locks a subscription's
  `writeMu` directly to stand in for a stuck write, and asserts an unrelated trigger's heartbeat
  still fires promptly.
- `TestResolver_RealDeadTCPClientDoesNotBlockHeartbeatForUnrelatedTrigger` — uses a real TCP
  connection where the "client" shrinks its own receive buffer and never reads, so the write
  genuinely blocks at the OS level rather than being simulated, and asserts the same holds.

Both were verified in both directions: they fail against the pre-fix code (confirming they
exercise the actual bug, not passing regardless) and pass against this change, consistently under
`-race`. Full `pkg/engine/resolve` suite passes with `-race`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved heartbeat delivery so a stalled connection does not block heartbeats for other active connections.
  - Heartbeats now skip connections currently busy with another write and continue processing independently.
  - Preserved existing connection-removal checks and heartbeat behavior.
- **Tests**
  - Added coverage for stalled and non-responsive connections, including real network scenarios, to verify timely heartbeat delivery.


ebachle:
Update: we've now reproduced this **end-to-end in a real deployment**, not just in the unit tests in this PR. Sharing the evidence in case it's useful for review.

## Setup

Router running **graphql-go-tools v2.15.0**, single replica, verified unfixed before the run (`sendHeartbeat` still takes a blocking `writeMu.Lock()`; `sendTriggerHeartbeats` still walks every trigger sequentially in one goroutine).

Two SSE subscribers connect **directly to the router pod**, both draining their sockets normally:

| probe | subscription | Kafka trigger |
|---|---|---|
| victim | a subscription on the shared events topic | trigger A |
| observer | a different subscription on a different topic | **trigger B** |

The observer is deliberately **idle** — it receives no data, so heartbeats are its only frames. (Necessary: `heartbeatTriggerSubscriptions` skips any subscription that wrote within `heartbeatInterval`, so an active observer shows no symptom.)

We then **blackhole the victim's inbound path** with an `iptables -j DROP` rule scoped to that one socket. No FIN, no RST — the client just stops acknowledging, which models a real dirty disconnect (NAT rebind, closed laptop, dropped mobile connection) rather than a client that politely stops reading.

## What happened

The router's send queue for the victim climbed and then **plateaued at 382,299 bytes** — that plateau is the blocked `Flush()`. Both goroutines, from a goroutine profile taken during the stall:

```
internal/poll.(*FD).Write(...)                               <- blocked in the kernel
core.(*HttpFlushWriter).Flush(0x61142e6fd270)                subscription_response_writer.go:154
resolve.executeSubscriptionUpdate(..., 0x6114328cd620, ...)  resolve.go:1077
resolve.(*Resolver).handleTriggerUpdate.func1()              resolve.go:1505

goroutine 171 [sync.Mutex.Lock, 2 minutes]:
internal/sync.(*Mutex).lockSlow(0x6114328cd668)              mutex.go:149
resolve.(*subscriptionState).sendHeartbeat(0x6114328cd620)   resolve.go:974
resolve.(*Resolver).executeSubscriptionHeartbeat(...)        resolve.go:1107
resolve.(*Resolver).heartbeatTriggerSubscriptions(...)       resolve.go:1552
resolve.(*Resolver).sendTriggerHeartbeats(...)               resolve.go:1633
resolve.(*Resolver).heartbeatLoop(...)                       resolve.go:1619
```

These are provably the **same** `*subscriptionState`, not merely the same lock type: `executeSubscriptionUpdate`'s third argument and `sendHeartbeat`'s receiver are both `0x6114328cd620`, and the contended mutex `0x6114328cd668` is that address + `0x48` — the `writeMu` field within that struct.

## The blast radius, measured from a client

The observer — **different trigger, own socket, never touched** — went **339 seconds without a single heartbeat**, against a rock-steady 5000ms ± 1ms cadence before and after:

| event | time (UTC) |
|---|---|
| victim blackholed | 19:15:18 |
| observer's last heartbeat | 19:15:29 |
| observer's next heartbeat | 19:21:08 (**gap 339.0s**) |

One stuck subscriber starved an unrelated subscription on an unrelated trigger for five and a half minutes. That is the behaviour this PR's `TryLock` change removes.

## Two notes that may be useful beyond this PR

**1. Time-to-stall is governed by bytes, not events.** A blocked `Flush()` requires filling the socket send buffer, which autotunes on a healthy connection — we measured a plateau at ~382KB, well above the `tcp_wmem` initial default. With a small per-frame payload (~386 bytes at ~1 frame/s) this takes hours and looks unreproducible; with a large selection set (~62KB/frame) the same subscription plugs the socket in **under 12 seconds**. Anyone trying to reproduce this against a real router should turn the *selection set* up, not the event rate.

**2. Recovery is not prompt once the path heals.** After we removed the DROP rule, the stall persisted for at least the 35 seconds we observed before intervening. By then the RTO had backed off to `TCP_RTO_MAX` (120s), so the next retransmission — the thing that would have recovered the connection — was up to two minutes away. With a permanently vanished peer and no middleware, the ceiling is `tcp_retries2` (~924s on our nodes) before the kernel returns `ETIMEDOUT` to the blocked write.

That second point is the practical severity argument for the SSE write-deadline discussed in the PR description: a client that vanishes without closing can hold `writeMu` for minutes, and nothing at the transport layer currently bounds it, since SSE has no equivalent of `ENGINE_WEBSOCKET_SERVER_WRITE_TIMEOUT`.

Happy to share a self-contained reproduction harness if that would help — the version we ran is coupled to our own infrastructure, but the mechanism only needs a router, one subscriber that stops acknowledging, and one idle subscriber on a second trigger.

v2/pkg/engine/resolve/resolve.go
```diff
@@ -969,9 +969,19 @@ func (s *subscriptionState) writeError(w AsyncErrorWriter, ctx *Context, err err
 }
 
 // sendHeartbeat sends a keep-alive frame to the downstream writer under writeMu.
-// @TODO: this is bad, see ENG-9356
+//
```
fiam, line 972: issue: I think most if not all of this block belongs inside the function instead of in its documentation. Could you please move the implementation details inside?

v2/pkg/engine/resolve/resolve.go
```diff
@@ -969,9 +969,19 @@ func (s *subscriptionState) writeError(w AsyncErrorWriter, ctx *Context, err err
...
+// downstream write stays unresponsive (this transport sets no write
+// deadline), and heartbeatTriggerSubscriptions processes every trigger on
+// the process sequentially in one goroutine. Blocking here would let one
+// stuck write freeze heartbeats for every other trigger, not just this one.
+// Skipping this cycle on contention is safe -- contention means a write is
+// genuinely in flight, not that the subscription is gone -- and the next
+// heartbeat tick retries.
 func (s *subscriptionState) sendHeartbeat() error {
-	s.writeMu.Lock()
+	if !s.writeMu.TryLock() {
```
fiam, line 982: question: should this be reported to the upper layer by changing the return type to `(bool, error)`?

v2/pkg/engine/resolve/resolver_heartbeat_blast_radius_realsocket_test.go
```diff
@@ -0,0 +1,199 @@
...
+	// never read from it again below, exactly like a client that silently
+	// stops draining its socket without ever closing the connection.
+	if tcpConn, ok := clientConn.(*net.TCPConn); ok {
+		_ = tcpConn.SetReadBuffer(1024)
+	}
+
+	var serverConn net.Conn
+	select {
+	case serverConn = <-serverConnCh:
+	case <-time.After(time.Second):
```
fiam, line 84: issue: we don't need this explicit timeout in the test, also at a second it is likely flaky. Let it hang if it fails for whatever reason, the global test timeout will eventually trigger anyway.

Also, please void sleep/etc... in tests unless it is within a `synctest` bubble

v2/pkg/engine/resolve/resolver_heartbeat_blast_radius_realsocket_test.go
```diff
@@ -0,0 +1,199 @@
...
+// locking writeMu directly to simulate a stuck write, it uses a real TCP
+// connection to a "client" that stops reading without closing the socket --
+// the exact scenario caught live in production -- so the server's actual
+// Flush() call genuinely blocks at the OS level, not a simulated stand-in,
+// and proves the fix (sendHeartbeat's TryLock) still decouples an unrelated
+// trigger's heartbeat from that real block, without claiming the stuck write
+// itself is in any way resolved by the fix -- it stays genuinely blocked
+// throughout, exactly as a real dead connection would.
+func TestResolver_RealDeadTCPClientDoesNotBlockHeartbeatForUnrelatedTrigger(t *testing.T) {
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
```
fiam, line 53: issue: in this repo we use `testify.require...`, please use it instead of checking for non-nil errors.

v2/pkg/engine/resolve/resolver_heartbeat_blast_radius_realsocket_test.go
```diff
@@ -0,0 +1,199 @@
...
+func (w *tcpFlushWriter) Heartbeat() error {
+	if _, err := w.bw.Write([]byte("H")); err != nil {
+		return err
+	}
+	return w.bw.Flush()
+}
+
+var _ SubscriptionResponseWriter = (*tcpFlushWriter)(nil)
+
+// TestResolver_RealDeadTCPClientDoesNotBlockHeartbeatForUnrelatedTrigger is
```
fiam, line 41: this is super verbose, a single line should be enough to describe the test

v2/pkg/engine/resolve/resolver_heartbeat_blast_radius_test.go
```diff
@@ -0,0 +1,124 @@
...
+import (
+	"context"
+	"sync"
+	"testing"
+	"time"
+)
+
+// RecordingHeartbeatWriter is a SubscriptionResponseWriter whose Heartbeat call
+// is observable from the test via a callback, with no other side effects.
+type RecordingHeartbeatWriter struct {
```
fiam, line 12: question: could we use a single for both tests or maybe even a single test (e.g. reuse `tcpFlushWriter` to build `RecordingHeartbeatWriter`?

issue: `RecordingHeartbeatWriter` should not be exported, should be `recordingHeartbeatWriter`

fiam (commented):
Thanks for your contribution! While the solution looks correct, I think the test and comments still need a bit of polish. Could you please take a look?

PR #1647, opened 2026-08-27
fix: compute operation depth from field nesting only

asoorm:
## Problem

The complexity metric reported wrong `Depth` values for operations with inline fragments.

The estimator derived `Depth` from the depth counter of the AST walker. That counter
increases for fields, selection sets, and inline fragments. A correction term subtracted
the selection sets. It did not subtract inline fragments. It also depended on traversal
order. This caused two defects:

1. Each inline fragment on the deepest path added 2 to the depth.
2. Sibling selections changed the depth. This occurred even when they did not extend
   the deepest path.

Smallest reproduction (true depth 3, reported depth 5):

```graphql
{
  node {
    ... on Success {
      child {
        leaf
      }
    }
  }
}
```

The existing tests did not show this defect. Their only inline fragment had no type
condition. Normalization removed that fragment before the estimator ran. Fragments
with type conditions on unions and interfaces stay in the operation.

## Fix

Count field nesting directly. `EnterField` records the number of field ancestors, plus
itself, plus one leaf level. Fragments do not add depth. The result does not depend on
traversal order.

The depth convention does not change for fragment-free operations. The operation root
is level 0. Each nested field adds one level. Leaf fields also count. An operation that
selects only root scalars has depth 0.

## Behavior change

Operations with inline fragments reported too much depth before this change. After this
change, they can pass depth limits that they failed before. If you calibrated depth
limits against the old values, review those limits.

With help from Claude

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved operation complexity depth calculations for nested fields.
  * Ensured selection sets and fragments do not incorrectly increase depth.
  * Corrected global and root-field depth reporting.
  * Prevented stale depth values when processing is interrupted and reused.

* **Tests**
  * Added coverage for inline fragments on union types.
  * Verified depth remains consistent regardless of sibling branch order.
  * Improved normalization error assertions.

### DIFF
--- a/v2/pkg/middleware/operation_complexity/operation_complexity.go
+++ b/v2/pkg/middleware/operation_complexity/operation_complexity.go
@@ -91,6 +91,11 @@ func (n *OperationComplexityEstimator) Do(operation, definition *ast.Document, r
 	n.visitor.maxFieldDepth = 0
 	n.visitor.multipliers = n.visitor.multipliers[:0]
 
+	// An aborted walk skips the root field cleanup in LeaveField.
+	// Reset the root field state so the next call does not read stale values.
+	n.visitor.currentRootFieldMaxDepth = 0
+	n.visitor.currentRootFieldStats = RootFieldStats{}
+
 	if n.visitor.calculatedRootFieldStats == nil {
 		n.visitor.calculatedRootFieldStats = make([]RootFieldStats, 0, len(definition.RootOperationTypeDefinitions))
 	}
--- a/v2/pkg/middleware/operation_complexity/operation_complexity.go
+++ b/v2/pkg/middleware/operation_complexity/operation_complexity.go
@@ -36,7 +36,10 @@ import (
 type OperationStats struct {
 	NodeCount  int
 	Complexity int
-	Depth      int
+	// Depth is the largest number of nested field levels in the operation.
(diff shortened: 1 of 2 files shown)


v2/pkg/middleware/operation_complexity/operation_complexity.go
```diff
@@ -218,14 +217,30 @@ func (c *complexityVisitor) EnterField(ref int) {
 	}
 
 	c.complexity = c.complexity + c.calculateMultiplied(1)
-	if c.Depth > c.maxFieldDepth {
-		c.maxFieldDepth = c.Depth
+
+	// This field has selections. The selections add one more level.
+	depth := c.fieldNestingLevel() + 1
```
fiam, line 222: blocking: This changes depth tracking from linear to quadratic. Before, we updated counters in O(1) on the walker callbacks, so the full traversal was O(N). Now every field scans all its ancestors, so a query of depth N does 1 + 2 + ... + N work: O(N^2). Inline fragments make this even worse without counting towards depth.

Some quick and dirty benchmarks show a ~5% performance penalty at depth 20 and ~12% at depth 50. Since this is also an input, I think we need to be careful because it can be used to cause a DoS.

I'll see if I can find time to take a crack at it in a different direction.

PR #1648, opened 2026-08-27
fix: correct operation depth for typed fragments

### LINKED ISSUE #1052: depth validation bug
Hello, I'm trying to validate a query using wundergraph tools. I have a problem with the [depth validation](https://github.com/wundergraph/graphql-go-tools/tree/master/v2/pkg/middleware/operation_complexity).

The Operation I have is as follows:
```
query ProductDetails {
  product {
    __typename
    id
    __typename
    externalId
    features { 
      __typename
      feature {
        __typename
        id
      }
      ... on LookupProductCharacteristicValue { 
        __typename
        value {
          __typename
          ... on CustomFeatureRecord {
            __typename
            customFeature {
              __typename 
              key
            }
            id
            name
          }
        }
      }
    }
  }
}
```
The operation depth for the above query is 8. 

When I remove the `feature`, the query will look like this:

```
query ProductDetails {
  product {
    __typename
    id
    __typename
    externalId
    features { 
      __typename
      ... on LookupProductCharacteristicValue { 
        __typename
        value {
          __typename
          ... on CustomFeatureRecord {
            __typename
            customFeature {
              __typename 
              key
            }
            id
            name
          }
        }
      }
    }
  }
}
```
The operation depth for the above query is 9.

So, I removed the `feature` object with some fields, and the query depth is bigger. The same happens when I move the `feature` at the end of the scope, for example, the query that is below also has depth `9`:

```
query ProductDetails {
  product {
    __typename
    id
    __typename
    externalId
    features { 
      __typename
      ... on LookupProductCharacteristicValue { 
        __typename
        value {
          __typename
          ... on CustomFeatureRecord {
            __typename
            customFeature {
              __typename 
              key
            }
            id
            name
          }
        }
      }
      feature {
        __typename
        id
      }
    }
  }
}
```

As far as I understand, the `selectionSetDepth` value is polluted, but only if there is an object with some fields before the spread (`...`).

fiam:
Linear: [ENG-9970](https://linear.app/wundergraph/issue/ENG-9970/depth-calculation-incorrect-under-complexity-limit-metric)

Alternative to #1647. This keeps the fix limited to the existing depth bookkeeping and does not scan field ancestors or register additional visitors.

## Summary

- Add anonymized regressions for the three customer query shapes.
- Track active field-owned selection sets so typed fragments do not affect field depth.
- Add depth 19 and depth 50 benchmarks.

The test-only commit reports depths 19, 18, and 17 instead of 11. The fix makes all three report 11.

## Benchmark

Apple M5 Pro, 8 interleaved samples:

| Depth | master | this PR | Allocations |
| --- | ---: | ---: | ---: |
| 19 | 2.520 µs | 2.468 µs | 29 → 29 |
| 50 | 5.532 µs | 5.405 µs | 41 → 41 |

There is no statistically significant timing change. Memory decreases by 16 B/op at both depths.

## Tests

go test ./v2/... -count=1

## Related issue

Fixes #1052.

This fixes the traversal-order problem reported there: adding, removing, or reordering the shallower `feature` sibling no longer changes the maximum depth around inline fragments. Under the estimator's field-nesting definition, all three query variants report an operation depth of 5.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GraphQL operation-depth calculation for root fields, nested selections, inline fragments, and sibling branches.
  * Depth reporting remains accurate with complex input arguments and configured multipliers.
  * Complexity calculations continue to account for field counts and multiplier effects.

* **Tests**
  * Added regression coverage for root, nested, and stacked multiplier scenarios.
  * Added benchmarks for operations at depths of 19 and 50 to validate depth reporting and resource usage.

### DIFF
--- a/v2/pkg/middleware/operation_complexity/operation_complexity.go
+++ b/v2/pkg/middleware/operation_complexity/operation_complexity.go
@@ -33,12 +33,19 @@ import (
 	"github.com/wundergraph/graphql-go-tools/v2/pkg/operationreport"
 )
 
+// OperationStats contains estimates for an operation or root field.
 type OperationStats struct {
-	NodeCount  int
+	// NodeCount is the maximum number of returned nodes.
+	NodeCount int
+	// Complexity is the maximum number of field requests.
 	Complexity int
-	Depth      int
+	// Depth is the maximum number of response field levels on a path.
+	// Root-field depth is relative to the root.
+	Depth int
 }
 
+// RootFieldStats contains the stats for one top-level field. Alias is empty
+// when the response name is the same as FieldName.
 type RootFieldStats struct {
 	TypeName  string
 	FieldName string
@@ -56,11 +63,15 @@ const (
 	__typeLiteral   = "__type"
 )
 
+// OperationComplexityEstimator estimates stats for normalized operations.
+// It may be reused sequentially, but is not safe for concurrent use.
 type OperationComplexityEstimator struct {
 	walker  *astvisitor.Walker
 	visitor *complexityVisitor
 }
 
+// NewOperationComplexityEstimator creates an estimator. If skipIntrospection
+// is true, __schema and __type root fields are excluded.
 func NewOperationComplexityEstimator(skipIntrospection bool) *OperationComplexityEstimator {
 	walker := astvisitor.NewWalker(48)
 	visitor := &complexityVisitor{
@@ -82,13 +93,14 @@ func NewOperationComplexityEstimator(skipIntrospection bool) *OperationComplexit
 	}
 }
 
+// Do returns global and per-root-field estimates for the operation.
 func (n *OperationComplexityEstimator) Do(operation, definition *ast.Document, report *operationreport.Report) (OperationStats, []RootFieldStats) {
 	n.visitor.count = 0
 	n.visitor.complexity = 0
-	n.visitor.maxFieldDepth = 0
+	n.visitor.maxOperationDepth = 0
 	n.visitor.multipliers = n.visitor.multipliers[:0]
 
-	n.visitor.selectionSetDepth = 0
+	n.visitor.fieldDepth = 0
 
 	if n.visitor.calculatedRootFieldStats == nil {
 		n.visitor.calculatedRootFieldStats = make([]RootFieldStats, 0, len(definition.RootOperationTypeDefinitions))
@@ -107,7 +119,7 @@ func (n *OperationComplexityEstimator) Do(operation, definition *ast.Document, r
 	globalResult := OperationStats{
 		NodeCount:  n.visitor.count,
 		Complexity: n.visitor.complexity,
-		Depth:      n.visitor.maxFieldDepth,
+		Depth:      n.visitor.maxOperationDepth,
 	}
 
 	return globalResult, n.visitor.calculatedRootFieldStats
@@ -125,15 +137,24 @@ type complexityVisitor struct {
 	operation, definition *ast.Document
 	count                 int
 	complexity            int
-	maxFieldDepth         int
-	multipliers           []multiplier
 
-	selectionSetDepth int
+	// maxOperationDepth includes the root field.
+	maxOperationDepth int
+
+	// multipliers contains @nodeCountMultiply argument values for the active
+	// field path.
+	multipliers []multiplier
+
+	// fieldDepth counts active fields with selections. Fragments add no depth.
+	fieldDepth int
 
 	rootOperationTypeNames map[string]struct{}
 
-	currentRootFieldStats    RootFieldStats
-	currentRootFieldMaxDepth int
+	// currentRootFieldStats is reused because root fields are visited depth-first.
+	currentRootFieldStats RootFieldStats
+
+	// maxRootFieldDepth is relative to the current root field.
+	maxRootFieldDepth int
 
 	calculatedRootFieldStats []RootFieldStats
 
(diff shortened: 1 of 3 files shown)


v2/pkg/middleware/operation_complexity/operation_complexity.go
```diff
@@ -88,7 +88,6 @@ func (n *OperationComplexityEstimator) Do(operation, definition *ast.Document, r
 	n.visitor.maxFieldDepth = 0
```
Noroth: I know it's not part of the PR, but this component lacks a bit of documentation. It would be good to understand with some comments and examples how the logic was implemented.

v2/pkg/middleware/operation_complexity/operation_complexity_test.go
```diff
@@ -582,6 +608,111 @@ const complexQuery = `
...
+
+type AlternateOne { alternateTwo: AlternateTwo }
+type AlternateTwo { alternateThree: AlternateThree }
+type AlternateThree { alternateFour: AlternateFour }
+type AlternateFour { alternateFive: AlternateFive }
+type AlternateFive { alternateSix: AlternateSix }
+type AlternateSix { alternateLeaf: String }
+`
+
+const depthRegressionQuery = `
```
Noroth: Did you also test arguments with a complex type?

* arguments on root fields
* arguments on fields within the query
* arguments on leaf fields
* arguments on fields inside of inline fragments.

Noroth (changes requested):

v2/pkg/middleware/operation_complexity/operation_complexity_test.go
fiam: I limited myself to the regression we knew we had and the original PR was covering, but you're right. I will check what is actually covered and what is not and I'll get back to you.

fiam: I've added a more tests trying to cover leaf fields and inline fragments, I think we had good coverage for the other two. PTAL and let me know, thanks!

v2/pkg/middleware/operation_complexity/operation_complexity.go
fiam: Thanks for the idea! I took the chance and also refactored a bit to unify the coordinate systems and use `max()` instead of `if` - PTAL

fiam:
Pinging @ysmolski for additional validation, since you edited these files not long ago. PTAL.

v2/pkg/middleware/operation_complexity/operation_complexity_test.go
```diff
@@ -582,6 +682,115 @@ const complexQuery = `
...
+union RootResult = RootSuccess | Failure
+type RootSuccess { results: [Result!]! }
+union Result = ResultSuccess | Failure
+type ResultSuccess { result: Record }
+type Failure { message: String }
+
+interface Record { pathA(input: RootInput): PathA alternateOne: AlternateOne }
+type BasicRecord implements Record { pathA(input: RootInput): PathA alternateOne: AlternateOne }
+type ExtendedRecord implements Record { pathA(input: RootInput): PathA alternateOne: AlternateOne sideLeaf: String branchOne: BranchOne }
+
```
ysmolski: This particular part of the definition is hard to read. And it also applied to other type with more than 2 fields.

v2/pkg/middleware/operation_complexity/operation_complexity_test.go
```diff
@@ -582,6 +682,115 @@ const complexQuery = `
...
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}`
```
ysmolski: Tests using this query are impossible to read and verify. Too deep parametrization leads to some saved lines count but makes all tests look like a mashed web instead each test being isolated (using just the query and smallest possible reproduction) and easy to comprehend.

ysmolski (changes requested):

ysmolski:
@fiam Please check if this closes https://github.com/wundergraph/graphql-go-tools/pull/1626 and https://github.com/wundergraph/graphql-go-tools/issues/1052

Noroth (approved):

fiam:
> @fiam Please check if this closes #1626 and #1052

Fixes #1052, I've added it to the PR description. https://github.com/wundergraph/graphql-go-tools/pull/1626 has a fix for an unrelated issue. I'll open a separate PR for that.

v2/pkg/middleware/operation_complexity/operation_complexity.go
```diff
@@ -217,18 +233,22 @@ func (c *complexityVisitor) EnterField(ref int) {
...
 
+	// A field's multiplier applies to its result, not its own request.
 	c.complexity = c.complexity + c.calculateMultiplied(1)
-	if c.Depth > c.maxFieldDepth {
-		c.maxFieldDepth = c.Depth
-	}
+	c.fieldDepth++
+
+	// Operation depth includes the selected child. Root depth is root-relative.
+	c.maxOperationDepth = max(c.maxOperationDepth, c.fieldDepth+1)
```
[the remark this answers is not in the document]
fiam: Looks like this was done intentionally in https://github.com/TykTechnologies/graphql-go-tools/pull/73 — we'll follow up on this if needed on a subsequent PR

v2/pkg/middleware/operation_complexity/operation_complexity_test.go
fiam: Addressed in e6d5feab: expanded the multi-field schema definitions to one field per line.

v2/pkg/middleware/operation_complexity/operation_complexity_test.go
fiam: Addressed in 92fb91d2: replaced the shared format-string query with isolated minimal queries and a small common schema. The regressions still fail against the pre-fix implementation.

ysmolski (approved):
Tests are easy to read now. Thanks

PR #1649, opened 2026-08-31
feat: report TTL for subgraph and cached responses

Noroth:
Allow the engine to propagate response cache TTLs

## Summary by CodeRabbit

* **New Features**
  * Added support for generating `Cache-Control` header values.
  * Exposed whether responses were served from cache and their remaining cache lifetime.
  * Cache hits now report the shortest applicable TTL across cached entries.

* **Bug Fixes**
  * Improved handling of zero and negative cache lifetimes.
  * Ensured cached responses do not include subgraph response headers.
  * Improved cache reporting for uncached responses and responses with no remaining lifetime.

## Checklist

## Open Source AI Manifesto

### DIFF
--- a/v2/pkg/engine/resolve/loader_test.go
+++ b/v2/pkg/engine/resolve/loader_test.go
@@ -1,6 +1,7 @@
 package resolve
 
 import (
+	"cmp"
 	"context"
 	"encoding/json"
 	"net/http"
@@ -2783,14 +2784,18 @@ func TestLoader_CachedFetches(t *testing.T) {
 		loader := &Loader{dataBuffer: &DataBuffer{data: astjson.ObjectValue(nil)}}
 
 		// Fetches run in parallel, so collect and sort before asserting.
+		type cacheReport struct {
+			hit bool
+			ttl time.Duration
+		}
 		var mu sync.Mutex
-		var ttls []time.Duration
-		collectTTLs := func() []time.Duration {
+		var reports []cacheReport
+		collectReports := func() []cacheReport {
 			mu.Lock()
 			defer mu.Unlock()
-			out := slices.Clone(ttls)
-			ttls = ttls[:0]
-			slices.Sort(out)
+			out := slices.Clone(reports)
+			reports = reports[:0]
+			slices.SortFunc(out, func(a, b cacheReport) int { return cmp.Compare(a.ttl, b.ttl) })
 			return out
 		}
 		ctx.SetEngineLoaderHooks(&spyLoaderHooks{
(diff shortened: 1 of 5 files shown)


v2/pkg/engine/resolve/response_cache.go
```diff
@@ -106,10 +107,24 @@ func (l *Loader) responseCacheLookup(prepared *preparedFetch) bool {
...
 	return true
 }
 
+// remainingTTL is the shortest life left across a fetch's entries: a fetch is only
+// as fresh as its least fresh entry. Non-positive TTLs are ignored, GetMany already
+// treats those as misses.
+func remainingTTL(found map[string]caching.Item, keys []string) time.Duration {
+	var ttl time.Duration
+	for _, key := range keys {
+		if t := found[key].TTL; t > 0 && (ttl == 0 || t < ttl) {
```
SkArchon: Was a bit hard to read, maybe currentTtl, etc

v2/pkg/engine/resolve/response_cache_test.go
```diff
@@ -71,3 +73,45 @@ func TestRootFetchCacheable(t *testing.T) {
 		require.False(t, rootFetchCacheable(rootItem(), fetch))
 	})
 }
+
+func TestRemainingTTL(t *testing.T) {
```
SkArchon, line 77: Just a note, when we introduce private, I think we will need to update this logic. Because private overrides non private.

Since we won't fundamentally cache a `no-store`, it makes sense we dont need to consider it.

v2/pkg/engine/resolve/loader_test.go
```diff
@@ -2706,36 +2684,188 @@ func TestLoader_CachedFetches(t *testing.T) {
...
+
+		testCache := newTestCache()
+
+		ctx := NewContext(context.Background())
+		ctx.SetResponseCache(testCache, time.Second*60, nil)
+		resolvable := NewResolvable(nil, ResolvableOptions{})
+		loader := &Loader{dataBuffer: &DataBuffer{data: astjson.ObjectValue(nil)}}
+		err := resolvable.Init(ctx, nil, ast.OperationTypeQuery)
+		assert.NoError(t, err)
+		err = loader.LoadGraphQLResponseData(ctx, response)
```
SkArchon, line 2729: Can you make it clearer (by comment and spacing around it) that this is the first call, and do the same for the second call, and other instances in these tests, was hard to figure out which was the actual call.

SkArchon (changes requested):
Just one comment, also ensure that the coderabbit comments are addressed

v2/pkg/engine/resolve/response_cache_test.go
Noroth: Does this matter in this case? Usually we will only receive private from the subgraph and not from the cache. Therefore the router will apply the most restrictive rule anyways. We just report what we get on each fetch.

v2/pkg/engine/resolve/loader_test.go
Noroth: Done

v2/pkg/engine/resolve/response_cache_test.go
SkArchon: For this pr doesnt matter, we will need to cache `private` at some point. So that would be the most restrictive.

SkArchon (approved):
lgtm

PR #1650, opened 2026-08-31
feat: support cache tag creation via extensions in response from subgraphs

SkArchon:
This PR introduces support of extracting tags from subgraph response extensions and storing them in the adapter. We use the following format

```
{
   "data": {"_entities": [
       {"__typename": "Student", "id": 1, "name": "Jack"},
       {"__typename": "Student", "id": 2, "name": "John"},
       {"__typename": "Student", "id": 3, "name": "Ann"},
       {"__typename": "Student", "id": 4, "name": "Paul"},
       {"__typename": "Student", "id": 5, "name": "Kylie"}
   ]},
   "extensions": {"apolloEntityCacheTags": [
       ["student", "student-1"],
       ["student", "student-2", "superuser"],
       ["student", "student-3"],
       ["student", "student-4"],
       ["student", "student-5"]
   ]}
}
```

The data from `apolloEntityCacheTags` is mapped to the same index in data._entities. Related PR https://github.com/wundergraph/cosmo/pull/3200

## Summary by CodeRabbit

* **New Features**
  * Added cache tags for targeted response-cache invalidation.
  * Added cache-tag key generation for subgraphs, types, and declared tags, with subgraph isolation.
  * Cache entries can retain associated tags for invalidation.

* **Bug Fixes**
  * Improved tag handling for root and entity responses.
  * Prevented inappropriate type indexing and protected tag keys from separator collisions.

* **Tests**
  * Expanded coverage for tag formatting, isolation, invalidation, and response-cache behavior.

## Checklist

## Open Source AI Manifesto


v2/pkg/engine/resolve/response_cache.go
```diff
@@ -245,7 +358,21 @@ func responseCacheHeaders(res *result) http.Header {
 	return res.httpResponseContext.Response.Header
 }
 
+const (
+	maxResponseCacheTagsPerValue = 10_000
+	maxResponseCacheTagLength    = 10_000
+)
+
+// Where a subgraph attaches its cache tags.
+const (
```
SkArchon, line 383: This is what apollo does

v2/pkg/engine/resolve/response_cache.go
```diff
@@ -245,7 +358,21 @@ func responseCacheHeaders(res *result) http.Header {
...
+	responseCacheEntityTagsExtensionKey = "apolloEntityCacheTags"
```
fiam, line 384: question: should we use our own tags keys and support `apollo`  prefixed ones under a compatibility toggle? It doesn't have to be in this PR, it could be a followup.

v2/pkg/engine/resolve/response_cache_test.go
```diff
@@ -74,6 +81,443 @@ func TestRootFetchCacheable(t *testing.T) {
...
+			responseCacheTagIdentities(nil, parse(t, `{"__typename":""}`), "accounts", all))
+	})
+
+	t.Run("an unnamed subgraph is not indexed at all", func(t *testing.T) {
+		// Every identity is scoped by the subgraph that answered, so without a
+		// name there is no scope to file the entry under. Indexing it unscoped
+		// would put it where another subgraph's invalidation could reach it,
+		// which is worse than not indexing it: the entry still expires on its
+		// own TTL either way.
+		require.Nil(t, responseCacheTagIdentities([]string{"users"}, entity(t), "", all))
```
fiam: nit: I think `require.Empty` would be more correct, a non-nil but empty slice would be an acceptable response

v2/pkg/engine/resolve/response_cache_test.go
```diff
@@ -74,6 +81,443 @@ func TestRootFetchCacheable(t *testing.T) {
...
+	})
+
+	t.Run("nothing to index at all is no tags rather than an empty list", func(t *testing.T) {
+		opts := ResponseCacheInvalidationOptions{CacheTag: true}
+		require.Nil(t, responseCacheTagIdentities(nil, entity(t), "accounts", opts))
```
fiam: question: does this imply that we handle `nil` slice different than a zero length? That might end up biting us later. Could you please help me understand if that's the case and why we do it? Thanks!

v2/pkg/caching/tags.go
```diff
@@ -0,0 +1,37 @@
...
+// contain the separator, so the subgraph is the only segment that could, and
+// splitting is never asked of these: they are built to be looked up, not read.
+func TypeTag(subgraph, typeName string) string {
+	return typeTagPrefix + subgraph + tagScopeSeparator + typeName
+}
+
+// DeclaredTag names a subgraph's entries carrying a tag the subgraph itself
+// declared in its response.
+func DeclaredTag(subgraph, tag string) string {
+	return declaredTagPrefix + subgraph + tagScopeSeparator + tag
```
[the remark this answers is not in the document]
fiam: This seems to me like a legitimate concern

v2/pkg/engine/resolve/context.go
```diff
@@ -257,16 +257,52 @@ func (c *Context) SetRateLimiter(limiter RateLimiter) {
...
-	onError    func(error)
+	store        caching.Cache
+	defaultTTL   time.Duration
+	onError      func(error)
+	invalidation ResponseCacheInvalidationOptions
+}
+
+// ResponseCacheInvalidationOptions selects which secondary indexes are built.
+// Each is independent; all off caches entries untagged.
+type ResponseCacheInvalidationOptions struct {
```
fiam: question: isn't this more like a secondary index builder more than invalidation options?

v2/pkg/engine/resolve/context.go
```diff
@@ -257,16 +257,52 @@ func (c *Context) SetRateLimiter(limiter RateLimiter) {
...
+		defaultTTL: defaultTTL,
+		onError:    onError,
+		// Overridden by SetResponseCacheInvalidation.
+		invalidation: DefaultResponseCacheInvalidationOptions(),
+	}
+}
+
+// SetResponseCacheInvalidation selects the secondary indexes. No effect before
+// SetResponseCache.
+func (c *Context) SetResponseCacheInvalidation(opts ResponseCacheInvalidationOptions) {
```
fiam: question: IMHO it would be more compact to manage this in `SetResponseCache`  instead of having the defaults in `SetResponseCache` and then having to call `SetResponseCacheInvalidation` to set just the invalidation options.

Could we construct a full `ResponseCacheOptions` and pass that to `SetResponseCache`? Maybe I'm missing some context here, please let me know if that makes sense.

v2/pkg/engine/resolve/response_cache_test.go
```diff
@@ -74,6 +81,443 @@ func TestRootFetchCacheable(t *testing.T) {
...
+	t.Run("an object where the array should be is not tags", func(t *testing.T) {
+		response := parse(t, `{"extensions":{"apolloEntityCacheTags":{"users":["user-42"]}}}`)
+		require.Nil(t, responseCacheTags(response, 1, false))
+	})
+}
+
+// An entry is indexed under three separate kinds of thing, and which kind a tag
+// came from has to survive into the index or one could be mistaken for another.
+func TestResponseCacheTagIdentities(t *testing.T) {
+	parse := func(t *testing.T, doc string) *astjson.Value {
```
fiam, line 232: question: could we do all these subtests table-driven? I think it would be more compact

v2/pkg/engine/resolve/response_cache_test.go
```diff
@@ -74,6 +81,443 @@ func TestRootFetchCacheable(t *testing.T) {
...
+		ctx := NewContext(context.Background())
+		ctx.SetResponseCache(newTestCache(), time.Minute, nil)
+		ctx.SetResponseCacheInvalidation(opts)
+
+		res := &result{
+			out:        []byte(body),
+			statusCode: http.StatusOK,
+			httpResponseContext: &httpclient.ResponseContext{
+				Response: &http.Response{
+					Header: http.Header{"Cache-Control": []string{"public, max-age=60"}},
```
fiam, line 409: question: IIRC the `public` should not be needed for the happy-path, just `max-age=60` should do the trick. It doesn't hurt to add it, but it saves a bit of context.

v2/pkg/engine/resolve/response_cache_test.go
```diff
@@ -74,6 +81,443 @@ func TestRootFetchCacheable(t *testing.T) {
...
+
+		require.Len(t, prepared.responseCacheItems, 1)
+		// No type: a root fetch's data object is a selection set rather than an
+		// entity, so there is no one typename it could be indexed under.
+		require.Equal(t, []string{
+			"declared:accounts:employees", "declared:accounts:homepage",
+			"subgraph:accounts",
+		}, prepared.responseCacheItems[0].Tags)
+		require.JSONEq(t, `{"employees":[{"id":1}]}`, string(prepared.responseCacheItems[0].Value))
+	})
```
fiam, line 577: Could we add this regression case? It goes red with the current implementation because a root `__typename` is treated as an entity type.

```suggestion
	})

	t.Run("a root fetch is never indexed by type", func(t *testing.T) {
		body := `{"data":{"__typename":"Query","employees":[{"id":1}]}}`

		loader, res := newLoader(t, body, ResponseCacheInvalidationOptions{Type: true})
		prepared := &preparedFetch{res: res, responseCacheKeys: []string{"root"}, isRootFetchCache: true}
		require.NoError(t, loader.responseCacheCollect(prepared))

		require.Len(t, prepared.responseCacheItems, 1)
		require.Empty(t, prepared.responseCacheItems[0].Tags)
	})
```

v2/pkg/engine/resolve/response_cache.go
```diff
@@ -219,6 +237,101 @@ func responseCacheValues(prepared *preparedFetch, response *astjson.Value) ([]*a
...
+	}
+
+	if opts.Subgraph {
+		identities = append(identities, caching.SubgraphTag(subgraph))
+	}
+
+	if opts.Type {
+		if name := value.GetStringBytes("__typename"); len(name) > 0 {
+			identities = append(identities, caching.TypeTag(subgraph, string(name)))
+		}
```
fiam: This derives the type index solely from the response payload, so a root response containing `__typename: "Query"` is incorrectly indexed as an entity type. The regression test suggested [here](https://github.com/wundergraph/graphql-go-tools/pull/1650#discussion_r3908748162) exposes it. Could we pass the fetch kind into this helper and skip `TypeTag` creation when `prepared.isRootFetchCache` is true?

ETA: disclosure - posted by agent

v2/pkg/engine/resolve/response_cache.go
SkArchon: I was thinking the same, but for a different PR. Will be creating a ticket for it.

v2/pkg/engine/resolve/context.go
SkArchon: We can specify indexes we don't care about in the config, like if we dont want to build an index for "Type" we will past "Type":false in the router yaml.

v2/pkg/engine/resolve/response_cache_test.go
SkArchon: Its there to mock the subgraph response which is a possibility, will remove it with the ticket of making public not mandatory

v2/pkg/engine/resolve/response_cache_test.go
SkArchon: No we dont treat it differently, just the naming of the test is incorrect, will update

v2/pkg/engine/resolve/response_cache_test.go
SkArchon: I found it a bit harder to read when I converted it to a table driven test, but can still change it, lmk

v2/pkg/engine/resolve/response_cache.go
SkArchon: Removed for now, was wondering if we want to actually allow this though, but added it to a separate ticket.

v2/pkg/engine/resolve/context.go
fiam: I mean the naming. I understand these indices are eventually used for deletion, but the type being call `ResponseCacheInvalidationOptions` feels a bit weird. Are we planning to extend it later? Otherwise, maybe it might be clearer as something as `ResponseCacheIndexingOptions`?

v2/pkg/engine/resolve/response_cache_test.go
fiam: IMHO it saves a bit of space because we avoid the repeated `t.Run(...)` in the code, but no strong opinion. Happy to leave it as is in this PR.

v2/pkg/engine/resolve/response_cache_test.go
fiam: Great, thank you!

fiam (approved):

v2/pkg/engine/resolve/context.go
SkArchon: got it, will rename it

SkArchon: renaming to `ResponseCacheTagIndexOptions`

PR #1656, opened 2026-09-03
feat: support caching in multi fetches

Noroth:
This PR adds support for response caching in multi entity fetches. 

## Checklist

## Open Source AI Manifesto

## Summary by CodeRabbit

* **New Features**
  * Multi-entity responses can be cached and reused independently by alias and entity.
  * Partially cached requests fetch only missing entries, while fully cached requests avoid network calls.
  * Cache entries remain isolated by request details and support aggregated expiration timing.
  * Cached results are available to response-processing hooks, including partial and fully cached requests.

* **Bug Fixes**
  * Improved handling of partial failures and unmatched errors when combining cached and newly fetched results.
  * Invalid or malformed cache entries are detected and refreshed from the origin.
  * Preserved upstream response details while retaining successfully cached entries during partial failures.


v2/pkg/engine/resolve/loader_multi_entity.go
```diff
@@ -173,41 +228,186 @@ func (l *Loader) renderEntryRepresentations(entry *MultiEntityFetchEntry, entryR
...
+// that renders representations. Each entry hashes its own slice of the request,
+// so warm entries can be recognised and switched off one at a time.
+func (l *Loader) setResponseCacheKeys(entries []preparedMultiEntry, representationHashMatrix [][]uint64, assembled *assembleMultiEntityResult) error {
+	if !l.responseCacheEnabled() {
+		return nil
+	}
+
+	if len(entries) != len(representationHashMatrix) && len(assembled.entryVariables) != len(entries) {
+		return errors.New("unexpected number of entries in representation hash matrix or entry variables")
+	}
```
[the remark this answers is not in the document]
Noroth: I decided to get rid of it. 

The slices are correctly initialized with fixed values for the size of the entries. 
In this case the code structure actually defines the correctness as there are not external influences on the size. 

Returning an error here would probably be less effective than having a panic that is pointing out that any changes that have been made will actually break the logic.

v2/pkg/engine/resolve/loader.go
```diff
@@ -430,6 +430,16 @@ func (l *Loader) loadPhase(ctx context.Context, prepared *preparedFetch) error {
 	if prepared.skipLoad {
 		return nil
 	}
+	// A merged fetch resolves its cache per entry, and that reshapes the request,
+	// so it happens here rather than on the whole-fetch path below.
+	if prepared.multiAssembly != nil {
+		if served, err := l.applyMultiEntityResponseCache(ctx, prepared); served || err != nil {
+			// If either every entry was warm or there was an error, we can return early.
+			// errors.WithStack() returns nil if the error is nil.
```
fiam: nit: `// errors.WithStack() returns nil if the error is nil.`  IMHO is a bit redundant

v2/pkg/engine/resolve/loader_multi_entity.go
```diff
@@ -16,6 +23,32 @@ type preparedMultiEntry struct {
...
+	// cachedValues is set only when every one of responseCacheKeys hit: the
+	// stored entity objects, written back into the response at merge time.
+	cachedValues [][]byte
+	// responseCacheTTL is the life left on the least fresh of cachedValues.
+	responseCacheTTL time.Duration
+}
+
+// cacheHit reports whether this entry was answered entirely from the response
+// cache, in which case it was switched off in the merged request.
+func (e *preparedMultiEntry) cacheHit() bool { return e.cachedValues != nil }
```
fiam: nit: comparing a slice to `nil` is a bit dangerous IMHO — Imagine we add support for pooling  `preparedMultiEntry` and we trim the slice to zero items before reuse.

Could we change it to `len() == 0`?

v2/pkg/engine/resolve/loader_multi_entity.go
```diff
@@ -77,12 +120,16 @@ func (l *Loader) prepareMultiEntityFetch(fetchItem *FetchItem, fetch *MultiEntit
 		res.fetchSkipped = true
 		prepared.skipLoad = true
 		if l.ctx.TracingOptions.Enable {
-			l.setTracingInput(fetchItem, input, fetch.Trace)
+			l.setTracingInput(fetchItem, assembled.out, fetch.Trace)
```
fiam: nit: this seems a bit confusing because it seems to set an input to an output — maybe we could document the fields in `assembleMultiEntityResult`? Or maybe we need to interpret is as a pipe, where an output becomes the input for the next step? No strong opinion TBH.

v2/pkg/engine/resolve/loader_multi_entity.go
```diff
@@ -173,41 +250,186 @@ func (l *Loader) renderEntryRepresentations(entry *MultiEntityFetchEntry, entryR
...
+		prepared.res.responseCacheTTL = shortestCachedEntryTTL(prepared.multiEntries)
+		prepared.responseCacheHit = true
+		if prepared.trace != nil {
+			prepared.trace.LoadSkipped = true
+		}
+		return true, nil
+	}
+
+	// Some entries are still cold, so a request goes out after all: rebuild the
+	// body with the warm ones switched off, so the origin is asked for no more
```
fiam, line 420: just a note for later: we need to make sure we document this behavior in the router docs

v2/pkg/engine/resolve/loader_multi_entity.go
```diff
@@ -113,26 +165,47 @@ func (l *Loader) authorizeEntry(entry *MultiEntityFetchEntry, entryRes *result)
...
+	items []*astjson.Value,
+	itemInput *arena.Buffer,
+	tools *batchEntityTools,
+) (*renderEntryRepresentationsResult, error) {
 	repsBuf := arena.NewArenaBuffer(tools.a)
 	batchStats := arena.AllocateSlice[[]*astjson.Value](tools.a, 0, len(items))
 	batchItemIndex := 0
 	addSeparator := false
+
+	responseCacheItemHashes := make([]uint64, 0, len(items))
```
fiam: question: is `len(items)` here guaranteed to be > 0?

fiam (commented):
Left you some comments, I think it would be great if we could go over it synchronously.

v2/pkg/engine/resolve/loader_multi_entity.go
Noroth: I can change the name to inputBuffer or smth like that

v2/pkg/engine/resolve/loader_multi_entity.go
Noroth: Well entries should be always greater than zero, otherwise we don't look up any entities.

But the slice would be empty when response caching is disabled.

v2/pkg/engine/resolve/response_cache.go
```diff
@@ -348,6 +367,153 @@ func responseCacheTagIdentities(input responseCacheTagInput) []string {
...
+			if value.Type() != astjson.TypeObject {
+				continue
+			}
+			item := caching.Item{
+				Key:   entry.responseCacheKeys[j],
+				Value: value.MarshalTo(nil),
+				TTL:   ttl,
+			}
+
+			if invalidation := l.ctx.responseCache.invalidation; invalidation.any() {
```
Noroth, line 424: @fiam we need to check if this is correct or if we need to handle this differently. The cache tags are a flat list, so basically when we have a multi fetch, we have aliased representations. Therefore we cannot assign this to any of those. I'm leaving the declared tags out for now.

v2/pkg/engine/resolve/loader_multi_entity.go
```diff
@@ -342,28 +590,109 @@ func (l *Loader) applyParsedResponseToEntries(prepared *preparedFetch, response
...
+	entryRes.statusCode = res.statusCode
+	entryRes.ds = res.ds
+	entryRes.out = res.out
+}
+
+// applyCachedEntriesToResponse writes the entities a cache hit already answered
+// into response under that entry's alias — they have none of their own, having
+// been switched off in the request — so that nothing downstream of here can tell
+// a hit from a fetch.
+func (l *Loader) applyCachedEntriesToResponse(prepared *preparedFetch, response *astjson.Value) error {
```
ysmolski, line 634: I found an issue with corrupted entries. Right now any corrupted entry would lead to the whole response being invalidated. But we could do better. Every cached value could be validated before being used. And only valid cache entries would lead to skipped representations calls/ids

ysmolski: Actually the fix is not that hard and would work for single fetches too

ysmolski: Fixed

v2/pkg/engine/resolve/loader_multi_entity.go
```diff
@@ -279,32 +500,52 @@ func (l *Loader) mergeMultiEntityResult(prepared *preparedFetch) error {
...
+		// fetch takes, no request having been sent for it.
 		l.applyTransportStateToEntries(prepared)
+
+		cached, err := l.serveCachedEntriesWithoutResponse(prepared)
+		if err != nil {
+			return err
+		}
+		// Only when nothing was sent. A request that failed keeps its body,
+		// which is the one a hook wants to see.
+		if len(res.out) == 0 && cached != nil && l.ctx.LoaderHooks != nil {
```
ysmolski: `res.out` could be empty here because the subgraph fetch has failed on a trasport level. going to fix

ysmolski: and fixed

v2/pkg/engine/resolve/loader_multi_entity_test.go
```diff
@@ -644,3 +696,605 @@ func TestLoadGraphQLResponseData_MultiEntity_SingleFlight(t *testing.T) {
...
+	// router's Cache-Control merge needs that lifetime, or it advertises f2's
+	// max-age for a response half of which expires sooner.
+	for key, item := range cache.items {
+		if bytes.Contains(item.Value, []byte("notes")) {
+			delete(cache.items, key)
+		}
+	}
+	partial := &recordingDataSource{
+		response:        []byte(`{"data":{"f2":[{"notes":"n"}]}}`),
+		responseHeaders: cacheableHeaders(),
```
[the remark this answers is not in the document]
ysmolski: fixed

ysmolski (approved):
I found some small issues and fixed those. I have documented another edge case with errors as TODO.
LGTM.

ysmolski:
I have reviewed and approved everything up to the commit: ea3812aa69ef71c991a4ab6653ca22107d29739b
Please review everything I have added on top: https://github.com/wundergraph/graphql-go-tools/pull/1656/changes/ea3812aa69ef71c991a4ab6653ca22107d29739b..1437c13f4cb57f8da1d882fc7ed4b33df0b1c608

v2/pkg/engine/resolve/loader_multi_entity.go
SkArchon: We only validate if the "json is valid" is what I understand? (if the data is in a valid json and invalid format it will still get stored in the cache?)

Can you make it a bit more explicit in the tests, I initially thought this test "TestLoadGraphQLResponseData_SingleFetch_ResponseCacheCorruptEntry" was checking if the data in the json was incorrect structurally (but is valid json) initially

SkArchon (changes requested):
1 comment

v2/pkg/engine/resolve/loader_multi_entity.go
ysmolski: Yes, format is not checked. Just a JSON validation. I have renamed 2 tests and added clarifying comments too.

SkArchon (approved):
lgtm

PR #1657, opened 2026-09-03
fix: improve performance of orderSequenceByDependencies

ysmolski:
There are multiple improvement that I have applied to this post processor. This happened on plans created for mutation with more than 30 fetches. Such plans have densely populated dependencies (each next fetch in a sequence depends on all fetches before).

In the end that post-processor was completely rewritten.
Things were done roughly in this order:

* I have pre-computed the calculation of all dependencies per fetch.
* Topological comparisons were simplified because the size of dependencies gives the good ordering, that is only followed by the fetch IDs.
* In the pre-compute instead of using maps I drafted very simple implementation of bitset since fetch IDs are densely populated.

Closes #1549

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved execution ordering for complex dependency graphs, including deferred dependencies.
  - Improved fetch scheduling and dependency handling for more reliable parallel execution.
  - Prevented potential ordering errors with very large fetch identifiers.

- **Performance**
  - Reduced processing overhead when sequencing large execution plans.

- **Quality Improvements**
  - Standardized deferred terminology and expanded automated coverage for dependency ordering and set operations.


fiam (approved):
LGTM

devsergiy (approved):

PR #1659, opened 2026-09-04
fix: reset operation complexity state after aborted walk

fiam:
Linear: [ROUTER-655](https://linear.app/wundergraph/issue/ROUTER-655/reset-operation-complexity-estimator-depth-state-after-an-aborted-walk)

Fixes [ENG-9974](https://linear.app/wundergraph/issue/ENG-9974/complexity-estimator-leaks-state-after-an-aborted-walk)

## Summary

- Reset the per-root-field maximum depth before each complexity estimation.
- Add a regression that aborts one walk, reuses the estimator, and compares the next result with a fresh estimator.

## Context

An aborted walk can skip the root-field cleanup in `LeaveField`. Reusing that estimator then carries a stale `maxRootFieldDepth` into the next operation and can report an incorrect root-field depth.

This PR is stacked on [#1648](https://github.com/wundergraph/graphql-go-tools/pull/1648), so its diff contains only the estimator-reuse fix. PRs [#1626](https://github.com/wundergraph/graphql-go-tools/pull/1626) and [#1647](https://github.com/wundergraph/graphql-go-tools/pull/1647) included equivalent resets alongside their alternative operation-depth rewrites; #1648 plus this focused follow-up supersede both.

## Tests

`go test ./v2/... -count=1`

Closes #1626

Closes #1647

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed operation complexity calculations when the estimator is reused across sequential operations.
  * Prevented depth values from carrying over after an interrupted or invalid operation analysis.
  * Ensured subsequent valid operations receive consistent complexity results after a prior analysis failure.

* **Tests**
  * Added regression coverage for reusing the estimator after an invalid inline fragment is encountered.
  * Verified reused-estimator results match those from a fresh estimator.

## Checklist

## Open Source AI Manifesto

This project follows the principles of the [Open Source AI Manifesto](https://human-oss.dev).

### DIFF
--- a/v2/pkg/middleware/operation_complexity/operation_complexity_test.go
+++ b/v2/pkg/middleware/operation_complexity/operation_complexity_test.go
@@ -608,6 +608,53 @@ func runDepthTest(t *testing.T, operationString string, expectedStats OperationS
 	assert.Equal(t, expectedStats, rootFieldStats[0].Stats)
 }
 
+func TestOperationComplexityEstimatorReuseAfterAbortedWalk(t *testing.T) {
+	definition := unsafeparser.ParseGraphqlDocumentString(testDefinition)
+	estimator := NewOperationComplexityEstimator(false)
+
+	invalidOperation := unsafeparser.ParseGraphqlDocumentString(`
+		{
+			users(first: 1) {
+				transactions(first: 1) {
+					sender {
+						address {
+							... on UnknownType {
+								city
+							}
+						}
+					}
+				}
+			}
+		}`)
+	invalidReport := operationreport.Report{}
+	estimator.Do(&invalidOperation, &definition, &invalidReport)
+	require.True(t, invalidReport.HasErrors())
+
+	operation := unsafeparser.ParseGraphqlDocumentString(`
+		{
+			users(first: 1) {
+				id
+				address {
+					city
+				}
+			}
+		}`)
+	normalizationReport := operationreport.Report{}
+	astnormalization.NormalizeOperation(&operation, &definition, &normalizationReport)
+	require.False(t, normalizationReport.HasErrors())
+
+	wantReport := operationreport.Report{}
+	wantGlobal, wantRootFields := NewOperationComplexityEstimator(false).Do(&operation, &definition, &wantReport)
+	require.False(t, wantReport.HasErrors())
+
+	gotReport := operationreport.Report{}
+	gotGlobal, gotRootFields := estimator.Do(&operation, &definition, &gotReport)
+	require.False(t, gotReport.HasErrors())
+
+	assert.Equal(t, wantGlobal, gotGlobal)
+	assert.Equal(t, wantRootFields, gotRootFields)
+}
+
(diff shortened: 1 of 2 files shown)


v2/pkg/middleware/operation_complexity/operation_complexity_test.go
```diff
@@ -608,6 +608,53 @@ func runDepthTest(t *testing.T, operationString string, expectedStats OperationS
...
+			users(first: 1) {
+				id
+				address {
+					city
+				}
+			}
+		}`)
+	normalizationReport := operationreport.Report{}
+	astnormalization.NormalizeOperation(&operation, &definition, &normalizationReport)
+	require.False(t, normalizationReport.HasErrors())
```
ysmolski: why there is normalization for operation, but not for invalidOperation?

ysmolski (commented):
I do not see Cosmo resuing this estimator anywhere. Also issue https://github.com/wundergraph/graphql-go-tools/pull/1626 fixes something different, like the order of fields mixed with fragments.

v2/pkg/middleware/operation_complexity/operation_complexity_test.go
fiam: `invalidOperation` would fail normalization. The idea is to pass an invalid operation to `estimator.Do()` to set up the invalid state, then pass an operation through the  full happy path (which typically includes normalization) and check that the `estimator` state was properly reset.

Nonetheless, now I realize that we can probably skip the normalization phase for the happy path and the test would still cover what we need.

ysmolski: I am still not sure what are we testing here :) Why do we want to care about reusability of estimator?

fiam: `OperationComplexityEstimator` is designed to be reusable, see this slice reuse [here](https://github.com/wundergraph/graphql-go-tools/pull/1659/changes#diff-664d279bca570ab2de2194ff1af9f6edf35d63d84b9a0c11f24a10c9af4267deR101)

All other fields were resetted at the start of `Do()` except for this one. This PR fixes that behavior.

fiam:
> I do not see Cosmo resuing this estimator anywhere

Correct. Although I started investigating this because we believed it could be a cause of non-determinism in query complexity calculation, I did not find any place in Cosmo where we reused an estimator. Nonetheless, the implementation is clearly trying to support reuse and it was failing because it was not resetting all fields. I decided that this was still worth fixing as a library bug.

> Also issue #1626 fixes something different, like the order of fields mixed with fragments.

Partially correct. #1626 fixes the depth calculation for fields mixed with fragments, but it also includes the resetting the internal state in `Do()`, which fixes the result after an aborted walk (which this PR addresses too)

The field-order/fragment fix part of #1626 is covered by #1648, which this PR is stacked on. I opted for a different fix because #1626 adds a `LeaveSelectionSet` visitor, introducing an additional skip check and visitor callback for each visited selection set, with some additional overhead.

#1648 handles the depth bookkeeping in the existing field visitors, avoiding that additional function call overhead.

So between #1648 and this PR, #1626 becomes obsolete.

ysmolski (approved):

PR #1662, opened 2026-09-07
feat: router cache tag based invalidation for cdns

SkArchon:
This PR allows the returning and combining of header tags in responses which can be written as headers by the router.

## Summary by CodeRabbit

* **New Features**
  * Added response-cache metadata that preserves and returns header tags with cached values.
  * Added support for generating, sorting, and deduplicating cache header tags for CDN purging.
  * Introduced a versioned cache-entry format with validation for malformed entries.
  * Response-cache tags are now available during request resolution and on cache hits or misses.
  * Added safeguards that reject oversized header-tag metadata.

* **Tests**
  * Added coverage for tag encoding, decoding, ordering, deduplication, propagation, size limits, and cache behavior.

## Checklist

## Open Source AI Manifesto


ysmolski:
Reviewing

v2/pkg/caching/entry.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"encoding/binary"
+	"errors"
+)
+
+// Entry envelope, header tags ahead of the value so a hit reads them without
+// touching the value:
+//
+//	[version byte][uvarint n]{[uvarint len][tag]}*n[value...]
```
ysmolski: This notation has some issues and hard to read. `version` is not a type, but `byte` is. I suggest to present layout in a less ambiguous format. The part for `len` and `tag` is not clear. what does `*n` mean?

v2/pkg/caching/key.go
```diff
@@ -5,12 +5,13 @@ import (
...
-// under the old layout instead of letting them be read back as something they
-// are not. Orphaned entries are not deleted, they simply stop being asked for
-// and fall out on their own TTL.
-const keyFormatVersion = "v1"
+// keyFormatVersion leads every key this package builds. A change to the key
+// layout, to the meaning of either hash, or to the stored entry format (see
+// entry.go) bumps it, which orphans the entries written under the old layout
+// instead of letting them be read back as something they are not. Orphaned
+// entries are not deleted, they simply stop being asked for and fall out on
+// their own TTL.
```
ysmolski, line 13: The passage about orphaned entries is just a wish or is it maintained by the code? Is it possible to create entries with TTL approaching infinity with our codes?

v2/pkg/engine/resolve/response_cache.go
```diff
@@ -321,31 +337,52 @@ type responseCacheTagInput struct {
 	opts        ResponseCacheTagIndexOptions
 }
 
-func responseCacheTagIdentities(input responseCacheTagInput) []string {
-	if input.subgraph == "" {
-		return nil
```
ysmolski, line 350: why did you remove this check? unnamed subgraph should be indexed?

v2/pkg/engine/resolve/response_cache.go
```diff
@@ -321,31 +337,52 @@ type responseCacheTagInput struct {
...
+	if !input.isRootFetch {
+		typeName = string(input.value.GetStringBytes("__typename"))
+	}
 
-	for _, tag := range input.declared {
-		identities = append(identities, caching.DeclaredTag(input.subgraph, tag))
+	if typeName != "" {
+		headerTags = append(headerTags, caching.TypeHeaderTag(input.subgraph, typeName))
 	}
+	headerTags = append(headerTags, input.declared...)
```
ysmolski: We do not prefix items in input.declared with anything.  Those things come from the subgraph response, right? And they potentially could contain anything. What if those things contains prefixes that routers uses `subgraph-` and `type-` ? I see this leading to invalid headerTags. Also the thing on the line 363 uses `caching.DeclaredTag` to wrap them with prefix…

v2/pkg/engine/resolve/context.go
```diff
@@ -29,6 +29,7 @@ type Context struct {
 	// the resolver will use the new name to look up the old name to render the variable in the query.
 	RemapVariables map[string]string
 
+	// Verification
```
ysmolski: Out-of-place comment

v2/pkg/engine/resolve/loader.go
```diff
@@ -468,6 +471,7 @@ func (l *Loader) mergePhase(prepared *preparedFetch) error {
 	if err := l.responseCacheCollect(prepared); err != nil {
 		l.reportResponseCacheError(fmt.Errorf("response cache collect error: %w", err))
 	}
+	l.responseCacheMergeHeaderTags(prepared.res)
```
ysmolski: We might need to handle errors specifically in `renderErrorsFailedToFetch` for inbound request deduplication (singleflight). This one I did not verify in full but asked Claude to explain it:

ysmolski (changes requested):

v2/pkg/engine/resolve/response_cache.go
ysmolski: Repro test:

```
// Same input as "a tag spelled like a derived one lands beside them and not
// among them", but read off the header tier instead of the index tier.
func TestReproDeclaredHeaderTagForgesDerivedTier(t *testing.T) {
	value, err := astjson.Parse(`{"__typename":"User","id":42}`)
	require.NoError(t, err)

	_, headerTags := responseCacheIdentities(responseCacheTagInput{
		// What the subgraph put in extensions, spelled as the router's own tiers.
		declared: []string{caching.SubgraphHeaderTag("payments"), caching.TypeHeaderTag("payments", "Invoice")},
		value:    value,
		subgraph: "accounts",
		opts:     ResponseCacheTagIndexOptions{},
	})

	t.Logf("header tags: %q", headerTags)

	// The accounts entry must not advertise payments' purge keys.
	require.NotContains(t, headerTags, "subgraph-payments")
	require.NotContains(t, headerTags, "type-payments-Invoice")
}
```

v2/pkg/engine/resolve/response_cache.go
ysmolski: An unnamed data source now indexes under `subgraph:` and `type::User` and emits header tags `subgraph-` and `type--User`, shared by every unnamed source. Repro:

```
// An unnamed data source must not be indexed: without a scope the entry lands
// where another subgraph's invalidation can reach it. Pinned on master by
// "an unnamed subgraph is not indexed at all", deleted in #1662.
func TestReproUnnamedSubgraphStillIndexed(t *testing.T) {
	value, err := astjson.Parse(`{"__typename":"User","id":42}`)
	require.NoError(t, err)

	// opts.any() is true: the index config is request level and says nothing
	// about whether this fetch's data source had a name.
	opts := ResponseCacheTagIndexOptions{CacheTag: true, Subgraph: true, Type: true}
	require.True(t, opts.any())

	tags, headerTags := responseCacheIdentities(responseCacheTagInput{
		declared: []string{"users"},
		value:    value,
		subgraph: "",
		opts:     opts,
	})

	require.Empty(t, tags, "an unnamed subgraph must not be indexed, got %q", tags)
	require.Empty(t, headerTags, "an unnamed subgraph must not emit header tags, got %q", headerTags)
}
```

v2/pkg/engine/resolve/loader.go
ysmolski: ### How a partial failure is handled today

When one subgraph fetch fails, the loader does not fail the request.
`renderErrorsFailedToFetch` in `ggt/v2/pkg/engine/resolve/loader.go` records a
subgraph error on the loader, appends a GraphQL error object to the response,
and returns `nil`. The other fetches still merge their data. Resolution
succeeds, and the client gets a body with both `errors` and partial `data`.
At the end the loader copies its recorded errors onto the request's `Context`,
where `SubgraphErrors()` exposes them.

### How the router reacts on the leader

The router's response writer at `cosmo/router/core/header_rule_engine.go:132`
checks `SubgraphErrors()` on its own context. When it finds any, it sets

    Cache-Control: no-store, no-cache, must-revalidate

on the HTTP response and flips `didSetSubgraphErrors`. The new `Cache-Tag`
block a few lines below skips the header when that flag is set. The comment
there says why: an errored body is marked no-store, so it must not be
advertised under purge keys. That logic is correct for the leader.

### Where it breaks

Look at the leader's tail in `ggt/v2/pkg/engine/resolve/resolve.go:520`.
Regardless of subgraph errors, it copies its collected header tags into
`inflight.HeaderTags` and calls `FinishOk` with the body bytes. The header tags
are non-empty because the fetches that succeeded were cache hits or cacheable
misses and contributed their tags. Only the failed fetch contributed nothing.

Now a follower wakes up at `resolve.go:427`. It has its own fresh `Context`.
The loader never ran for it, so its `SubgraphErrors()` is `nil`. It adopts
`inflight.HeaderTags` and writes the leader's body bytes through its own router
writer. In that writer the subgraph-error check finds nothing, so no
`Cache-Control: no-store` is set and `didSetSubgraphErrors` stays `false`.
The `Cache-Tag` block then runs and emits the tags.

### The result

Two identical HTTP responses leave the router within milliseconds.

| Response | Body            | `Cache-Control`                          | `Cache-Tag`                     |
|----------|-----------------|------------------------------------------|---------------------------------|
| Leader   | errors + data   | `no-store, no-cache, must-revalidate`    | absent                          |
| Follower | same bytes      | absent (from this path)                  | `subgraph-b,type-b-Product`     |

The follower's body is the errored one. A CDN in front stores it under B's
purge keys and serves it until its own TTL. Purging B's tags evicts it, but
nothing else does, and the object is an error response that should never have
been cacheable.

### Why this is new to the branch

The follower missing `Cache-Control: no-store` on an errored body is a
pre-existing gap in singleflight, since subgraph errors were never part of the
shared state. On master that gap is mostly harmless, because there is no
positive cache signal either. This branch adds a positive signal, the
`Cache-Tag` header, and shares it across followers without the negative signal
that gates it on the leader.

ysmolski: ### Fix options

1. **Narrow fix (ggt).** Leave `inflight.HeaderTags` empty when
   `ctx.SubgraphErrors()` is non-nil, so followers inherit the leader's
   decision.
2. **Structural fix (router).** Write the `Cache-Tag` into the propagation
   header map that `SharedData` already snapshots after the leader's write and
   applies before the follower's write. The follower then receives a clone of
   the leader's actual headers, including the absence of `Cache-Tag` and the
   presence of no-store. This also removes `InflightRequest.HeaderTags` and
   resolves the shared-slice finding.

Either way a router test with two concurrent identical requests where one
subgraph fails is needed. Neither repo covers the follower path with errors.

v2/pkg/engine/resolve/loader.go
```diff
@@ -147,6 +147,9 @@ type result struct {
 	// unlocked load phase is safe.
 	responseCacheHit bool
 	responseCacheTTL time.Duration
+	// responseCacheHeaderTags is what the fetch contributes to the cache tag
+	// header, read back on a hit and computed on a miss.
+	responseCacheHeaderTags []string
```
ysmolski: I guess you have not handled the Multi-entity fetches in this PR. Because it was not merged at the time of your PR. It is merged right now.

ysmolski: ### Where

`ggt/v2/pkg/engine/resolve/loader.go:467`, in `mergePhase`:

    if prepared.multiEntries != nil {
        return l.mergeMultiEntityResult(prepared)
    }
    ...
    l.responseCacheMergeHeaderTags(prepared.res)

The multi-entity branch returns before the header tag merge.
`mergeMultiEntityResult` never calls it either. The same gap applies to any
fetch that is silent to the cache: no `Cache-Control`, `private`, `no-store`,
or a non-GraphQL root data source. Those fetches put data in the body but add
nothing to the request's tag set.

### Effect

The `Cache-Tag` header describes only part of the response.

    query { products { id price reviews { body } } }

`products` is cacheable and tags the response `subgraph-products,type-products-Product`.
`reviews` runs as a multi-entity fetch, or answers without `Cache-Control`,
and contributes nothing. The CDN stores the whole body under products' keys
only. An operator edits a review and purges `subgraph-reviews`. Nothing
matches, and the stale review body is served until the CDN TTL ends.

### Why it matters

The header's purpose is to let a CDN purge everything a subgraph touched.
A partial tag set breaks that contract silently, since the response looks
fully tagged. Only the config schema description mentions the limitation.
No test covers a mixed cacheable/uncacheable response or a multi-entity fetch
with the header on.

### Fix options

1. **Suppress.** Mark the request partially covered when any data fetch
   contributes no tags, and emit no header at all.
2. **Floor.** Always add `subgraph-<name>` for every subgraph touched,
   cacheable or not, so a subgraph-level purge always reaches the object.

Either needs a router test with one cacheable and one untagged subgraph, and
one with a multi-entity fetch.

v2/pkg/engine/resolve/response_cache.go
SkArchon: My reasoning was from what I checked cosmo can never have a subgraph that was empty, but Ig from the point of graphqlgotools we should have that check, will add back.

v2/pkg/engine/resolve/response_cache.go
SkArchon: Imo I think this is fine, we have two cases (let me know if I misunderstood the point)

**Case One**: Where you invalidate based on the tag type and subgraph, for which we need to differentiate. Like we can deprecate by calling the invalidation API with the following json (typing from the top of my head so the actual syntax may vary). 
```
{ subgraph: "employee", invalidationType: "subgraph" }
```
OR (note that subgraph is a mandatory for now technically, I ignored to avoid confusion)
```
{ invalidationType: "tag", tagName: "random-string" }
```

if someone wants to shoot themselves in the foot by sending a tag called "subgraph-employee" from upstream, I think its fine. HOWEVER for invalidating "subgraph-employee" as a tag is different from invalidating all entries related to the subgraph "employee".

**Case Two**: When it comes to the emitting cdn tags, we emit them verbatim. So "subgraph-employee" can be emitted for the accounts subgraph if that was emitted verbatim from upstream.
 
cc: also thoughts @fiam?

v2/pkg/caching/key.go
SkArchon: There can be no entry stored without a ttl, but they could have an unbounded ttl like 10 years, etc.
We don't have a max set, maybe we can allow users to set an optional max, though this still seems like cases of "users shooting themselves in the foot". Thoughts? @fiam also?

v2/pkg/caching/entry.go
SkArchon: updated comment

v2/pkg/engine/resolve/loader.go
SkArchon: done

v2/pkg/engine/resolve/loader.go
SkArchon: done

v2/pkg/engine/resolve/response_cache.go
ysmolski: Thanks!

v2/pkg/engine/resolve/response_cache.go
ysmolski: I think I am lost again after rereading the code. What is the difference between tags and headerTags? I think this code has an issue with naming things.

v2/pkg/engine/resolve/loader.go
ysmolski: hmm i cannot find the commit with the fix

SkArchon: https://github.com/wundergraph/graphql-go-tools/pull/1662/changes/4a461f0359322b6cbcce0c416b4d8345a703a470

v2/pkg/engine/resolve/response_cache.go
SkArchon: tags are what we use as indexes in redis for invalidation, whereas header tags is what returned to the user. They are from the same source. The source being, whenever an upstream subgraph request, responds with lets say ```{ extensions.tags: {entity_1A: ["senior-employee", ...]} }``` We create indexes based on this for the entities in the response. In addition to this we store these tags directly on the entity also (so we can get everything with one call to redis and dont need to traverse the indexes, etc).

I've attached a diagram to help out
image

v2/pkg/caching/key.go
ysmolski: Let's keep as it is until more is wanted

v2/pkg/engine/resolve/response_cache.go
ysmolski: Thanks for explaining it

ysmolski (approved):
+2 LGTM

PR #1672, opened 2026-09-14
fix(resolve): read remapped variables in subscription filter type check

ksteweee:
## What
The filter compared the rendered value correctly. It looked up the variable type with the post-normalization name (`a`) against the un-renamed variables JSON. Any variable segment except `$a` skipped every event.

Example: EDFS `@openfed__subscriptionFilter(condition: { IN: { fieldPath: "id", values: ["{{ args.id }}"] } })`. The client passes `id` as a variable, and no events pass the filter. Inline literals work because they take the `StaticSegmentType` branch.

## Fix
Use `ctx.VariablesView().Get` like the rest of the resolver. One line.

## Tests
- Remapped string variable matches
- Remapped string variable does not match
- Remapped variable missing from variables
- Remapped variable with a type mismatch still skips
- Remapped number and boolean match
- No remap regression
- Remap present but path not in the map

## Notes
This fix was discussed with the maintainers beforehand.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

- **Bug Fixes**
  - Subscription event filtering now correctly evaluates remapped variables in `in` predicates.
  - Filtering handles matching, missing, mismatched, and remapped string, number, and boolean values consistently.

- **Tests**
  - Added coverage for variable remapping, absent mappings, missing variables, and type mismatches.


v2/pkg/engine/resolve/subscription_filter.go
```diff
@@ -117,7 +117,7 @@ func (f *SubscriptionFieldFilter) SkipEvent(ctx *Context, data []byte) (bool, er
 
 				switch f.Values[i].Segments[0].SegmentType {
 				case VariableSegmentType:
-					value := ctx.Variables.Get(f.Values[i].Segments[0].VariableSourcePath...)
+					value := ctx.VariablesView().Get(f.Values[i].Segments[0].VariableSourcePath...)
```
ysmolski: @dkorittki Ther only thing is that we call `VariablesView()`for every item of `f.Values`. Will it matter?

ysmolski (approved):

PR #1673, opened 2026-09-15
feat: add selected field count to operation complexity estimates

fiam:
## Summary

Expose `OperationStats.FieldCount` globally and per root field so callers can limit the number of selected fields. For `{ user { address { city } } }`, the field count is 3; adding 199 aliased `city` selections raises it to 202 while `NodeCount` remains 2.

Count object fields, leaf fields, and implicit `__typename` once per normalized selection, without list-size multipliers. Preserve the existing global `NodeCount`, `Complexity`, and `Depth` calculations and respect `@nodeCountSkip` and the introspection exclusion option. Limit per-root accounting to top-level operation fields: when `Query.self` returns `Query`, `{ self { __typename } }` reports one root entry for `self` instead of misattributing its statistics to the nested field.

Part of [ROUTER-705](https://linear.app/wundergraph/issue/ROUTER-705). Companion: [Cosmo #3249](https://github.com/wundergraph/cosmo/pull/3249) uses this metric for `total_fields` and requires a tagged release of this PR.

## Validation

- `cd v2 && go test -race ./pkg/middleware/operation_complexity`
- `cd execution && go test -race ./graphql`
- `cd v2 && go tool golangci-lint run ./pkg/middleware/operation_complexity/...`

Existing fixtures verify field counts for leaf fields, introspection, skipped subtrees, and multipliers, and the existing estimator-reuse test compares all stats. Focused regressions cover scalar roots and aliases, implicit `__typename`, reused and merged fragments, nested root types with implicit and regular fields, and retained field names after operation-buffer reuse. Existing node-count, complexity, and depth expectations remain unchanged. The companion Cosmo PR covers widening from 1 to 200 leaves against the router limit.

## Checklist

- [x] Changes are scoped to the linked issue.
- [x] Coding standards and static analysis checks pass.
- [x] Tests have been added or updated.

## Open Source AI Manifesto

Prepared with AI assistance; the implementation and validation are described above.

## Summary by CodeRabbit

- **New Features**
  - Operation complexity statistics now include selected field counts globally and for individual root fields.
  - Field counts account for aliases, fragments, merged fields, nested selections, and implicit type-name fields while excluding introspection fields and list multipliers.
  - Field-count results remain consistent when operations are reused and when nested selections are present.

- **Tests**
  - Added coverage for field counting across queries, mutations, aliases, fragments, introspection, nested root selections, depth, multipliers, and reused operations.


v2/pkg/middleware/operation_complexity/operation_complexity.go
```diff
@@ -59,8 +64,9 @@ var (
 )
 
 const (
-	__schemaLiteral = "__schema"
-	__typeLiteral   = "__type"
+	__schemaLiteral   = "__schema"
+	__typeLiteral     = "__type"
+	__typenameLiteral = "__typename"
```
ysmolski: dont' we have those defined somewhere else in ggt? maybe `literal.TYPENAME` ?

ysmolski (approved):
Just 1 nit

alepane21 (approved):
LGTM

v2/pkg/middleware/operation_complexity/operation_complexity.go
fiam: Updated to use those variables, I was not aware of them, thanks! THey're `[]byte` instead of `string` but I did some benchmarks and I cannot see any statistically significant regression in performance.

ysmolski: Thanks!

ysmolski:
+2 LGTM

PR #1675, opened 2026-09-16
feat: response cache Vary support

SkArchon:
## Summary

Adds `Vary` header support to the response cache.

- `caching.Vary(headers)` parses the subgraph's `Vary` header into names. `*` and more than 100 names are not stored.
- Base key holds either a body or a **vary record** (header names, no value). A record yields `VariantKey(base, VaryDigest(names, sent))` and a second `GetMany` for the variant body.
- Values are the headers the router sent to that subgraph. Absent headers hash as empty.
- Write stores the record at the base key and the body under the variant key, same tags on both.
- `entry.go` envelope gains a kind byte (1=body, 2=record). `DecodeEntry` now returns `(value, headerTags, vary, err)`; `EncodeItem` picks the envelope.
- Key format bumped to `v4`.

## Tests

- `v2/pkg/caching/vary_test.go`
- `v2/pkg/engine/resolve/response_cache_vary_test.go`

## Summary by CodeRabbit

* **New Features**
  * Response caching now supports request-dependent representations using `Vary` headers, such as language-specific responses.
  * Cached variants are selected using the request headers actually sent, improving accuracy for single and multi-entity fetches.
  * Previously observed variants remain available when cache entries are updated.
  * Private cache entries retain precedence over shared entries.
  * Variant handling supports multiple header combinations while limiting excessive variation.

* **Bug Fixes**
  * Invalid or unsupported `Vary` responses are safely skipped.
  * Cache lookups avoid serving incomplete or mismatched variants and prevent distinct responses from being incorrectly merged.


v2/pkg/caching/vary.go
```diff
@@ -0,0 +1,39 @@
+package caching
+
+import (
+	"net/http"
+	"slices"
+	"strings"
+)
+
+// MaxVaryHeaders bounds a vary record. A response naming more is not stored.
+const MaxVaryHeaders = 100
```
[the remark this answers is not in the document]
SkArchon: updated pr description

v2/pkg/engine/resolve/response_cache.go
```diff
@@ -242,11 +294,18 @@ func (l *Loader) responseCacheCollect(prepared *preparedFetch) error {
...
-	ttl, private, ok := caching.TTL(responseCacheHeaders(res), l.ctx.responseCache.defaultTTL)
+	headers := responseCacheHeaders(res)
+	ttl, private, ok := caching.TTL(headers, l.ctx.responseCache.defaultTTL)
 	if !ok {
 		return nil
 	}
 
+	// "Vary: *" matches no request; a record past the cap is not kept either.
+	vary, star := caching.Vary(headers)
+	if star || len(vary) > caching.MaxVaryHeaders {
```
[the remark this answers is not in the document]
SkArchon: updated pr description

v2/pkg/caching/vary.go
```diff
@@ -0,0 +1,39 @@
...
+
+func Vary(headers http.Header) (names []string, star bool) {
+	for _, field := range headers.Values("Vary") {
+		for _, name := range strings.Split(field, ",") {
```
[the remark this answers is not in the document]
SkArchon: Done in cc66d8f2.

v2/pkg/engine/resolve/response_cache.go
```diff
@@ -176,6 +221,13 @@ func (l *Loader) responseCacheLookup(prepared *preparedFetch) bool {
 	return true
 }
 
+// responseCacheVaryDigest digests the values this request sends the fetch's
+// subgraph for names, the same on the way in and the way out of the cache.
+func (l *Loader) responseCacheVaryDigest(prepared *preparedFetch, names []string) caching.Digest {
+	sent, _ := l.ctx.HeadersForSubgraphRequest(prepared.res.ds.Name)
+	return caching.VaryDigest(names, sent)
```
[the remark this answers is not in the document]
SkArchon: Done in f87356b9: dispatch records the headers on the result and collect digests those.

v2/pkg/caching/key.go
```diff
@@ -14,13 +14,16 @@ import (
 //
 // Layout: version ':' entity ':' selection, plus ':' privateID for an entry
 // that belongs to one user, each a Digest in hex. The segment count keeps the
-// two forms apart, so a private key never reads as a public one.
-const keyFormatVersion = "v3"
+// two forms apart, so a private key never reads as a public one. Either form
+// may carry '+' vary for a variant of a response that varies by request
+// header; the separator differs so a variant never reads as a private key.
+const keyFormatVersion = "v4"
```
fiam, line 20: nit: don't increase this for now, revert it to `v1` for now. We will only start increasing after GA (and and that point we have to be able to decode older formats)

Also, maybe `1` would suffice? It saves a byte per entry.

v2/pkg/caching/entry.go
```diff
@@ -5,63 +5,99 @@ import (
...
+
 func EncodeEntry(value []byte, headerTags []string) []byte {
-	size := 1 + binary.MaxVarintLen64 + len(value)
-	for _, headerTag := range headerTags {
-		size += binary.MaxVarintLen64 + len(headerTag)
+	out := appendList(entryFormatBody, headerTags, len(value))
+	return append(out, value...)
+}
+
+func EncodeVaryRecord(names []string) []byte {
```
fiam, line 42: question: could we make this non-exported? I think `EncodeItem` should be enough

v2/pkg/caching/entry.go
```diff
@@ -5,63 +5,99 @@ import (
...
 
+// EncodeItem picks the envelope the item calls for.
+func EncodeItem(item Item) []byte {
+	if len(item.Vary) > 0 {
+		return EncodeVaryRecord(item.Vary)
+	}
+	return EncodeEntry(item.Value, item.HeaderTags)
+}
+
 func EncodeEntry(value []byte, headerTags []string) []byte {
```
fiam, line 37: question: IMHO this one could be non-exported too

fiam (commented):
Looking good, only small issues. It seems to me this is not handling multi-fetch yet, though. Could you please confirm?

v2/pkg/caching/key.go
SkArchon: I had it at 3 without the v, changed it to 1.

v2/pkg/caching/cache.go
```diff
@@ -31,6 +31,11 @@ type Item struct {
 	// Surrogate keys are what a client response carries so a CDN can purge by them.
 	// Stored with the value and returned by GetMany.
 	SurrogateKeys []string
+	// Vary makes the entry a vary record rather than a body: the sets of
+	// request headers responses at this key have varied on, newest first.
+	// Each set's values select a variant stored under VariantKey. A record has
+	// no Value and no SurrogateKeys.
+	Vary [][]string
```
fiam, line 38: question: is order relevant here? If not we might be better served with a map, or even a type that encapsulates some of the complexity. Would `http.Header` work here?

fiam (approved):

fiam (changes requested):
Fat fingered the approval, sorry. Requesting changes to prevent accidental merge.

v2/pkg/caching/cache.go
fiam: Aligned on a call, no changes needed here

v2/pkg/engine/resolve/response_cache.go
```diff
@@ -519,7 +675,6 @@ func (l *Loader) responseCacheCollectMultiEntity(prepared *preparedFetch, respon
 				continue
 			}
 			item := caching.Item{
-				Key:   writeKeys[j],
```
fiam, line 522: question: is this removal ok?

SkArchon: yes, this is done at a different location.

fiam: cool, thank you!

fiam (approved):

PR #1682, opened 2026-09-21
feat: per-subgraph response cache options

SkArchon:
## Summary

Lets the response cache be configured per subgraph. Keyed by `FetchInfo.DataSourceName`.

- `ResponseCacheOptions` gains `Subgraphs map[string]ResponseCacheSubgraphOptions{Disabled, DefaultTTL, PrivateID}` and `DefaultDisabled`. `DefaultTTL` and `PrivateID` stay as the defaults for subgraphs without an entry.
- An entry replaces the defaults whole for its subgraph: its private id is its own or none, and a zero `DefaultTTL` falls back to the default.
- The key-building sites for root, entity, batch and merged fetches consult the subgraph first. A disabled subgraph gets no keys, so no lookup and no write. Merged fetches are single-subgraph, so one check covers every entry.
- The inbound single-flight key now uses one digest over every private id in play (the default plus sorted per-subgraph pairs), so two users differing only in a per-subgraph id never share a resolution. Disabled entries and a disabled default contribute nothing.

## Tests

- `v2/pkg/engine/resolve/response_cache_subgraph_test.go`: gate per fetch shape, per-subgraph TTL and private keys, single-flight ids, merged fetch.

## Summary by CodeRabbit

* **New Features**
  * Added per-subgraph response-cache configuration.
  * Subgraphs can independently disable caching and customize cache duration and privacy settings.
  * Cache behavior now applies per subgraph, including multi-entity requests and request deduplication.

* **Bug Fixes**
  * Disabled or unconfigured subgraphs no longer create, read, or store cache entries.
  * Private cache isolation now accounts for configured subgraph privacy identifiers.


v2/pkg/engine/resolve/context.go
```diff
@@ -323,22 +369,72 @@ type ResponseCacheOptions struct {
...
+	if opts.DefaultDisabled {
+		// Nothing is keyed by it, so it must not keep users apart either.
+		opts.PrivateID = ""
 	}
 	if opts.PrivateID != "" {
-		c.responseCache.privateID = caching.DigestString(opts.PrivateID)
-		c.responseCache.hasPrivateID = true
+		rc.def.privateID = caching.DigestString(opts.PrivateID)
+		rc.def.hasPrivateID = true
+		rc.singleFlightID = rc.def.privateID
```
fiam, line 411: question: is this intentionally part of this PR? It seems like this feature should no produce changes here.

SkArchon: This is due to refactoring and also because how single flight changed. Essentially previously when "Jack" and "Bob" queried something we would only have one private id (for all subgraphs). However, now we can have multiple private ids, so we need to use all of them for the single flight id. E.g. :- as it stands now (if we used the default private id), if there are 10 subgraphs, one where the "Bob's" expression would resolve into a different value, with the current singleflight key we will get "Jack's" value.

fiam: Ah, I see. Thanks for the clarification!

v2/pkg/engine/resolve/response_cache_subgraph_test.go
```diff
@@ -0,0 +1,337 @@
...
+		loader.responseCacheSetKeys(prepared, sel, entities)
+		require.Equal(t, []string{caching.PrivateKey(e1, sel, caching.DigestString("acct-7"))}, prepared.responseCachePrivateKeys)
+
+		_, other := subgraphLoader(t, newTestCache(), "products", "public", body, opts)
+		prepared = &preparedFetch{res: other}
+		loader.responseCacheSetKeys(prepared, sel, entities)
+		require.Equal(t, []string{caching.PrivateKey(e1, sel, caching.DigestString("u1"))}, prepared.responseCachePrivateKeys)
+	})
+
+	t.Run("an entry without a private id has none, whatever the default", func(t *testing.T) {
```
fiam: nit: I'm not sure what is the "less surprising" behavior here, but if I'm reading this correctly the TTL gets inherited while the private ID does not. i.e. a subgraph with its own configuration with just private ID gets the default TTL, while a subgraph with its own configuration with just TTL gets no private_id.

I don't think we should settle this now (the feature is still marked as experimental), but we should document this very clearly.

fiam (approved):
LGTM

